PR #28, opened 2015-02-15
Introduced the Cython prange for parallel execution in ufuncify_matrix.

moorepants:
- [x] Add openmp as a dependency.

So this seems to work. All my cores are used to evaluate the ufuncified matrix, but there doesn't seem to be any speed gain. Here is some code to test it:

```
In [1]: from opty.utils import ufuncify_matrix

In [2]: from pydy.models import multi_mass_spring_damper
/home/moorepants/src/pydy/pydy/codegen/code.py:18: DeprecationWarning: This module, 'pydy.codgen.code', is deprecated. The function 'generate_ode_function' can be found in the 'pydy.codegen.ode_function_generator' module. 'CythonGenerator' has been removed, use 'pydy.codegen.cython_code.CythonMatrixGenerator' instead.
  DeprecationWarning)

In [3]: import numpy as np

In [4]: import sympy as sm

In [5]: sys = multi_mass_spring_damper(5)

In [6]: ms = list(sm.ordered(sys.constants_symbols))[-5:]

In [7]: m0 = np.random.random(int(1e7))

In [8]: m1 = np.random.random(int(1e7))

In [9]: m2 = np.random.random(int(1e7))

In [10]: m3 = np.random.random(int(1e7))

In [11]: m4 = np.random.random(int(1e7))

In [12]: res = np.empty((int(1e7), 25))

In [13]: f = ufuncify_matrix(ms, sys.eom_method.mass_matrix, tmp_dir='parallely')
```

For the parallel version (using my 4 cores):

```
In [14]: %timeit r = f(res, m0, m1, m2, m3, m4)
1 loops, best of 3: 573 ms per loop
```

For the single threaded version:

```
In [15]: %timeit r = f(res, m0, m1, m2, m3, m4)
1 loops, best of 3: 558 ms per loop
```

Oddly the single threaded version is faster.

**Note** that `res` is 2GB, so make sure enough memory is available.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ install:
   - conda info -a
   - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION
   - source activate test-environment
-  - conda install sympy cython cyipopt nose coverage sphinx matplotlib
+  - conda install sympy cython cyipopt nose coverage sphinx matplotlib openmp
 script:
   - nosetests -v --with-coverage --cover-package=opty
   - python setup.py install
--- a/README.rst
+++ b/README.rst
@@ -59,6 +59,7 @@ To run all of the examples the following additional dependencies are required:
 - pytables
 - pandas
 - yeadon
+- openmp
 
 **Currently only Linux and Mac are officially supported.** Although, it should
 be possible to install this on Windows with an appropriate Cython compilation
--- a/examples/parallel_example.py
+++ b/examples/parallel_example.py
@@ -1,0 +1,45 @@
+"""This script generates equations of motion that have a very large number of
+operations. It then creates a parallelized and non-parallelized version and
+compares the evaluation time."""
+
+import timeit
+
+import numpy as np
+import sympy as sm
+from pydy.models import n_link_pendulum_on_cart
+from opty.utils import ufuncify_matrix
+
+sys = n_link_pendulum_on_cart(10, False, False)
+
+c = np.random.random(int(1e7))
+
+args = [c for i in range(len(list(sys.constants_symbols) +
+                             list(sys.coordinates) + list(sys.speeds)))]
+
+res = np.empty((int(1e7), 11))
+
+udots = sm.Matrix([5.0 for u in sys.speeds])
+
+expr = sys.eom_method.mass_matrix * udots - sys.eom_method.forcing
+
+q_sub = {q: sm.Symbol(q.__class__.__name__) for q in sys.coordinates}
+
+u_sub = {u: sm.Symbol(u.__class__.__name__) for u in sys.speeds}
+
+expr = expr.subs(q_sub).subs(u_sub)
+
+f_nonpar = ufuncify_matrix(list(sys.constants_symbols) + list(q_sub.values()) +
+                           list(u_sub.values()), expr, parallel=False)
+
+f_par = ufuncify_matrix(list(sys.constants_symbols) + list(q_sub.values()) +
+                        list(u_sub.values()), expr, parallel=True)
+
+start_time = timeit.default_timer()
+f_nonpar(res, *args)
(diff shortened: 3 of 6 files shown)


moorepants:
@jcrist Thought you might be interested in this. I'm not sure why the parallel version doesn't gain any computation time. Maybe it would if the expressions were much more complicated.

moorepants:
This code has a much more computationally intensize expression:
```
In [1]: from opty.utils import ufuncify_matrix

In [2]: from pydy.models import n_link_pendulum_on_cart
/home/moorepants/src/pydy/pydy/codegen/code.py:18: DeprecationWarning: This module, 'pydy.codgen.code', is deprecated. The function 'generate_ode_function' can be found in the 'pydy.codegen.ode_function_generator' module. 'CythonGenerator' has been removed, use 'pydy.codegen.cython_code.CythonMatrixGenerator' instead.
  DeprecationWarning)

In [3]: import numpy as np

In [4]: import sympy as sm

In [5]: sys = n_link_pendulum_on_cart(10, False, False)

In [6]: c = np.random.random(int(1e7))

In [7]: args = [c for i in range(len(sys.constants_symbols + sys.coordinates + sys.speeds))]

In [8]: res = np.empty((int(1e7), 11))

In [9]: udots = sm.Matrix([5.0 for u in sys.speeds])

In [10]: expr = sys.eom_method.mass_matrix * udots - sys.eom_method.forcing

In [11]: expr = expr.subs(q_sub).subs(u_sub)
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-11-9d8b957997a8> in <module>()
----> 1 expr = expr.subs(q_sub).subs(u_sub)

NameError: name 'q_sub' is not defined

In [12]: q_sub = {q : sm.Symbol(q.__class__.__name__) for q in sys.coordinates}

In [13]: u_sub = {u : sm.Symbol(u.__class__.__name__) for u in sys.speeds}

In [14]: expr = expr.subs(q_sub).subs(u_sub)

In [15]: f = ufuncify_matrix(sys.constants_symbols + q_sub.values() + u_sub.values(), expr, tmp_dir='parallely')
```
Single threaded version:
```
In [16]: %timeit f(res, *args)
1 loops, best of 3: 17.3 s per loop
```
Parallel version:
```
In [22]: %timeit f(res, *args)
1 loops, best of 3: 6.93 s per loop
```
So that gives a 2.5X speed up using 4 cores instead of 1.

moorepants:
Maybe this should be optional because so that small expressions can use a single thread and large expressions multicores. I could check the "length" of the expression and decide for the user.

moorepants:
Funny I already implemented this in the parallel-jac branch.

jcrist:
Looks interesting. I assume that the overhead of the parallel evaluation is too large for smaller problems. Setting it as a user option would be one way. Checking the length (`count_ops`) is another. However, for very large models this is also extremely slow (the bike model took 30 minutes to count-ops for me).

A better way may be to time a single run of the calculation, and have some cutoff value that determines if it's run in parallel or not. Because this will vary per machine, you could do some kind of benchmark on installation, and cache that value with the install?

moorepants:
@jcrist Thanks for the ideas. I like the timing idea. In this package I may just set it as a user option. Right now this thing is just a workhorse for my research. I merge it into pydy though, so we can have a sympy backend optimization tool box.

moorepants:
Note that openmp is in conda-forge for all OSs. So this could be added, but it probably should still be treated as an optional dependency.

moorepants:
I just ran this on my 16 core, 2 thread per core server and got about a ~10X speed up (19s vs 1.9s). So this is pretty nice, especially for high core machines and very large equations of motion. It is now merged and can be discussed in the JOSS paper.

PR #40, opened 2016-05-19
Huawei wang homotopy direct collocation

HuaweiWang:
Added the extra term of homotopy method into direct collocation codes. 

But only added into "midpoint" integration method right now, as I am not very clear about the codes of "back euler" integration method.

Please check whether I am doing wrong, and give me some comments if possible.


moorepants:
@HuaweiWang Do you know what unit tests are? It will be good to write some unit tests first for the code additions. The unit tests will codify the expected behavior. You can see some examples of the tests I've already written for basic examples in the tests directory. Do you have a simple problem we could use for a test, like solving the pendulum parameter id problem with homotopy?

tvdbogert:
@moorepants: is it a good idea to test this with the vyasarayani2011.py in the examples folder?

HuaweiWang:
I only know a little bit unit test stuff. Yes I just thinking about use vyasarayani's example too

HuaweiWang:
I would like to learn and have a try in unit tests, how about we use vyasarayani's example ? @moorepants

moorepants:
Yes, we can use that example to build out a test. That's basically what I've already done for the current tests.

opty/direct_collocation.py
```diff
@@ -166,6 +172,13 @@ def __init__(self, equations_of_motion, state_symbols,
             ndarrays of floats of shape(N,). Any time varying parameters in
             the equations of motion not provided in this dictionary will
             become free trajectories optimization variables.
+        known_experiment_map : dictionary, optional
+            A dictionary that maps the state SymPy functions of time to
+            ndarrays of floats of shape(N,4). Any time varying parameters in
+            the equations of motion not provided in this dictionary will
+            become free trajectories optimization variables.
+	    Lamda: float
+	        A parameter in homotopy method that controls the change of motion equations.
```
moorepants: We should be careful with the name of this. `lambda` is a reserved word in python, it is a built in function. And sympy has `Lambda` which is a symbolic version of the Python `lambda` and creates a symbolic function. Maybe a variable name like `homotopy_control` or something similar is better.

opty/direct_collocation.py
```diff
@@ -188,7 +201,24 @@ def __init__(self, equations_of_motion, state_symbols,
...
 
         """
-        self.eom = equations_of_motion
+        Lamda_Matrix = sym.Matrix([[0,0,0,0], [0,0,0,0], [0,0,Lamda,0],[0,0,0,Lamda]])
+	    """
+	    define matrix Lamda =   [0 0 0     0    ]
+			            [0 0 0     0    ]
+ 			            [0 0 lamda 0    ]
+			            [0 0 0     lamda]
+	    """
```
moorepants: I try to follow the Python PEP8 style guide: https://www.python.org/dev/peps/pep-0008/ for all the code. Please read over that to help with consistent formatting.

Also, you can setup your text editor or IDE to automatically detect/fix these kinds of things.

opty/direct_collocation.py
```diff
@@ -205,6 +235,8 @@ def __init__(self, equations_of_motion, state_symbols,
 
         self.known_parameter_map = known_parameter_map
         self.known_trajectory_map = known_trajectory_map
+        
+        self.know_experiment_map = known_experiment_map " get input experiment data into self
```
moorepants: Use the `#` for single line and short comments.

opty/direct_collocation.py
```diff
@@ -166,6 +172,13 @@ def __init__(self, equations_of_motion, state_symbols,
             ndarrays of floats of shape(N,). Any time varying parameters in
             the equations of motion not provided in this dictionary will
             become free trajectories optimization variables.
+        known_experiment_map : dictionary, optional
```
moorepants: It isn't clear to me what this is.

opty/direct_collocation.py
```diff
@@ -133,10 +133,16 @@ class ConstraintCollocator(object):
     """
 
     time_interval_symbol = sym.Symbol('h', real=True)
+    expeirment_data_symbol = sym.Symbol('e', real=True)
```
moorepants: spelling

HuaweiWang: Thanks for the check, I will adjust them.

opty/direct_collocation.py
```diff
@@ -166,6 +172,15 @@ def __init__(self, equations_of_motion, state_symbols,
             ndarrays of floats of shape(N,). Any time varying parameters in
             the equations of motion not provided in this dictionary will
             become free trajectories optimization variables.
+        known_experiment_map : dictionary, optional
+            A dictionary that maps the state SymPy functions of time to
+            ndarrays of floats of shape(N,4). Any time varying parameters in
+            the equations of motion not provided in this dictionary will
+            become free trajectories optimization variables.
+	Lambda: float
```
moorepants: This should be an optional flag. If the user supplies the lambda value then it will use the homotopy method and if they don't it should do what the previous behavior was.

Or the default of lambda should be a value that gives the previous behavior (no lambda in the dynamics).

Another approach here would be to make a new Collocator class (subclass this one) and then change out the relevant methods for the homotopy functionality.

opty/direct_collocation.py
moorepants: The `known_parameter_map` already maps numpy arrays and floats to symbols in the equations of motion. Why is a new dictionary needed?

Currently this class isn't aware of the concept of measurement data in a parameter id sense.

opty/direct_collocation.py
moorepants: What text editor/IDE do you use? You should find one that checks all this stuff for you automatically. I can give recommendations if you tell me what OS you are working on.

opty/direct_collocation.py
```diff
@@ -188,7 +203,30 @@ def __init__(self, equations_of_motion, state_symbols,
             to a directory here.
 
         """
-        self.eom = equations_of_motion
+        Lambda_Matrix = sym.Matrix([[0,0,0,0], [0,0,0,0], [0,0,Lambda,0],[0,0,0,Lambda]])
```
moorepants: Why is this of fixed size? Shouldn't it be a function of the number of states, number of inputs, etc?

opty/direct_collocation.py
HuaweiWang: This known_experiment_map is the experimental data we need use in the extra term and I want to write a explanation for that. I am not sure how to write this explanation, so I wrote this according to known_trajectory_map. (might be wrong)

opty/direct_collocation.py
```diff
@@ -188,7 +203,30 @@ def __init__(self, equations_of_motion, state_symbols,
...
+ 			             [0 0 lambda 0     ]
+			             [0 0 0      lambda]
+	    """
+        K_Matrix = sym.Matrix([[0,0,0,0], [0,0,0,0], [0,0,K,0],[0,0,0,K]])
+	    """
+	    define matrix Lambda =   [0 0 0 0 ]
+			             [0 0 0 0 ]
+ 			             [0 0 K 0 ]
+			             [0 0 0 K ]
+	    """
```
moorepants: I think these need to be something like:

```python
k_matrix = sym.zeros(2 * num_states)
k_matrix[num_states:, num_states:] = k * sym.eye(num_states)
```

opty/direct_collocation.py
HuaweiWang: Yes, Lambda is a optional float number. If Lambda = 0, the homotopy will not applied. Yes, we could create a new Collocator subclass, but as I am new in python, I am not sure whether I could correctly change out all the relevant methods for the homotopy functionality.

opty/direct_collocation.py
HuaweiWang: I am working on Ubuntu 14.04. I am struggling in the online editing, please
recommend me a text editor/IDE. Thanks.

opty/direct_collocation.py
HuaweiWang: Yes, your way is more general.

opty/direct_collocation.py
moorepants: Do you have time to google hangout for a bit right now?

opty/direct_collocation.py
```diff
@@ -137,6 +137,7 @@ class ConstraintCollocator(object):
     def __init__(self, equations_of_motion, state_symbols,
                  num_collocation_nodes, node_time_interval,
                  known_parameter_map={}, known_trajectory_map={},
+                 homotopy_control, tracing_dynamic_control,
```
moorepants: Please make these as keyword arguments with defaults settings: `homotopy=0`, `tracing=1`.

opty/direct_collocation.py
```diff
@@ -188,7 +195,13 @@ def __init__(self, equations_of_motion, state_symbols,
...
 
         """
-        self.eom = equations_of_motion
+        Lamda_matrix = sym.zeros(num_states)
+        Lamda_matrix[num_states/2:, num_states/2:] = homotopy_control * sym.eye(num_states/2)
+
+        k_matrix = sym.zeros(num_states)
+        k_matrix[num_states/2:, num_states/2:] = tracing_dynamic_control * sym.eye(num_states/2)
+	
+        # self.eom = equations_of_motion
```
moorepants: delete this line

opty/direct_collocation.py
```diff
@@ -205,13 +218,32 @@ def __init__(self, equations_of_motion, state_symbols,
...
         self.instance_constraints = instance_constraints
 
         self.num_constraints = self.num_states * (num_collocation_nodes - 1)
 
         self.tmp_dir = tmp_dir
 
+        self.eom = (1-Lamda_matrix)*equations_of_motion + Lamda_matrix
+		    *(self.state_derivative_symbols - K_matrix*
+		    (known_trajectory_map[self.num_know_trajectory_map 
+		    - self.num_states:]-self.state_symbols))
```
moorepants: I think this should be:

```python
self.eom = (1-Lamda_matrix)*equations_of_motion + Lamda_matrix *(self.state_derivative_symbols - K_matrix*
		  (sympy.Matrix([predefined symbols]) - self.num_states:]-self.state_symbols))
```

moorepants: If a given state is named `x1`, then we should create new symbols that mirror the states with a name `x_1_m`.

HuaweiWang:
This pull request is only for methods testing added in unit tests. It is not a final version. So there is no need to look at it @moorepants @tvdbogert

moorepants:
@HuaweiWang What's the status on this? Have you gotten the method to work?

HuaweiWang:
Hi,

I am still in Unit Testing right now. As I will take Qualify Exam in August
8th, I have few time working with the codes.

Hopefully I could pass the Qualify Exam and put more effort into the codes.

Will let you know If I have some progression.

Bests,

Huawei

moorepants:
Ok, sounds good. Good luck on your quals!

moorepants:
Closing, reopen if still interested.

Issue #62, opened 2017-10-19
Add list of references


moorepants:
From: https://github.com/openjournals/joss-reviews/issues/300#issuecomment-325696614
> A list of key references including a link to the software archive - References are not rendered in the paper.md file - No link to software archive - Other than Betts2010, the other citations appear to be just for dependencies. The paper should include citations to other software for direct colocation, and a comparison of feature sets, etc.

- [x] finish up the table of comparisons

moorepants:
- Casadi
- Gpops

moorepants:
@tvdbogert If you have any to add to this list here, feel free to. I'm not sure I'm that equipped to do a feature comparison, as I haven't used the other software. But I'll give it a shot.

moorepants:
@tvdbogert Your paragraph looks good. Let's add in the references.

tvdbogert:
References were added in paper.bib.  How can I get the rendered version of the list of references?

The comparison table still has gaps.  Which ones should I try to fill?

moorepants:
I'm not sure how they render the references. I would suggest using a reference manager like jabref, zotero, mendely, etc to export the references to see if they look ok.

moorepants:
Also, can you help me fill out the table in paper.md. You might know some of the things off hand or where the answer is more quickly.

moorepants:
I added a paper.tex file that can be compiled to check the bib entry renderings. I also opened paper.bib in jabref and let it format things and fixed a few errors.

tvdbogert:
Thanks.  Github's rendering of paper.md ends with the heading "References" and I was expecting to see the list of references there.  I don't even know what a .md file is supposed to do.

moorepants:
I think that the journal processing the md file and the bib file to create something like: http://joss.theoj.org/papers/050362b7e7691d2a5d0ebed8251bc01e

BTW, that paper is pretty thorough. I though you were just supposed to create an abstract of sorts.

tvdbogert:
That paper is a couple of paragraphs longer maybe.  I suggest we just address the reviewer comments and not get too ambitious.  As I mentioned earlier, when you try to be thorough, the expectations will rise and you get more criticism.

moorepants:
Agreed. I think the main remaining things are getting the one example working again (or removing it) and to add some of the mathematical description to the documentation.

moorepants:
@tvdbogert We are only missing information for DYNOPT and GESOP. Do you have info or should I research them?

[...]

Issue #61, opened 2017-10-19
Improve statement of need


moorepants:
From: https://github.com/openjournals/joss-reviews/issues/300#issuecomment-325696614
> A clear statement of need that illustrates the purpose of the software - Motivation should be expanded, why is this software useful, what targeted applications?

moorepants:
I think this is now addressed by #74.

Issue #59, opened 2017-10-19
Improve the high level summary


moorepants:
From: https://github.com/openjournals/joss-reviews/issues/300#issuecomment-325696614

> A summary describing the high-level functionality and purpose of the software for a diverse, non-specialist audience — current summary is written in highly technical language, I suggest a lay introduction followed by a more detailed technical description of the software that includes equations to clarify what types of problems are applicable for the software

moorepants:
@tvdbogert Would you say this is now addressed?

tvdbogert:
Yes, but there are still a few gaps in the table (in the rows you added).

moorepants:
This specific review comment didn't ask for the table, so I'm going to close this one.

PR #74, opened 2017-10-20
Improved paper.md and added comparison table.

moorepants:


moorepants:
This is addresing #59, #61, and #62.

PR #75, opened 2017-10-24
First shot at the theory. Fixes #60.

### LINKED ISSUE #60: Improve the theory explanation in the docs
From: https://github.com/openjournals/joss-reviews/issues/300#issuecomment-325696614
> — [optional] a full theory guide would be very beneficial for introducing the basic concepts of direct colocation. Although this type of information can be found in textbooks / papers (e.g. the Betts paper cited), but it can be very helpful to include it within the software paper, as the software authors typically have their own interpretation of the theory, and the equations can be better matched to the software implementation. For example, both FEBio and Simbody have excellent theory guides, which are arguably one of the main contributions of those works.

moorepants:


moorepants:
@tvdbogert Here is the start of a theory document for the documentation. Let me know if you think this is going in the right direction. It should translate pretty closely to the code.

moorepants:
Here is the current view:
screenshot-2017-10-26 theory opty 0 3 0 dev0 documentation

tvdbogert:
That is just the right amount of detail.  I like how it is general enough to deal with any combination of known and unknown input signals and parameters.

In the last equation, bounds are specified on u and p.  This should be r_u and p_u.

A couple of notation issues, no big deal for me, but reviewers may be confused.

1. When you first introduce J, it is a function of x, r, and p.  For x and r that should be a function of their trajectories.  One way to make that clear is by writing J as an integral of a cost rate L(x,(t),r(t),p).  Opty is more general in the type of cost function, so I am not sure how to write that properly.  Maybe integral of L is general enough if you consider that L may include delta functions.

2. The NLP formulation in the end uses x and r_u as unknowns, but those were earlier defined as state and unknown input at time t which is possibly confusing.  I personally prefer a notation that is closer to how it is actually implemented.  Define a vector X which stacks the x and r_u vectors at the collocation points, and the parameters p_u.  Then the NLP is simply defined as: minimize J(X), subject to F(X)=0 and L<=X<=U.  F can be defined as the stacked dynamics residuals f_i which you already defined earlier.

The comments about efficient computation of the Jacobian dF/dX are very important and it's good to end with that.  This is what Opty is all about.  Maybe include gradient dJ/dX in that comment also.

moorepants:
Ok, here is an update:

screenshot-2017-10-26 theory opty 0 3 0 dev0 documentation 1

moorepants:
@tvdbogert Here is the lasted version of the theory:
screenshot-2018-1-9 theory opty 0 3 0 dev0 documentation

What do you think? I'm going to merge soon and will adjust if needed later.

tvdbogert:
It's good and clear.  I corrected a type and committed directly to the master branch.

Other comments:

- At the end, I would want to continue some more and also formulate the final NLP, with a large vector of unknowns containing the unknown discretize-time trajectories and parameters, an optimization objective, and a large vector of constraints. And a reference to Betts, at least. Unless you already have that in the "paper" document.
- Which raises the other question: how do you divide the material between the theory and paper documents?

moorepants:
> At the end, I would want to continue some more and also formulate the final NLP, with a large vector of unknowns containing the unknown discretize-time trajectories and parameters, an optimization objective, and a large vector of constraints.

Isn't the equation right after "The above equations will create nm constraint equations and the optimization problem can formally be written as:" what your are describing?

> Which raises the other question: how do you divide the material between the theory and paper documents?

I don't think the "paper" for this journal is supposed to be that in-depth. For example here is their example paper: https://github.com/arfon/fidgit/blob/master/paper/paper.md

See "What should my paper contain?" here http://joss.theoj.org/about. Seems like we have more than enough in the paper. in fact, I may move the table to the documentation.

tvdbogert:
Oops, yes, you already described the NLP.

Betts reference may still be useful to add.

moorepants:
Ok, added Bett's reference to the theory page. This is done then. Thanks.

PR #86, opened 2020-11-16
Fix outdated kwarg for sm.external.import_module

ashander:
allows use of more recent sympy (possibly causing other problems...)


moorepants:
Not seeing any file changes.

moorepants:
This is a backwards incompatible change in sympy from this PR: https://github.com/sympy/sympy/pull/18244 (made this last January).

PR #88, opened 2020-11-17
Drop Py35 and add Py37/Py38 to CI tests.

moorepants:


moorepants:
There are still issues with Py27, just gonna merge.

Issue #100, opened 2023-08-23
vyasarayani2010 and park2004 examples fail with `double free or corruption (out)`


moorepants:
```
Python 3.8.17 | packaged by conda-forge | (default, Jun 16 2023, 07:06:00) 
Type 'copyright', 'credits' or 'license' for more information
IPython 8.12.2 -- An enhanced Interactive Python. Type '?' for help.

In [1]: run vyasarayani2011.py
Using noisy measurements for the trajectory initial guess and a random positive value for the parameter.

******************************************************************************
This program contains Ipopt, a library for large-scale nonlinear optimization.
 Ipopt is released as open source code under the Eclipse Public License (EPL).
         For more information visit http://projects.coin-or.org/Ipopt
******************************************************************************

This is Ipopt version 3.12.13, running with linear solver mumps.
NOTE: Other linear solvers might be more efficient (see Ipopt documentation).

Number of nonzeros in equality constraint Jacobian...:    49990
Number of nonzeros in inequality constraint Jacobian.:        0
Number of nonzeros in Lagrangian Hessian.............:        0

double free or corruption (out)
```

I don't think this is an issue with opty because the obj, constraints and their derivatives seem to evaluate just fine:

```
Python 3.8.17 | packaged by conda-forge | (default, Jun 16 2023, 07:06:00) 
Type 'copyright', 'credits' or 'license' for more information
IPython 8.12.2 -- An enhanced Interactive Python. Type '?' for help.

In [1]: run vyasarayani2011.py
Using noisy measurements for the trajectory initial guess and a random positive value for the parameter.
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
File ~/src/opty/examples/vyasarayani2011.py:200
    195 parser.add_argument('-p', '--plot', action="store_true",
    196                     help="Show result plots.")
    198 args = parser.parse_args()
--> 200 main(args.initialguess, do_plot=args.plot)

File ~/src/opty/examples/vyasarayani2011.py:118, in main(initial_guess, do_plot)
    115     # Known solution as initial guess.
    116     initial_guess = np.hstack((y1_meas, y2_meas, 10.0))
--> 118 pause
    120 # Find the optimal solution.
    121 solution, info = prob.solve(initial_guess)

NameError: name 'pause' is not defined
> /home/moorepants/src/opty/examples/vyasarayani2011.py(118)main()
    116         initial_guess = np.hstack((y1_meas, y2_meas, 10.0))
    117 
[...]

PR #101, opened 2023-08-23
Update to use correct name for cyipopt package and Problem class

brocksam:
Since the release of CyIpopt v1.0, the package name has changed to `cyipopt` from `ipopt`. Also the name of the main class has changed to `Problem` from `problem`. This PR updates these two things to stop the warnings being raised by CyIpopt.


moorepants:
Python 3.6 and 3.7 had the conda process killed (timed out?). Python 3.8 and 3.9 had errors similar to #100 .

PR #115, opened 2024-02-02
Function to generate object and objective gradient evaluator functions

tjstienstra:
This PR proposes a utility function to allow one to specify an analytic expression as the objective function.
It is a relatively simple function with one major limitation, namely that one cannot use time-dependent and time-independent variables in the objective function.

Additionally, I noticed while opening this PR that #31 aims to implement a similar utility in an entirely different way. Will have to look into that one to see if it would be a better solution.

### DIFF
--- a/opty/utils.py
+++ b/opty/utils.py
@@ -5,7 +5,6 @@ import sys
 import shutil
 import tempfile
 import subprocess
-import textwrap
 import importlib
 from functools import wraps
 import warnings
@@ -231,8 +230,9 @@ def parse_free(free, n, q, N):
     return free_states, free_specified, free_constants
 
 
-def create_objective_function(objective, state_symbols, input_symbols,
-                              unknown_symbols, N, node_time_interval=1.0):
+def create_objective_function(
+    objective, state_symbols, input_symbols, unknown_symbols,
+    N, node_time_interval=1.0, integration_method="backward euler"):
     """Creates function to evaluate the objective and objective gradient.
 
     Parameters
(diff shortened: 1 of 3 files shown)


tjstienstra:
Okay, based on the other PR. I would say that this is a fun function for quite a few cases. However, we optimally want to define the objective expression as a continuous function, possibly with an integral.
```python
import sympy as sm
import sympy.physics.mechanics as me

x, v, f = me.dynamicsymbols("x v f")
t = me.dynamicsymbols._t
m = sm.symbols("m")
integration_method = "backward euler"
node_times = np.linspace(0, 2, 101)  # duration=2.0 ; N=101 ; interval_value=0.02

obj_expr = m**2 + sm.Integral(f**2, (t, 0.5, 1.5))
obj, obj_grad = create_objective_function(obj_expr, (s, v), (f,), (m,), integration_method, node_times)

# Should give something like
# free[303] is just the value of m
# free[228] is f.subs(t, 0.52) (we are using backward euler), therefore it is not f.subs(t, 0.5)
# free[278] is f.subs(t, 1.50)
def obj(free):
    return free[303] ** 2 + np.sum(free[228:278] ** 2)

def obj_grad(free):
    return np.hstack((np.zeros(101), np.zeros(101), np.zeros(25), 2 * free[228:278], np.zeros(26), 2 * free[303]))
```
Not yet fully sure if this is the ideal method to construct obj_grad, but it is the version which looks most like what this PR currently tried to do.

tjstienstra:
A temporary simplification would be to just ignore the bounds of the integral for now, i.e. `assert integral.limits == ((t,),)`

tjstienstra:
Here is a fun objective, that will give even when trying to keep things simple quite a few problems.
```python
import sympy as sm
import sympy.physics.mechanics as me

x, v, f = me.dynamicsymbols("x v f")
t = me.dynamicsymbols._t
m, k, c = sm.symbols("m k c")
integration_method = "backward euler"
node_times = np.linspace(0, 2, 11)

states = sm.ImmutableDenseMatrix([x, v])
inputs = sm.ImmutableMatrix([f])
params = sm.ImmutableMatrix([c, k, m])
n, q, r = states.shape[0], inputs.shape[0], params.shape[0]

obj = m**2 + sm.Integral(f**2, (t,)) + sm.Integral(c**2 * f, (t,)) + sm.Integral(k**2 , (t,))
obj_grad = sm.ImmutableMatrix([obj]).jacobian(states.col_join(inputs).col_join(params))
```

tjstienstra:
@moorepants I think I figured out a method to combine it all. It is not the most beautiful function I have ever written, but it seems to do the job as the tests show.

moorepants:
https://github.com/csu-hmc/opty/pull/116 fixes the test failuers (pytest 8 just came out with a deprecation). You can merge master.

opty/utils.py
```diff
@@ -229,6 +230,140 @@ def parse_free(free, n, q, N):
...
+        euler".
+
+    """
+    def lambdify_function(expr, multiplication_array, take_sum):
+        if take_sum:
+            def integration_function(x):
+                return node_time_interval * np.sum(x * multiplication_array)
+        else:
+            def integration_function(x):
+                return node_time_interval * x * multiplication_array
```
moorepants, line 265: Did you try an see if lambidy(module=scipy) works to convert these integrals to the proper scipy function?

opty/utils.py
```diff
@@ -229,6 +230,140 @@ def parse_free(free, n, q, N):
...
+    params = sm.ImmutableMatrix(sort_sympy(unknown_symbols))
+    if states.shape[1] > 1 or inputs.shape[1] > 1 or params.shape[1] > 1:
+        raise ValueError(
+            'The state, input, and unknown symbols must be column matrices.')
+    n, q = states.shape[0], inputs.shape[0]
+    i_idx, r_idx = n * N, (n + q) * N
+
+    # Compute analytical gradient of the objective function
+    objective_grad = sm.ImmutableMatrix([objective]).jacobian(
+        states[:] + inputs[:] + params[:])
```
moorepants, line 283: Just as note for me to see what this looks like:

```python
import sympy as sm
a, t = sm.symbols('a, t')
f, g = sm.Function('f')(t), sm.Function('g')(t)
expr = f**2 + a*sm.sin(g)
int_expr = sm.Integral(expr, t)
sm.ImmutableMatrix([int_expr]).jacobian([a, f, g])
```

```
Matrix([[Integral(sin(g(t)), t), Integral(2*f(t), t), Integral(a*cos(g(t)), t)]])
```

opty/utils.py
moorepants: ah `NotImplementedError: Only definite integrals are supported`

opty/utils.py
```diff
@@ -229,6 +230,140 @@ def parse_free(free, n, q, N):
...
+        ) + tuple(objective_grad[n + q:])
+
+    # Define evaluation functions based on the integration method
+    if integration_method == "backward euler":
+        obj_expr_eval = lambdify_function(
+            objective, np.hstack((0, np.ones(N - 1))), True)
+        obj_grad_time_expr_eval = lambdify_function(
+            objective_grad[:n + q], np.hstack((0, np.ones(N - 1))), False)
+        obj_grad_param_expr_eval = lambdify_function(
+            objective_grad[n + q:], np.hstack((0, np.ones(N - 1))), True)
```
moorepants, line 306: I don't follow these lines perfectly. What is the reason to force the sum or not?

opty/utils.py
tjstienstra: I did not check, but thanks for checking. I have never experimented with what the scipy printer does. I should do that some time.

opty/utils.py
tjstienstra: Yeah, it all feels a bit like a cheat.
So, the multiplications with arrays make sure you always have the right shape and they also give weights to the nodes. E.g. the first node doesn't do anything in backward Euler.
For the gradient, we in general of course do not want to sum the states, as we actually want to compute the separate pieces. However, if you for some reason got an integral over a state or input as the gradient of an unknown, then we do need to just integrate it, i.e. take the sum.

P.S. The explanation is a bit messy. I can try to make it more clear.

opty/utils.py
moorepants: I think it works with definite integrals. I'm wondering if we need these special cases in this function it integrals already work in lambdify.

tjstienstra: We would still like to specify the integration method it uses.

tjstienstra: When using `scipy` with properly set limits, then it uses [quad](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.quad.html)

tjstienstra: It should be possible to support integrals with limits. It can be done by updating the multiplication array. Only bit difficult part is of course that you may have different limits for different integrals in the same function.

opty/utils.py
```diff
@@ -229,6 +230,136 @@ def parse_free(free, n, q, N):
...
+def create_objective_function(
+    objective, state_symbols, input_symbols, unknown_symbols,
+    N, node_time_interval=1.0, integration_method="backward euler"):
+    """Creates function to evaluate the objective and objective gradient.
+
+    Parameters
+    ----------
+    objective : sympy.Expr
+        The objective function to be minimized, which is a function of the
+        states and inputs.
```
moorepants, line 242: Does this work with any kind of objective function? If not, can you say what its limits are?

tjstienstra: Added some clarification. Additionally, I switched to crawling the expression tree myself instead of using replace with a wildcard. This allows to raise NotImplementedErrors for all integral versions which are not supported.

tjstienstra: Additionally, it is also a bit more future proof and faster, as one can now easily access the limits for parsing purposes.

moorepants:
You happy with this? I think it is good to go in if you do.

tjstienstra:
I think that it is good for now. Unless you're planning to add a lot of advanced algorithms  soon, I really don't see the need to make it more advanced. Like the internals may change. The only thing is that the keyword arguments expect fixed time stepping.
All other limitations seem to be covered using errors.

moorepants:
Ok. let's merge. Nice addition!

PR #120, opened 2024-02-07
Use the create_objective_function in the pendulum swing up example.

moorepants:


examples/pendulum_swing_up.py
```diff
@@ -43,18 +43,11 @@
...
-
-def obj_grad(free):
-    grad = np.zeros_like(free)
-    grad[2 * num_nodes:] = 2.0 * interval_value * free[2 * num_nodes:]
-    return grad
+obj_func = sym.Integral(T(t)**2, t)
+obj, obj_grad = create_objective_function(obj_func, state_symbols,
+                                          specified_symbols, tuple(),
+                                          num_nodes,
+                                          node_time_interval=interval_value)
```
moorepants, line 50: Seems like `node_time_interval` should be an arg not a kwarg. I don't think there is a case when you wouldn't need to set that. Also, should specified symbols and unknown parameters be kwargs since you make not have them.

PR #130, opened 2024-02-28
Show args and kwargs in Problem init explicitly

### LINKED ISSUE #129: Make args and kwargs in Problem explicit (so default values are shown)
This is optional, as I read the docs. _midpoint_ and _backward_ euler are available.
Which one is the default value? (I would assume, that the default method is normally better)

Thanks!

tjstienstra:
Fixes #129

Note that the docstring is already being updated by the `_doc_inherit` decorator.

### DIFF
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -91,7 +91,12 @@ class Problem(cyipopt.Problem):
     INF = 10e19
 
     @_doc_inherit
-    def __init__(self, obj, obj_grad, *args, **kwargs):
+    def __init__(self, obj, obj_grad, equations_of_motion, state_symbols,
+                 num_collocation_nodes, node_time_interval,
+                 known_parameter_map={}, known_trajectory_map={},
+                 instance_constraints=None, time_symbol=None, tmp_dir=None,
+                 integration_method='backward euler', parallel=False,
+                 bounds=None):
         """
 
         Parameters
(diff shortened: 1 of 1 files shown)


opty/direct_collocation.py
```diff
@@ -110,10 +115,14 @@ def __init__(self, obj, obj_grad, *args, **kwargs):
...
 
-        self.bounds = kwargs.pop('bounds', None)
-
-        self.collocator = ConstraintCollocator(*args, **kwargs)
+        self.collocator = ConstraintCollocator(
+            equations_of_motion, state_symbols, num_collocation_nodes,
+            node_time_interval, known_parameter_map, known_trajectory_map,
+            instance_constraints, time_symbol, tmp_dir, integration_method,
+            parallel
+            )
```
moorepants, line 123: Why would this one have rendered the full list correctly if it was `*args, **kwargs`?

opty/direct_collocation.py
```diff
@@ -110,10 +115,14 @@ def __init__(self, obj, obj_grad, *args, **kwargs):
...
 
+        self.bounds = bounds
```
moorepants, line 125: Where does bounds come from?

opty/direct_collocation.py
moorepants: Sorry, I thought that was the class definition.

moorepants:
LGTM

PR #149, opened 2024-05-03
Added option to show Cython compilation output and display more informative error.

moorepants:


moorepants:
@tjstienstra Any thoughts on this? Not sure if you are interested in reviewing things here.

tjstienstra:
Haven't tested it, but it looks like a really nice feature. In terms of review, I can look at it if you like. As you currently always capture the output you may consider providing that in the error message as well.

moorepants:
> As you currently always capture the output you may consider providing that in the error message as well.

Good idea.

moorepants:
I'll go with this.

PR #150, opened 2024-05-05
Implements support for variable duration solutions

### LINKED ISSUE #19: Implement h as a free variable
The time between nodes is currently a fixed value. We can solve variable duration problems if I enable it as a free variable. This may be necessary for getting the ackermann example to work.

moorepants:
Fixes #19

This PR allows a user to pass in a symbol as the node interval value instead of a float and then specify instance constraints in terms of node index instead of time. This adds the node interval value as a free optimization variable (as the last element) and the resulting constraint equations and Jacobian takes into account the variable interval. I demonstrate it on the pendulum swing up problem.

- [x] Allow user to set bounds on the node time interval.
- [x] Update `parse_free` to deal with having h as the last term.
- [x] Make example problem for variable duration.
- [x] Show how to use this in the documentation examples.

### DIFF
--- a/examples/pendulum_swing_up_variable_duration.py
+++ b/examples/pendulum_swing_up_variable_duration.py
@@ -59,10 +59,10 @@ def obj_grad(free):
 
 # Specify the symbolic instance constraints, i.e. initial and end conditions
 # using node numbers 1 to N.
-instance_constraints = (theta(1),
-                        theta(num_nodes) - target_angle,
-                        omega(1),
-                        omega(num_nodes))
+instance_constraints = (theta(1*h),
+                        theta(num_nodes*h) - target_angle,
+                        omega(1*h),
+                        omega(num_nodes*h))
 
 # Create an optimization problem.
 prob = Problem(obj, obj_grad, eom, state_symbols, num_nodes, h,
--- a/examples/pendulum_swing_up_variable_duration.py
+++ b/examples/pendulum_swing_up_variable_duration.py
@@ -1,0 +1,122 @@
+"""This solves the simple pendulum swing up problem presented here:
+
+http://hmc.csuohio.edu/resources/human-motion-seminar-jan-23-2014
+
+A simple pendulum is controlled by a torque at its joint. The goal is to
+swing the pendulum from its rest equilibrium to a target angle by minimizing
+the energy used to do so.
+
+"""
+
+from collections import OrderedDict
+
+import numpy as np
+import sympy as sym
+from opty.direct_collocation import Problem
+from opty.utils import building_docs, create_objective_function
+import matplotlib.pyplot as plt
+import matplotlib.animation as animation
+
+target_angle = np.pi
+num_nodes = 500
+save_animation = False
+
+interval_value = sym.symbols('h')
+
+# Symbolic equations of motion
+I, m, g, d, t = sym.symbols('I, m, g, d, t')
+theta, omega, T = sym.symbols('theta, omega, T', cls=sym.Function)
+
+state_symbols = (theta(t), omega(t))
+constant_symbols = (I, m, g, d)
+specified_symbols = (T(t),)
+
+eom = sym.Matrix([theta(t).diff() - omega(t),
+                  I * omega(t).diff() + m * g * d * sym.sin(theta(t)) - T(t)])
+
+# Specify the known system parameters.
+par_map = OrderedDict()
+par_map[I] = 1.0
+par_map[m] = 1.0
+par_map[g] = 9.81
+par_map[d] = 1.0
+
+
+# Specify the objective function and it's gradient.
+def obj(free):
+    """Minimize the sum of the squares of the control torque."""
+    T = free[2 * num_nodes:]
+    return free[-1] * np.sum(T**2)
+
+
+def obj_grad(free):
+    T = free[2 * num_nodes:]
+    grad = np.zeros_like(free)
+    grad[2 * num_nodes:] = 2.0 * free[-1] * free[2 * num_nodes:]
+    grad[-1] = np.sum(T**2)
+    return grad
+
+
+# Specify the symbolic instance constraints, i.e. initial and end
+# conditions.
+instance_constraints = (theta(1),
+                        theta(num_nodes) - target_angle,
+                        omega(1),
+                        omega(num_nodes))
+
+# Create an optimization problem.
+prob = Problem(obj, obj_grad, eom, state_symbols, num_nodes, interval_value,
+               known_parameter_map=par_map,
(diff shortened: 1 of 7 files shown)


moorepants:
@tjstienstra, if you're interested in reviewing this, it is much appreciated. I've only tried it out on the example pendulum problem here, nothing more complicated yet.

moorepants:
If you have any comments on the API for the instance constraints, I'd be interested. The basic idea is that if you have fixed duration problems you give something like `x(1.23)` where the argument is time and then I find the nearest node index. If you have a variable duration, then you give `x(4)` as the 4th node as the argument. I'm not sure this is the best approach.

moorepants:
Variable duration pendulum swing up solution:
image

examples/pendulum_swing_up_variable_duration.py
```diff
@@ -0,0 +1,120 @@
...
+    grad[-1] = np.sum(T**2)
+    return grad
+
+
+# Specify the symbolic instance constraints, i.e. initial and end conditions
+# using node numbers 1 to N.
+instance_constraints = (theta(1),
+                        theta(num_nodes) - target_angle,
+                        omega(1),
+                        omega(num_nodes))
```
moorepants, line 65: When providing the node index here instead of time, we could ask for node numbers 0  to N - 1 or 1 to N.

I am a bit unsure which choice is the better. I guess it doesn't matter as long as it is clearly stated in the documentation.

tjstienstra: When referring to index I would say that from 0 is more logical.

tjstienstra:
> If you have any comments on the API for the instance constraints, I'd be interested.

I would have a preference not to use round brackets substituting t for indices. That may not just be confusing at first, especially since we normally find the index for the supplied time value, but it will also cause problems for possible future features.

The closest solution would be to use a multiple of `h`, e.g. `4*h` when selecting the fourth node.

tjstienstra:
One of my main curiosities is also what would the API look like if variable time steps were supported?

examples/pendulum_swing_up_variable_duration.py
```diff
@@ -0,0 +1,120 @@
...
+
+# Specify the known system parameters.
+par_map = OrderedDict()
+par_map[m] = 1.0
+par_map[g] = 9.81
+par_map[d] = 1.0
+
+
+# Specify the objective function and it's gradient.
+def obj(free):
```
tjstienstra, line 46: Why not use the `create_objective_function`?

moorepants: I wanted to write it out to see how the new variable affected the equations and did not know if `create_objective_function()` supported `h` as a symbol.

moorepants:
> I would have a preference not to use round brackets substituting t for indices.

I suspect I could support something like `theta[0]` but that could take considerably use more work. I'm using (or abusing) SymPy `Function` behavior. I guess you could create an `sympy.Indexed`.

> The closest solution would be to use a multiple of h, e.g. 4*h when selecting the fourth node.

I originally wrote it like this and then realized it was much simpler to ask the node number from the user, so switched to it.

moorepants:
> One of my main curiosities is also what would the API look like if variable time steps were supported?

I guess the API wouldn't really change other than `free` being made longer with all the h values. We could have a flag `fixed_interval_size` or something.

examples/pendulum_swing_up_variable_duration.py
tjstienstra: Not yet but we can easily make it support it (will have to check ones more). But I think that the only difference is that we should change the default keyword argument to an integer (bit nicer if someone doesn't specify it) and we have to multiply the objective with the node_time_interval and remove it from the functions where it manually multiplies the solution with the node time interval.

tjstienstra:
> I suspect I could support something like theta[0] but that could take considerably use more work.

Yeah, that was the other thing I thought of.

> The closest solution would be to use a multiple of h, e.g. 4*h when selecting the fourth node.\n\nI originally wrote it like this and then realized it was much simpler to ask the node number from the user, so switched to it.

I would prefer the additional difficulty, because it is clearer and mathematically correct. t=4h for the fifth node.

Another solution is to create a boolean keyword argument to interpret supplied time values as indices.

tjstienstra:
> > One of my main curiosities is also what would the API look like if variable time steps were supported?
> 
> I guess the API wouldn't really change other than `free` being made longer with all the h values. We could have a flag `fixed_interval_size` or something.

I was more thinking toward supplying an iterable as node_time_interval, which may include multiples of `h`.

moorepants:
> I was more thinking toward supplying an iterable as node_time_interval, which may include multiples of h.

I'm not sure why you'd want to do that. What is the reason that you'd want to specify unequal proportional spacing? Wouldn't it mean you are trying to guess where the solution is stiff (smaller spacing in stiff areas) without knowing the solution?

examples/pendulum_swing_up_variable_duration.py
moorepants: I'll let you update it after I merge this. Or I could merge in changes onto this branch if you want.

tjstienstra:
> > I was more thinking toward supplying an iterable as node_time_interval, which may include multiples of h.
> 
> I'm not sure why you'd want to do that. What is the reason that you'd want to specify unequal proportional spacing? Wouldn't it mean you are trying to guess where the solution is stiff (smaller spacing in stiff areas) without knowing the solution?

If a problem is partially reusable then we could start iterating to improve the solution. A bit like what pycollo does only then way simpler. In several problems you know where you want a bit higher resolution and where you do not need that many points.

moorepants:
I see, yes, if you have a sense of the solution you could proportional time to reflect that.

moorepants:
I switched to integer multiples of the time interval symbol.

examples/pendulum_swing_up_variable_duration.py
```diff
@@ -0,0 +1,120 @@
...
+    grad[-1] = np.sum(T**2)
+    return grad
+
+
+# Specify the symbolic instance constraints, i.e. initial and end conditions
+# using node numbers 1 to N.
+instance_constraints = (theta(1*h),
+                        theta(num_nodes*h) - target_angle,
+                        omega(1*h),
+                        omega(num_nodes*h))
```
moorepants, line 65: The other option for this is:

```python
instance_constraints = (theta(0*h),
                        theta((num_nodes - 1)*h) - target_angle,
                        omega(0*h),
                        omega((num_nodes - 1)*h))
```

Which is clunkier and `0*h` will just evaluate to `0`, needing a special case to detect it.

PR #170, opened 2024-06-22
Switched to Cython memoryviews

moorepants:


moorepants:
Switching to memoryviews didn't fix the issue with Numpy 2.0. I'm thinking that Numpy 2.0 has some variables declared in their new headers, like `I` that clash with variables I use in the examples.

PR #175, opened 2024-06-25
Try different way to get the encoding on Windows.

moorepants:


Peter230655:
Let me know, when I should check any new version, please.

moorepants:
`locale.getencoding()` is new in Python 3.11. So I can't use that solution for Python 3.8-3.10.

PR #195, opened 2024-07-24
Plot sliding block 2

Peter230655:
code line correct length, ran with opty-dev environment, only twi objective functions for speed.


moorepants:
Title has incorrect formatting:
image

All plots display together and which makes them tiny and unreadable:
image

examples-gallery/plot_sliding_block.py
```diff
@@ -0,0 +1,333 @@
...
+from collections import OrderedDict
+import numpy as np
+import sympy as sm
+from opty.direct_collocation import Problem
+import matplotlib.pyplot as plt
+import matplotlib as mp
+from matplotlib.animation import FuncAnimation
+from copy import deepcopy
+# %%
+# define the road
```
moorepants, line 43: Use complete sentences with capitalization and punctuation.

examples-gallery/plot_sliding_block.py
```diff
@@ -0,0 +1,333 @@
...
+
+m, g, reibung = sm.symbols('m, g, reibung')
+a, b = sm.symbols('a b')
+
+P0.set_pos(O, x * N.x + strasse(x, a, b) * N.y)
+P0.set_vel(N, ux * N.x + strasse(x, a, b).diff(x)*ux * N.y)
+BODY = [me.Particle('P0', P0, m)]
+
+# The control force and the friction are acting in the direction of
+# the tangent at the street at the point whre the particle is.
```
moorepants, line 67: These show up as code comments, Make them "cells" so that they display as rendered text.

examples-gallery/plot_sliding_block.py
```diff
@@ -0,0 +1,333 @@
+# %%
+""""
+Sliding a Block on a Road
+=========================
+A block, modeled as a particle is sliding on a road to cross a
+hill. The block is subject to gravity and speed dependent
+friction.
+Gravity points in the negative Y direction.
+A force tangential to the road is applied to the block.
+Three objective functions to me minimized may be selected:
```
moorepants, line 10:
```suggestion
Three objective functions to be minimized may be selected:
```

examples-gallery/plot_sliding_block.py
```diff
@@ -0,0 +1,333 @@
...
+
+- selektion = 0: time to reach the end point is minimized
+- selektion = 1: time to reach the end point and the energy are minimized.
+- selektion = 2: energy consumed is minimized.
```
moorepants, line 14: Only two animations are displayed, shouldn't there be a result per selection?

Peter230655: I only picked two of the three, to speed up the running time. Should I go back to all three?

examples-gallery/plot_sliding_block.py
```diff
@@ -0,0 +1,333 @@
...
+=========================
+A block, modeled as a particle is sliding on a road to cross a
+hill. The block is subject to gravity and speed dependent
+friction.
+Gravity points in the negative Y direction.
+A force tangential to the road is applied to the block.
+Three objective functions to me minimized may be selected:
+
+- selektion = 0: time to reach the end point is minimized
+- selektion = 1: time to reach the end point and the energy are minimized.
```
moorepants, line 13: This is ill-defined, no? If you try to minimize time and minimize energy, those are at odds, so it would only depend on what you weight more than the other, right?

Peter230655: YES, in line 109 I define the variable verstaerkung = 2.e5. (The numerical value I picked by trial and error to make it look nice) This is the relative weight of the 'time' vs. energy.
(Verstaerkung = weight or strength in German. I often use German words, to make sure I do not accidentally overwrite Python key words)

examples-gallery/plot_sliding_block.py
Peter230655: You mean something like: _The function below defines the road on which the block slides_ ?

examples-gallery/plot_sliding_block.py
Peter230655: meaning like:
# %%
# The control force.....
# the tangent...
?

examples-gallery/plot_sliding_block.py
moorepants: I'd remove option 1, as 0 and 2 are sufficient for an example and option 1 is totally dependent on how you weight the two components to the objective function.

examples-gallery/plot_sliding_block.py
moorepants: Cells are delimited by the `# %%`

Peter230655: That's what I meant. Typo again! :-(

examples-gallery/plot_sliding_block.py
Peter230655: I found this 'relative weight' quite interesting mathematically - but I agree, it distracts from the example.
I will make the corrections this evening and make a new pull request.
It did come in a branch like it should?

examples-gallery/plot_sliding_block.py
```diff
@@ -0,0 +1,333 @@
+# %%
+""""
```
moorepants, line 2:
```suggestion
"""
```

Peter230655:
Needs many more changes. I will make them in the next few day and push them.

moorepants:
I've very close to releasing Opty 1.3. If you want this example in 1.3, please wrap it up ASAP.

moorepants:
This one still needs the German changed to English, e.g. selektion and friktion.

Peter230655:
> This one still needs the German changed to English, e.g. selektion and friktion.
I will change it.

moorepants:
I've cleaned this one up a bit more: c9da164f22ff8af0a5908274e415e67844348837

moorepants:
There are still many German words.

moorepants:
But I favor keeping "strasse" for some reason :)

Peter230655:
> There are still many German words.

I noticed this moring, that I screwed up and pushed the wrong, non-corrected file. I am sure that before I turn 75, I will be good at Github!

Peter230655:
> But I favor keeping "strasse" for some reason :)

:-)
I think, I should have called autobahn! German autobahns are famous everywhere! :-)

PR #196, opened 2024-07-24
Plot ball running on spinning disc

Peter230655:
A ball rolls on a spinning disc without slipping. opty applies a torque to the ball with the goal to get it to the center of the disc.

PR #197, opened 2024-07-27
Added a example: crane moving a load

Peter230655:
A load hanging on the horizontal jib of a crane is to be moved from A to B as fast as possible. However, the load must not 'swing over' the final position B, say there is a wall it should not crash into. opty's solution is quite different from what I naively expected.


Peter230655:
I tried to copy all 'formalities' from an existing example in examples_gallery, so I hope it will run with sphinx.

Peter230655:
This seems to look o.k. in html

moorepants:
Each plot should be associated with a new cell so that the plots are maximized in size and readable. These are not:
image

moorepants:
I  ran this and it failed:

```
message from optimizer: b"Restoration phase failed, algorithm doesn't know how to proceed."
Iterations needed 338
Objective value  2.338e-02
```

Examples should be robust and always converge. The most reliable way is to give a better initial guess (possibly even a solution).

moorepants:
> This seems to look o.k. in html

Yes, nice work!

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
+# %%
+"""
+Crane moving a load
```
moorepants, line 3:
```suggestion
Crane Moving a Load
```

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+===================
+
+A load is moved by a crane. The load is rotably connected to the crane by a
+massless rod. The goal is to move the load from the initial position to the
+final position in the shortest possible time. The load must not 'overswing'
```
moorepants, line 8:
```suggestion
final position in the shortest possible time. The load must not overswing
```

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+the final position. The load is moved by a force applied to its
+suspension point. The jib of the crane is extended in the horizontal
+X direction.
+
+**Constants**
+
+- l: length of the rod attaching the load to the crane [m]
+- m1: mass of mover attached to the arm of the crane [kg]
```
moorepants, line 16: The variables should be latex or code:

```rst
- :math:`m_1` OR ``m1``
```

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+**States**
+
+- xc: x-coordinate of the mover [m]
+- xl: x-coordinate of the load [m]  dependent coordinate
+- yl: y-coordinate of the load [m]  dependent coordinate
+- q: angle of the rod [rad]
+- uxc: velocity of the mover in x-direction [m/s]
+- uxl: velocity of the load in x-direction [m/s] dependent speed
+- uyl: velocity of the load in y-direction [m/s] dependent speed
+- u: angular velocity of the rod [rad/s]
```
moorepants, line 29: The states are technically only the minimal set of coordinates and speeds, which are likely q and u.

Peter230655:
> Each plot should be associated with a new cell so that the plots are maximized in size and readable. These are not: image

I know how cto fix it, fill come soon in the next commit.

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+
+**Specifieds**
+
+- F: force applied to the mover [N]
+
+"""
+import sympy.physics.mechanics as me
+from collections import OrderedDict
```
moorepants, line 37: OrderedDict is no longer needed, it is a remnant of Python 2 not ordering dictionaries by default.

examples-gallery/plot_crane_moving_a_load.py
Peter230655: > The states are technically only the minimal set of coordinates and speeds, which are likely q and u.

I did not know this definition, I will correct this and the other suggestions. I needed the dependent coordinates to model the non over . swing.

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+from scipy.interpolate import CubicSpline
+from opty.direct_collocation import Problem
+from opty.utils import parse_free
+import matplotlib.pyplot as plt
+import matplotlib.animation as animation
+from matplotlib import patches
+
+# %%
+# Set up Kane's equations of motion.
+# Just the standard way of doing it.
```
moorepants, line 49: Remove the second line, unnecessary.

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+P1.set_pos(O, xc * N.x)
+P1.set_vel(N, uxc * N.x)
+
+P2.set_pos(P1, -l * A.y)
+P2.v2pt_theory(P1, N, A)
+
+P1a = me.Particle('P1a', P1, m1)
+P2a = me.Particle('P2a', P2, m2)
+
+BODY = [P1a, P2a]
```
moorepants, line 71:
```suggestion
bodies = [P1a, P2a]
```

Use PEP8 naming conventions if at all possible.

Peter230655:
> I ran this and it failed:
> 
> ```
> message from optimizer: b"Restoration phase failed, algorithm doesn't know how to proceed."
> Iterations needed 338
> Objective value  2.338e-02
> ```
> 
> Examples should be robust and always converge. The most reliable way is to give a better initial guess (possibly even a solution).

Strange. I never happened with me.

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+
+FL = [(P1, F * N.x - m1*g*N.y), (P2, -m2*g*N.y)]
```
moorepants, line 73:
```suggestion
forces = [(P1, F * N.x - m1*g*N.y), (P2, -m2*g*N.y)]
```

examples-gallery/plot_crane_moving_a_load.py
moorepants: I see below you are doing a non-minimal coordinate formulation. The states are still then only the independent coordinates and speeds.

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+KM = me.KanesMethod(N,
+    q_ind=q_ind,
+    q_dependent=q_dep,
+    u_ind=u_ind,
+    u_dependent=u_dep,
+    kd_eqs=kd,
+    configuration_constraints=config_constr,
+    velocity_constraints=speed_constr)
+
+(fr, frstar) = KM.kanes_equations(BODY, FL)
```
moorepants, line 94:
```suggestion
(fr, frstar) = KM.kanes_equations(bodies, loads=forces)
```

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+EOM = kd.col_join(fr + frstar)
```
moorepants, line 95:
```suggestion
eom = kd.col_join(fr + frstar)
```

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+EOM = EOM.col_join(config_constr)
+sm.pprint(EOM)
+# %%
+# Set up the optimization problem and solve it.
+# Note: While opty could calculate the reaction forces on, say, the mover,
+# (add them to the state variables), my trials showed that even with this
+# small example the calculation time increases enourmously.
```
moorepants, line 102: The note is not helpful in this example, as it is vaguely explained and likely just distracts. I recommend removing it.

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+
+state_symbols = tuple((*q_ind, *q_dep, *u_ind, *u_dep))
+constant_symbols = (l, m1, m2, g)
+specified_symbols = (F,)
+h_opty = sm.symbols('h_opty')
```
moorepants, line 107: I'm curions why you've chosen to use `h_opty`. I changed the variable name internally to this so that there would be no clashes with using `h` as a symbol name by the user. It is my intention that a user would never know about this naming convention for the internals and showing it here reveals that!

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+unknown_symbols = []
+methode = "backward euler"
+
+num_nodes = 250
+duration =(num_nodes - 1) * h_opty
+interval_value = h_opty
+
+# Specify the known system parameters.
+par_map = OrderedDict()
```
moorepants, line 116:
```suggestion
par_map = {}
```

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+def obj(free):
+    """Minimize h_opty, the time interval between nodes."""
+    return free[-1]
+
+def obj_grad(free):
+    grad = np.zeros_like(free)
+    grad[-1] = 1.
+    return grad
+
+# start location and and final location of the load.
```
moorepants, line 133: Make a cell so text is rendered.

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+anfang = 0.0
+ende = 15.0
```
moorepants, line 135: It is preferable if the documentation is all in English to be consistent.

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+# Set up the optimization problem and solve it.
+# Note: While opty could calculate the reaction forces on, say, the mover,
+# (add them to the state variables), my trials showed that even with this
+# small example the calculation time increases enourmously.
+
+state_symbols = tuple((*q_ind, *q_dep, *u_ind, *u_dep))
+constant_symbols = (l, m1, m2, g)
+specified_symbols = (F,)
+h_opty = sm.symbols('h_opty')
+unknown_symbols = []
```
moorepants, line 108: Is this used below?

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+    num_nodes,
+    interval_value,
+    known_parameter_map=par_map,
+    instance_constraints=instance_constraints,
+    bounds=bounds,
+    integration_method=methode,
+    )
+# %%
+# Using a 'reasonable' initial guess speeds up the optimization process a lot.
+# 'Bad' initial guesses may even prevent convergence.
```
moorepants, line 178: Just state what initial guess you are constructing. You can plot it easily with `plot_trajectories()` to show the reader what it is.

examples-gallery/plot_crane_moving_a_load.py
Peter230655: > It is preferable if the documentation is all in English to be consistent.

I did this because in my early python times, I used English words, which were standard Python terms - and then I wondered why my program did not run (like I called a variable 'range'. Will change it to English.

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+i6 = [0.01]
+initial_guess = np.array(i1 + i2 + i3 + i4 + i5 + i6)
+
+# %%
+# Plot the initial guess.
+prob.plot_trajectories(initial_guess)
+
+
+# this allows to change the maximum number of iterations.
+# Standard is 3000.
```
moorepants, line 193: Still a code comment. Render as cell text.

moorepants: Or remove.

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+prob.add_option('max_iter', 1000)
+# Find the optimal solution.
+for _ in range(1):
```
moorepants, line 196: Remove this loop if there is only one iteration.

Ideally all examples solve with a single call to `.solve()`.

examples-gallery/plot_crane_moving_a_load.py
Peter230655: I had wondered about the OrderedDict - but just copied it from your examples. Will change it.

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+    solution, info = prob.solve(initial_guess)
+    print('message from optimizer:', info['status_msg'])
+    prob.plot_objective_value()
+    initial_guess = solution
+print('Iterations needed', len(prob.obj_value))
+print(f"Objective value {solution[-1]: .3e}")
+
+# %%
+# plot the accuracy of the solution.
```
moorepants, line 205:
```suggestion
# Plot the accuracy of the solution.
```

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+prob.plot_constraint_violations(solution)
+
+# %%
+# Plot the state trajectories.
+fig, ax = plt.subplots(9, 1, figsize=(7.25, 3.0*9), sharex=True)
+prob.plot_trajectories(solution, ax)
+
+# %%
+# Animate the simulation.
+fps = 30
```
moorepants, line 215: This is a function of the optimal `h` value. If you choose 30 frames per second, the animation time will not align with the actual solution.

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+input_sol = CubicSpline(t_arr, input_vals.T)
+
+coordinates = P1.pos_from(O).to_matrix(N)
+coordinates = coordinates.row_join(P2.pos_from(O).to_matrix(N))
+
+pL, pL_vals = zip(*par_map.items())
+coords_lam = sm.lambdify((*state_symbols, F, *pL),
+    coordinates, cse=True)
+
+# Dimensions of the mover and the load.
```
moorepants, line 231: Code comment, will no render as text.

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+        width=width, height=height, fill=True, color='red', ec='black')
+    ax.add_patch(recht)
+    load = patches.Circle((0, -par_map[l]), radius=radius, fill=True,
+        color='blue',ec='black')
+    ax.add_patch(load)
+    return fig, ax, line1, recht, load, pfeil
+
+fig, ax, line1, recht, load, pfeil = init_plot()
+
+# Function to update the plot for each animation frame
```
moorepants, line 268: Code comment.

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,294 @@
...
+
+    line1.set_data([coords[0, 0], coords[0, 1]], [coords[1, 0], coords[1, 1]])
+    recht.set_xy((coords[0, 0] - width/2., coords[1, 0] - height/2.))
+    load.set_center((coords[0, 1], coords[1, 1]))
+    pfeil.set_offsets([coords[0, 0], coords[1, 0]+0.25])
+    pfeil.set_UVC(input_sol(t) , 0.25)
+    return line1, recht, load, pfeil
+
+
+# Create the animation
```
moorepants, line 283: Code comment.

examples-gallery/plot_crane_moving_a_load.py
Peter230655: > The variables should be latex or code:
> 
> ```
> - :math:`m_1` OR ``m1``
> ```

I had played around with :math:`.....` but did not get it to work right.
Will it only work in lines like:

"""
- :math:`....`

"""
Will it not work in lines like:
# -:math:`...`
?

examples-gallery/plot_crane_moving_a_load.py
moorepants: That's a good method, but we should just make the docs consistent in language. Sphinx does have mechanisms to make multiple versions of the docs in different languages. I've never done that, but I'm sure it would be helpful to many learners.

examples-gallery/plot_crane_moving_a_load.py
moorepants: Math in RestructuredText is explained here:

https://ikerdocs-sphinx.readthedocs.io/syntax/math.html

examples-gallery/plot_crane_moving_a_load.py
Peter230655: > Is this used below?

No. I also copied from some example, maybe from Timo. Allows the prob = Problem(..) to be always the same.
Better delete it?

moorepants: If it isn't used, then delete it.

examples-gallery/plot_crane_moving_a_load.py
Peter230655: > I'm curions why you've chosen to use `h_opty`. I changed the variable name internally to this so that there would be no clashes with using `h` as a symbol name by the user. It is my intention that a user would never know about this naming convention for the internals and showing it here reveals that!

Now I understand your intention with h_opty. Will change back to h.

examples-gallery/plot_crane_moving_a_load.py
Peter230655: > Code comment, will no render as text.

Should code comments be avoided in general?

examples-gallery/plot_crane_moving_a_load.py
Peter230655: > Math in RestructuredText is explained here:
> 
> https://ikerdocs-sphinx.readthedocs.io/syntax/math.html

This is what I looked at, but never got it to work. I will play around some more.

examples-gallery/plot_crane_moving_a_load.py
Peter230655: > This is a function of the optimal `h` value. If you choose 30 frames per second, the animation time will not align with the actual solution.

I  just copied this idea from Timo. I run the animation for the duration determined by h, that is duration = (num_nodes - 1) * h.
So, I am not sure I understand what you mean.

moorepants: If h = 0.135, for example, then there is 1 frame every 0.135 seconds. If you set the fps to 30 in the animation, there is one frame per 0.05 seconds.

examples-gallery/plot_crane_moving_a_load.py
moorepants: For these examples, yes, but it is better to have lots of rendered explanatory text so people can easily learn from the examples.

examples-gallery/plot_crane_moving_a_load.py
Peter230655: > If h = 0.135, for example, then there is 1 frame every 0.135 seconds. If you set the fps to 30 in the animation, there is one frame per 0.05 seconds.

Do you mean, the animation will run in slow motion or in fast motion compared to 'reality'?

NB: I will make the corrections you suggested this evening an push a new version. I will also make these corrections on my other simulations, so don't waste your time looking at them before I pshed new versions.

examples-gallery/plot_crane_moving_a_load.py
Peter230655: > The states are technically only the minimal set of coordinates and speeds, which are likely q and u.

they are xc, uxc, q, u. xc

Peter230655:
Looks like I completely screwed it up! Lost all my changes! Will do them again.

moorepants:
I've very close to releasing Opty 1.3. If you want this example in 1.3, please wrap it up ASAP.

Peter230655:
> I've very close to releasing Opty 1.3. If you want this example in 1.3, please wrap it up ASAP.

I thought I did correct everything?

moorepants:
>  Are there still mistakes?

I don't know. I'll have to review it again.

examples-gallery/plot_crane_moving_a_load.py
moorepants: Sphinx-gallery treats the module level "docstring" as a "cell" of RestructuredText.

Peter230655:
> > Are there still mistakes?
> 
> I don't know. I'll have to review it again.

I pushed #195, #196, #197 a few days ago. I can look at them in HTML now, so I corrected what I saw wrong / what you told me, like oem instead of OEM, bodies instead of BODY, etc.

examples-gallery/plot_crane_moving_a_load.py
moorepants: Setting fps to match the spacing in a time vector will cause the animation to at a realistic rate.

Peter230655: > Setting fps to match the spacing in a time vector will cause the animation to at a realistic rate.

So, fps approx. = 1 / h, where h = duration / (num_nodes - 1) ?

Peter230655:
> > I've very close to releasing Opty 1.3. If you want this example in 1.3, please wrap it up ASAP.
> 
> I thought I did correct everything, and pushed it. Are there still mistakes? I corrected two more and pushed them, a few days ago.

You will release opty 1.3 soon. Did you get a chance to look at Discussion #191? Is this just my misunderstanding of the docs, or is there a bug in info['g'], or is it no very relevant?

examples-gallery/plot_crane_moving_a_load.py
moorepants: This has not bee addressed.

Peter230655:
I changed, say, q to ``q``, but it seems I cannot use `` in commit comments.

moorepants:
I made a number of formatting fixes and tried to correct the fps. You can see what I changed here:

https://github.com/csu-hmc/opty/pull/197/commits/082d9ebb6a02d9624bef5a8b54ac0da13e06a779

You can do `git pull origin plot_crane_moving_a_load` to get my updates in your branch.

Two issues remain:

1. One out of three times this fails to find an optimal solution for me. If the initial guess can be improved, maybe that will not happen.
2. I don't think the animation is animating to the final time, as it doesn't look like it ever hits the wall. This may be due to the way you are generating the animation frames and that `arange()` will skip the actual last moment in time when it hits the wall.

Peter230655:
> I made a number of formatting fixes and tried to correct the fps. You can see what I changed here:
> 
> [082d9eb](https://github.com/csu-hmc/opty/commit/082d9ebb6a02d9624bef5a8b54ac0da13e06a779)
> 
> You can do `git pull origin plot_crane_moving_a_load` to get my updates in your branch.
> 
> Two issues remain:
> 
>     1. One out of three times this fails to find an optimal solution for me. If the initial guess can be improved, maybe that will not happen.
> 
>     2. I don't think the animation is animating to the final time, as it doesn't look like it ever hits the wall. This may be due to the way you are generating the animation frames and that `arange()` will skip the actual last moment in time when it hits the wall.

1. I improved the initial guess by using the solution of a previous run. It never failed to find a solution with me, at least always an acceptable one. What I pushed found an optimal solution with me.
2. I thought arange would get the last point, I will check.
3. I just got this error:
image
This is the file where I store the solution. I had no problem with it doing make html ?

Peter230655:
In this example, I stored the result on a run, and use this result like this initial_guess = np.load('solution of a previous run').
This makes it a bit faster.
Is this acceptable for examples-gallery? Will examples-gallery 'have' this previous solution?

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,338 @@
...
+# guess.
+
+# - i1 = [(ending_location - starting_location)/num_nodes*i for i in
+#        range(num_nodes)]
+# - i2 = [0.0 for _ in range(num_nodes)]
+# - i3 = i1
+# - i4 = [-par_map[l] for _ in range(num_nodes)]
+# - i5 = [0.0 for _ in range(5*num_nodes)]
+# - i6 = [0.01]
+# - initial_guess = np.array(i1 + i2 + i3 + i4 + i5 + i6)
```
moorepants, line 222: Move this to a full comment cell as an explanation, something like:

```
# %%
# This initial guess was used to get a solution which forms the actual initial
# guess::
#
#    i1 = ...
#    i2 = ...

File: examples-gallery/plot_crane_moving_a_load.py [hunk 26]
```diff
@@ -0,0 +1,338 @@
...
+    line1.set_data([coords[0, 0], coords[0, 1]], [coords[1, 0], coords[1, 1]])
+    recht.set_xy((coords[0, 0] - width/2., coords[1, 0] - height/2.))
+    load.set_center((coords[0, 1], coords[1, 1]))
+    pfeil.set_offsets([coords[0, 0], coords[1, 0]+0.25])
+    pfeil.set_UVC(input_sol(t), 0.25)
+    return line1, recht, load, pfeil
+
+anim = animation.FuncAnimation(fig, update,
+                               frames=np.arange(t0, tf, 1/fps),
+                               interval=1/fps*1000)
```
moorepants, line 330: The animation does not run until it hits the for me. Maybe it is running out of memory. I suggest reduce the fps and number of frames.

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -210,10 +210,9 @@ def obj_grad(free):
 
 # %%
 # This initial guess was used to get a solution which forms the actual initial
-# guess.
-
-# - i1 = [(ending_location - starting_location)/num_nodes*i for i in
-#        range(num_nodes)]
+# guess::
+#
+# - i1 = [(ending_location - starting_location)/num_nodes*i for i in range(num_nodes)]
```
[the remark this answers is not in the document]
moorepants: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks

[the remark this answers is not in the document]
Peter230655: > The spacing needs to be such that it renders as code.

Should this run, and then be overwritten by initial_guess = np.load('solution.npy') ?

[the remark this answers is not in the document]
moorepants: Your idea to execute it but then just overwrite the initial guess with the solution is better.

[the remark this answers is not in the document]
Peter230655: > Your idea to execute it but then just overwrite the initial guess with the solution is better.

I will change this and make a new PR.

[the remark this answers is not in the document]
moorepants: Please don't make new prs for the same topic, just update this one.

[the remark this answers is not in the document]
Peter230655: > Please don't make new prs for the same topic, just update this one.

Sorry, I meant I would add to this one only!

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,337 @@
...
+i5 = [0.0 for _ in range(5*num_nodes)]
+i6 = [0.01]
+initial_guess = np.array(i1 + i2 + i3 + i4 + i5 + i6)
+
+# %%
+# The initial guess is the solution of some previous run, saved in the file
+# 'solution.npy', to speed up convergence. To get this solution, the initial
+# guess above was used, which is now overwritten.
+
+initial_guess = np.load('solution.npy')
```
moorepants, line 228: Can you rename the file to `crane_moving_a_load_solution.npy` because we will likely add more of these files. I think even another PR has one named the same and they are all in the same directory.

Peter230655: > Can you rename the file to `crane_moving_a_load_solution.npy` because we will likely add more of these files. I think even another PR has one named the same and they are all in the same directory.

Makes emminent sense! Will do so right now

moorepants:
You need to `git rm solution.npy`.

moorepants:
Yes

moorepants:
Yes

examples-gallery/plot_crane_moving_a_load.py
```diff
@@ -0,0 +1,337 @@
...
+i2 = [0.0 for _ in range(num_nodes)]
+i3 = i1
+i4 = [-par_map[l] for _ in range(num_nodes)]
+i5 = [0.0 for _ in range(5*num_nodes)]
+i6 = [0.01]
+initial_guess = np.array(i1 + i2 + i3 + i4 + i5 + i6)
+
+# %%
+# The initial guess is the solution of some previous run, saved in the file
+# 'solution.npy', to speed up convergence. To get this solution, the initial
```
moorepants, line 225: Needs updating to new filename.

Peter230655: > Needs updating to new filename.

How come you see EVERYTHING?!? I guess, because you are a teacher by trade. Will change it right now.

moorepants:
Thanks! Nice example.

examples-gallery/plot_crane_moving_a_load.py
moorepants: Yes, maybe grading papers causes that.

Peter230655:
> Thanks! Nice example.

Thank YOU! I must admit, it makes me feel good! :-)
What I like about this example: Intuitively I had expected that it would crawl towards the right wall. Actually, it goes full speed!
opty often (in my toy eamples) finds an unexpected (by me) solution.
NB: I am quite pleased that I found a solution to convert the variable h into a correctly rounded number for plot_constraint_violations. :-)

PR #205, opened 2024-08-05
Variable time support in create_objective_function.

moorepants:


moorepants:
Only if I can figure it out.

Peter230655:
> Only if I can figure it out.

I feel like you know it already! Surely will avoid these ugly expressions like in my ball/disc simulation.

moorepants:
A general solution to this needs to support this cost function:
image

(general optimal control cost function)

Creating the discrete version of that and the discrete gradient from a general SymPy expression is not so simple.

There are past open issues about this: #30 #31

Peter230655:
Just to make sure I understand it correctly:
- E and F are functions, differentiable in all their arguments.
- x is the state vector of the system.
- u is the control (input) vector.
- J[x(.), u(.), t_0, t_f] is to be minimized.

A necessary condition for a minimum is that grad_(x, u, t_f)(J) = 0.

moorepants:
Yes, I was starting to implement your suggestion but I think it was missing taking the derivative wrt to `h` and I started getting hung up on the solution. I basically copied your code and started trying to make it work, but haven't gotten it to yet.

tjstienstra:
> but I think it was missing taking the derivative wrt to h

I would expect that you can just add it to the objective_grad computation, the list of symbols w.r.t. which the jacobian is computed. However, I don't remember testing the solution (in depth), but it is nice to see that you have already written some tests.

opty/tests/test_utils.py
```diff
@@ -200,6 +200,43 @@ def test_invalid_integration_limits(self):
...
+        def expected_obj(free):
+            f = free[2*self.N:-1]
+            return free[-1]*np.sum(f**2)
+
+        def expected_obj_grad(free):
+            f = free[2*self.N:-1]
+            grad = np.zeros_like(free)
+            grad[2*self.N:-1] = 2.0*free[-1]*free[2*self.N:-1]
+            grad[-1] = np.sum(f**2)
+            return grad
```
tjstienstra, line 214: You should take into account that this is backward Euler, so the first term falls out, see `test_backward_single_input`.
If free is [x(t), v(t), f1(t), f2(t), c, k, m, h] then the objective should be `(f1_vals[1:]**2 + f2_vals[1:]**2).sum() * h_val`.
Similarly, the gradient should be a stack of zeros(2*N+1), 2*h_val*f1_vals[1:], [0] 2*h_val*f2_vals[1:], [0, 0, 0, (f1_vals[1:]**2 + f2_vals[1:]**2).sum()]

P.S. quickly wrote out the equations on my phone so would advise checking them.

moorepants: I don't think I assumed any specific integration routine in the manually created objective functions.

Peter230655:
> A general solution to this needs to support this cost function: image
> 
> (general optimal control cost function)
> 
> Creating the discrete version of that and the discrete gradient from a general SymPy expression is not so simple.
> 
> There are past open issues about this: #30 #31

To check whether an expression E explicitly depends on the time vatiable t, one could check whether $\dfrac{\partial E}{\ partial t} = 0$. I checled the sympy docs for partial derivatives but could not find anything to do this. Did I miss something?
Thanks!

opty/tests/test_utils.py
Peter230655: > I don't think I assumed any specific integration routine in the manually created objective functions.

I have a basic question:
In the current create_objective_function(...) there is a distinction in forming obj, obj_grad depending on the integration method.
But, for example, in the examples_gallery example plot_pendulum_swing_up_variable_duration, the gradient is formed as I would naively expect it to be formed.

Can the gradient always be formed as per the method used in plot_pendulum_swing_up_variable_duration, or is this only valid for midpoint euler or...?
**Thanks for any explanations!**

Peter230655: Thanks!
This would mean one could calculate obj and obj_grad 'naively' without committing a large error?

moorepants: The gradient of the objective has to be the valid gradient within some numerical tolerance. But your choice of solving the integral in the objective does not really matter, as there are numerous integration methods.

Peter230655: Thanks!
The gradient as calculated in the simulation _plot_pendulum_swing_up_variable_duration_ surely is the valid gradient (?)

PR #208, opened 2024-08-20
Mississippi steamboat on a lake

Peter230655:
A boat is on a lake. It has a water wheel on each side to drive and to stir it, hence the name Mississippi. The control variables are the torques applied to each wheel. The goal is to get it from A to B. the objective function is a linear combination of minimizing the energy and minimizing the time needed. the drag force on a body with speed $\bar v$ is modeled as
$F_D = -c \cdot A \cdot | \bar v |^2 \cdot \dfrac{\bar v}{| \bar v |}$,  where c is the drag coefficient and A is the surface facing the flow.
I do not print the EOMs as they have around 850 operations. Else, I hope all close to examples-gallery standards.


moorepants:
Error:

```
/home/moorepants/src/opty/docs/examples/plot_Mississippi_steamboat.rst:307: WARNING: Title underline too short.

Set up the Optimization Problem and Solve it.
--------------------------------------------
/home/moorepants/src/opty/docs/examples/plot_Mississippi_steamboat.rst:307: WARNING: Title underline too short.

Set up the Optimization Problem and Solve it.
--------------------------------------------
```

examples-gallery/plot_Mississippi_steamboat.py
```diff
@@ -0,0 +1,445 @@
...
+# -------------------------------
+#
+# Set up the geometry of the system.
+#
+# - :math:`N`: inertial frame of reference
+# - :math:`O`: origin of the inertial frame of reference
+# - :math:`A_S`: body fixed frame of the steamboat
+# - :math:`A_{LW}`: body fixed frame of the left wheel
+# - :math:`A_{RW}`: body fixed frame of the right wheel
+# - :math:`Dmc_S`: center of the steamboat
```
moorepants, line 68: Is the "mc" supposed to be a subscript?

examples-gallery/plot_Mississippi_steamboat.py
```diff
@@ -0,0 +1,445 @@
...
+# - :math:`Dmc_{LW}`: center of the left wheel
+# - :math:`Dmc_{RW}`: center of the right wheel
+# - :math:`FP_{LW}`: point where the force due to rotation of the left
+#   wheel is applied
+# - :math:`FP_{RW}`: point where the force due to rotation of the right
+#   wheel is applied
+
+N, AS, ALW, ARW = sm.symbols('N, AS, ALW, ARW', cls = me.ReferenceFrame)
```
moorepants, line 76:
```suggestion
N, AS, ALW, ARW = sm.symbols('N, AS, ALW, ARW', cls=me.ReferenceFrame)
```

PEP8: no space around kwarg equals signs.

examples-gallery/plot_Mississippi_steamboat.py
```diff
@@ -0,0 +1,445 @@
...
+FPLW.set_vel(N, DmcLW.vel(N) + uLW*AS.x.cross(-rW*N.z))
+FPRW.set_pos(DmcRW, -rW*N.z)
+FPRW.set_vel(N, DmcRW.vel(N) + uRW*AS.x.cross(-rW*N.z))
+
+
+# %%
+# Set up the drag forces acting on the boat.
+#
+# The drag force acting on a body moving in a fluid is given by
+# :math:`F_D = -c \cdot A \cdot | \bar v|^2 \cdot \dfrac{\bar v}{| \bar v|}`,
```
moorepants, line 113: Drag force is more typically:

$$
F = -\textrm{sgn}(v)\frac{1}{2} \rho C_D A v^2
$$

where $\rho$ is the density of the fluid.

examples-gallery/plot_Mississippi_steamboat.py
```diff
@@ -0,0 +1,445 @@
...
+# where :math:`c` is the drag coefficient, :math:`\bar v` is the velocity
+# vector of the body and :math:`A` is the cross section area of the body facing
+# the flow. This may be found here:
+#
+# https://courses.lumenlearning.com/suny-physics/chapter/5-2-drag-forces/
```
moorepants, line 118: You can remove this link, as it is a textbook equation. Or better link to a wikipedia article about it.

examples-gallery/plot_Mississippi_steamboat.py
```diff
@@ -0,0 +1,445 @@
...
+#
+# In order to avoid numerical issues with .normalize(), sm.sqrt(..)
+# I will use the following:
+#
+# ( :math:`\circ` denotes the inner product)
+#
+# :math:`F_{D_x} = -c \cdot A \cdot (A.x \circ \bar v)^2 \cdot \operatorname{sgn}(A.x \circ \bar v) \cdot A.x`
+# :math:`F_{D_y} = -c \cdot A \cdot (A.y \circ \bar v)^2 \cdot \operatorname{sgn}(A.y \circ \bar v) \cdot A.y`
```
moorepants, line 126: I typically use `\cdot` for dot product and do not use a symbol for scalar multiplication, e.g. %a\hat{b})x\cdot\hat{a}_x$.

Also, sympy prints unit vectors like: `\hat{\mathbf{A}}_x`.

examples-gallery/plot_Mississippi_steamboat.py
```diff
@@ -0,0 +1,445 @@
...
+#
+# As an (infinitely often) differentiable approximation of the sign function,
+# I will use the fairly standard approximation:
+#
+# :math:`\operatorname{sgn}(x) \approx \tanh( \alpha \cdot x )` with :math:`\alpha \gg 1`
```
moorepants, line 131: If you specify all the variables to have correct assumptions (for example `real=True` or `nonegative=True`, then you can use `Abs()` and it will differentiate once and print with the C printers. I don't think you have to use this tangent function.

examples-gallery/plot_Mississippi_steamboat.py
```diff
@@ -0,0 +1,445 @@
...
+# %%
+# Set up the objective function and its gradient. The objective function
+# to be minimized is:
+#
+# :math:`\text{obj} = \int_{t_0}^{t_f} \left( t_{LW}^2 + t_{RW}^2 \right) \, dt + \text{weight} \cdot h`
+#
+# where weight > 0 is the relative importance of the duration of the motion,
+# and h > 0.
+
+weight = 1.e7
```
moorepants, line 230:
```suggestion
weight = 1.0e7
```

examples-gallery/plot_Mississippi_steamboat.py
```diff
@@ -0,0 +1,445 @@
...
+    }
+
+final_state_constraints = {
+    q: -np.pi/2.,
+    x: 10,
+    y: 10,
+    u: 0.0,
+    ux: 0.0,
+    uy: 0.0,
+    }
```
moorepants, line 271:
```suggestion
}
```

examples-gallery/plot_Mississippi_steamboat.py
```diff
@@ -0,0 +1,445 @@
...
+    uy.subs({t: t0}) - initial_state_constraints[uy],
+    uLW.subs({t: t0}) - initial_state_constraints[uLW],
+    uRW.subs({t: t0}) - initial_state_constraints[uRW],
+    q.subs({t: tf}) - final_state_constraints[q],
+    x.subs({t: tf}) - final_state_constraints[x],
+    y.subs({t: tf}) - final_state_constraints[y],
+    u.subs({t: tf}) - final_state_constraints[u],
+    ux.subs({t: tf}) - final_state_constraints[ux],
+    uy.subs({t: tf}) - final_state_constraints[uy],
+    )
```
moorepants, line 290:
```suggestion
)
```

examples-gallery/plot_Mississippi_steamboat.py
```diff
@@ -0,0 +1,445 @@
...
+
+limit_torque = 25.
+bounds = {
+    tLW: (-limit_torque, limit_torque),
+    tRW: (-limit_torque, limit_torque),
+    h: (0.0, 1.0),
+    }
```
moorepants, line 297:
```suggestion
}
```

examples-gallery/plot_Mississippi_steamboat.py
```diff
@@ -0,0 +1,445 @@
...
+    obj_grad,
+    eom,
+    state_symbols,
+    num_nodes,
+    interval_value,
+    time_symbol=t,
+    known_parameter_map=par_map,
+    instance_constraints=instance_constraints,
+    bounds=bounds,
+    )
```
moorepants, line 310:
```suggestion
)
```

examples-gallery/plot_Mississippi_steamboat.py
```diff
@@ -0,0 +1,445 @@
...
+fig, ax = plt.subplots(12, 1, figsize=(8, 20), constrained_layout=True,
+    sharex=True)
+prob.plot_trajectories(solution, ax)
+
+# %%
+# Animate the solutions and plot the results.
+fps = 30
+
+def add_point_to_data(line, x, y):
+# to trace the path of the point.
```
moorepants, line 343:
```suggestion
    # to trace the path of the point.
```

examples-gallery/plot_Mississippi_steamboat.py
```diff
@@ -0,0 +1,445 @@
...
+    ax.scatter(initial_state_constraints[x], initial_state_constraints[y],
+        color='red', s=10)
+    ax.scatter(final_state_constraints[x], final_state_constraints[y],
+        color='green', s=10)
+    ax.axhline(initial_state_constraints[y], color='black', lw=0.25)
+    ax.axhline(final_state_constraints[y], color='black', lw=0.25)
+    ax.axvline(initial_state_constraints[x], color='black', lw=0.25)
+    ax.axvline(final_state_constraints[x], color='black', lw=0.25)
+
+# draw the wheels and the line connecting them
```
moorepants, line 394:
```suggestion
    # draw the wheels and the line connecting them
```

examples-gallery/plot_Mississippi_steamboat.py
```diff
@@ -0,0 +1,445 @@
...
+    line1, = ax.plot([], [], lw=2, marker='o', markersize=0, color='green',
+    alpha=0.5)
+    line2, = ax.plot([], [], lw=2, marker='o', markersize=0, color='black')
+    line3, = ax.plot([], [], lw=2, marker='o', markersize=0, color='black')
+
+# draw the torque vektor
```
moorepants, line 400:
```suggestion
    # draw the torque vektor
```

examples-gallery/plot_Mississippi_steamboat.py
```diff
@@ -0,0 +1,445 @@
...
+    pfeil1 = ax.quiver([], [], [], [], color='red', scale=100, width=0.004)
+    pfeil2 = ax.quiver([], [], [], [], color='blue', scale=100, width=0.004)
+# draw the boat
```
moorepants, line 403:
```suggestion
    # draw the boat
```

examples-gallery/plot_Mississippi_steamboat.py
Peter230655: > Is the "mc" supposed to be a subscript?

No, Sir. I always use, if possible Dmc to be the ceter of mass of some body. I must have copied this from some example a long time ago. If I have Dmc in a name in my code I know it is a center of mass of something.

examples-gallery/plot_Mississippi_steamboat.py
Peter230655: In my article, which I will remove and replace, there was the 'suggestion' to lump $\rho, C_D, \frac{1}{2}$ into one variable.
Hence I did it.

examples-gallery/plot_Mississippi_steamboat.py
Peter230655: > You can remove this link, as it is a textbook equation. Or better link to a wikipedia article about it.

this one o.k.?
https://en.wikipedia.org/wiki/Drag_(physics)

examples-gallery/plot_Mississippi_steamboat.py
Peter230655: > If you specify all the variables to have correct assumptions (for example `real=True` or `nonegative=True`, then you can use `Abs()` and it will differentiate once and print with the C printers. I don't think you have to use this tangent function.

Mathematically abs(x-a) is not differentiable at x=a, or so I understand it. How does sympy go around this?
(I had a problem with opty and abs before, but I do not think, I declared the variables to be real)

examples-gallery/plot_Mississippi_steamboat.py
moorepants: I'm asking more because the math renders awkwardly, not because I don't like your naming conventions. You have these symbols as proper math, so I would expect them to have proper super and subscripts.

examples-gallery/plot_Mississippi_steamboat.py
moorepants: There is a whole page about the "drag equation": https://en.wikipedia.org/wiki/Drag_equation

examples-gallery/plot_Mississippi_steamboat.py
moorepants:
```python
In [1]: import sympy as sm

In [2]: t = sm.symbols('t', real=True, nonnegative=True)

In [3]: v = sm.Function('v', real=True)(t)

In [4]: f = sm.Abs(v)

In [5]: f
Out[5]: Abs(v(t))

In [6]: f.diff(t)
Out[6]: sign(v(t))*Derivative(v(t), t)

In [7]: f.diff(v)
Out[7]: sign(v(t))*Derivative(v(t), v(t))
```

Peter230655:
I (hopefully) made all the corrections you suggested, except I did not dare replacing the tanh(x) expression with the equivalent abs(x) expression. opty does not 'like' functions which are not differentiable, you told me at least once before.

examples-gallery/plot_Mississippi_steamboat.py
Peter230655: Thanks!
This means, sympy defines $\dfrac{d}{dt} abs(x(t)) = 0$ at x(t) = 0, it seems.
Would opty not 'prefer' functions which have continuous derivatives, or does this not affect convergence?

examples-gallery/plot_Mississippi_steamboat.py
Peter230655: > I'm asking more because the math renders awkwardly, not because I don't like your naming conventions. You have these symbols as proper math, so I would expect them to have proper super and subscripts.

If I understand you correctly, something like $Dmc_S$ looks awkward?

moorepants: I would expect something more like:

$$
D_{{mc}_s}
$$

or

$$
D^{mc}_s
$$

moorepants: We typically use `A_o` or `A^*` for the mass center of rigid body $A$.

examples-gallery/plot_Mississippi_steamboat.py
moorepants: Probably just leave it as the tanh function.

>  I recall you told me it is better to use t = me.dynamicsymbols._t

That is required if you use the default symbol for t in sympy.physics.mechanics. But you can set t to any variable you want.

examples-gallery/plot_Mississippi_steamboat.py
Peter230655: > We typically use A o or A ∗ for the mass center of rigid body A .

I thought capital A was 'reserved' for body fixed frames?
So, I could change to $D_{S}^{mc}$

moorepants:
Here is a trick for using Abs() (but not required for this example). https://github.com/csu-hmc/gait2d/blob/master/pygait2d/segment.py#L444

Peter230655:
I used the link to drag force you gave.
I used $D^{mc}_X$ to designate the center of gravity of body X

examples-gallery/plot_Mississippi_steamboat.py
moorepants: You can use any letter you want for a frame, I meant that if you have a frame $X$, then the mass center is $X_o$ of $X^*$.

Peter230655: > You can use any letter you want for a frame, I meant that if you have a frame X , then the mass center is X o or X ∗ .

So, if the body fixed frame of my steamboat is AS, then $AS_o$ could be its mass center?

Peter230655:
I renamed the mass centers as follows:
if $A_S$ is the body fixed frame of S, then its center of mass is $A^o_S$

Peter230655:
> Here is a trick for using Abs() (but not required for this example). https://github.com/csu-hmc/gait2d/blob/master/pygait2d/segment.py#L444

Now I remember, clever! I believe you showed this no me in another context, it was this trick:
penetration = (abs(y_position) - y_position) / 2.
you showed to me then.

Would it be a sound rule of thumb for opty to avoid function which are not differentiable at certain points unless one is sure, these points will never be close to the optimum solution of opty?

Peter230655:
Hold off with this one, please.
I may have oversimplified the moment caused by a rotation of the steamboat.

Peter230655:
These two commits are the same, I just had some problem (I am still not very perfect with this github stuff), so I committed again.
All I changed was a more realistic formula for the torque ehrn the boat rotates.

moorepants:
Thanks, merged!

Peter230655:
> Thanks, merged!

Thank you! :-)
So, even proper names, like Mississippi should be lower case in the name.

moorepants:
I just follow PEP8 on Python module filenames:

> Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.

moorepants:
https://google.github.io/styleguide/pyguide.html#3164-guidelines-derived-from-guidos-recommendations

Peter230655:
> https://google.github.io/styleguide/pyguide.html#3164-guidelines-derived-from-guidos-recommendations

Clear!
So modules like this: this_is_a_module
Packages like this: this-is-a-package

moorepants:
No, they all use underscores. If you name a file `a-b-c.py` then you can't import it `import a-b-c` because the `-` is interpreted as subtraction in python.

Peter230655:
> No, they all use underscores. If you name a file `a-b-c.py` then you can't import it `import a-b-c` because the `-` is interpreted as subtraction in python.

so, either thisisapackage or this_is_a_package

moorepants:
yes

PR #209, opened 2024-08-21
Plot ball rolling on spinning disc

Peter230655:
This is the ball rolling on a spinning disc. Setting up the eoms once only now works fine.


Peter230655:
I try again

PR #214, opened 2024-08-29
sailboat 1st commit

Peter230655:
A sailboat in still water with constant wind is supposed to reach some point and return to its origin. Basicall I have been wondering how a sailboat can sail against the wind, this optimization shows it is possible with my assumptions on drag forces.


moorepants:
Neat problem to test!

Peter230655:
Github gets this error:
image
This is the file where I save the solution as initial gues so the next run is much faster.

Peter230655:
> Neat problem to test!

Thanks! :-)

moorepants:
You will have to commit the npy file in this PR.

Peter230655:
> You will have to commit the npy file in this PR.

Thanks!
I will fix my other PRs, too.
I am in **NO** rush for you to look at them, you have better work to do!

Peter230655:
This iexample does not add anything to examples-gallery, so I close it.
As I think, it is quite nice on its own, I moved it to pst-notebooks

PR #216, opened 2024-08-29
Subexpression replacements in the Jacobian should be real.

moorepants:


moorepants:
@ricdigi this is an interesting case that just came up when taking the derivative with the cse method. If you apply assumptions to the variables that make up the expressions and then create symbols for the subexpressions that do not somehow carry the assumptions of the subexpression it matches, then you may get a different derivative than you would by applying the regular Jacobian method. A simple test case is added `Abs()` of a variable that has assumptions making the argument real or real and positive. cc/ @tjstienstra

PR #218, opened 2024-08-30
Plot constraint violations

Peter230655:
I tried to make the prints of the method plot_constraint violations a bit easier to read. In particular
- maximally 10 bars per plot. If there are more constraint violations, more plots will be made
- changed printing of the constraints, now in exponential writing, like e.g. $1.24 \cdot 10^{-3}$ instead of 0.000123456


Peter230655:
I have a very basic question about testing such a thing:
When I started, I made a new branch, and in this branch, I opened direct_allocation.py
To test what I did, I simply copied a simulation below direct_allocation. and ran the wohle thing every time.

But this seems inefficient if I want to test it with different simulations.
In a normal simulation I wrtie: 
- from opty.direct_allocation import Problem.
But what do I write, if I want to use my modified direct_allocation ?

**Thanks for any help!**

Peter230655:
The error Github found may be due to this:
If the user give the axes, he must give 'enough' axes, just two may not be the right amount.
It should be
- 2 if len(instant_constraints) $leg$ 10
- len(instant_constraints) // 10 + 1 otherwise

moorepants:
My approach to testing edits I make to a package, like opty, is to first create a conda environment with all the development dependencies. We include this file in the repo for such a thing: https://github.com/csu-hmc/opty/blob/master/opty-dev-env.yml. After that, I install the development version of opty in a way that any import of opty will always reflect the current state of the opty repo on my computer.

```bash
$ cd /path/to/opty/repo
$ conda env create -f opty-dev-env.yml
$ conda activate opty-dev
$ python -m pip install -e .  # installs version in the current directory, i.e. the development version
```

Now as long as I have that conda environment activated it will use the opty version in `/path/to/opty/repo`.

So then, if I want to build the docs, just:

```bash
$ cd docs
$ make html
```

All of the constraint violations plots should generate using your edits to the direct_collocation.py file.

moorepants:
Looks like your code doesn't work on the cycling example:

```
Extension error:
Here is a summary of the problems encountered when running the examples:

Unexpected failing examples (2):

    ../examples-gallery/plot_one_legged_time_trial.py failed leaving traceback:

    Traceback (most recent call last):
      File "/home/runner/work/opty/opty/examples-gallery/plot_one_legged_time_trial.py", line 748, in <module>
        problem.plot_constraint_violations(solution, axes=axes)
      File "/home/runner/work/opty/opty/opty/utils.py", line 218, in wrapper
        func(*args, **kwargs)
      File "/home/runner/work/opty/opty/opty/direct_collocation.py", line 524, in plot_constraint_violations
        axes[i+1].bar(
        ~~~~^^^^^
    IndexError: index 2 is out of bounds for axis 0 with size 2

    ../examples-gallery/plot_drone.py failed leaving traceback:

    Traceback (most recent call last):
      File "/home/runner/work/opty/opty/examples-gallery/plot_drone.py", line 235, in <module>
        prob.plot_constraint_violations(solution, axes=axes)
      File "/home/runner/work/opty/opty/opty/utils.py", line 218, in wrapper
        func(*args, **kwargs)
      File "/home/runner/work/opty/opty/opty/direct_collocation.py", line 524, in plot_constraint_violations
        axes[i+1].bar(
        ~~~~^^^^^
    IndexError: index 2 is out of bounds for axis 0 with size 2

-------------------------------------------------------------------------------
make: *** [Makefile:25: html] Error 2
Error: Process completed with exit code 2.
```

moorepants:
>  If the user give the axes, he must give 'enough' axes, just two may not be the right amount.

As long as you document this, I think it is reasonable to ask the user to pass in the right number and combination of axes. They already have to do that if you create the axes first.

moorepants:
A development setup of opty is explained in the README.

Peter230655:
> > If the user give the axes, he must give 'enough' axes, just two may not be the right amount.
> 
> As long as you document this, I think it is reasonable to ask the user to pass in the right number and combination of axes. They already have to do that if you create the axes first.

This seems to be the problem with the drone simulation. It was more than 10 constraints.
How should I document it? in the method itself?

tjstienstra:
In addition to what Jason has already mentioned. You can easily run personal scripts, like some fun example, by creating a file like `temp_playground.py` in the root directory:
```
.
├── .git
│   ├── ...
│   ├── info
│   │    └── exclude
│   └── ...
├── ...
├── opty
│   ├── tests
│   ├── __init__.py
│   └── ...
├── ...
├── README.md
└── temp_playground.py
``` 
This file can just use `from opty.direct_collocation import Problem` and you can run it using `python -m temp_playground.py`. Just make sure not to add this file to git. To make my life easier I like to update the `.git/info/exclude` to also ignore all `temp_*.py` files. There are of course tons of other options, but this one I personally like.

Peter230655:
> In addition to what Jason has already mentioned. You can easily run personal scripts, like some fun example, by creating a file like `temp_playground.py` in the root directory:
> 
> ```
> .
> ├── .git
> │   ├── ...
> │   ├── info
> │   │    └── exclude
> │   └── ...
> ├── ...
> ├── opty
> │   ├── tests
> │   ├── __init__.py
> │   └── ...
> ├── ...
> ├── README.md
> └── temp_playground.py
> ```
> 
> This file can just use `from opty.direct_collocation import Problem` and you can run it using `python -m temp_playground.py`. Just make sure not to add this file to git. To make my life easier I like to update the `.git/info/exclude` to also ignore all `temp_*.py` files. There are of course tons of other options, but this one I personally like.

**Thanks to both of you!**
I have to think about these suggestions, I am a bit better with Github / git - but they are still far away from being my friends! :-)

Peter230655:
The errors with cycling and with drone
image
image
were as they should be, I believe. I added a check if len(axes) is not correct.

Peter230655:
> My approach to testing edits I make to a package, like opty, is to first create a conda environment with all the development dependencies. We include this file in the repo for such a thing: https://github.com/csu-hmc/opty/blob/master/opty-dev-env.yml. After that, I install the development version of opty in a way that any import of opty will always reflect the current state of the opty repo on my computer.
> 
> ```shell
> $ cd /path/to/opty/repo
> $ conda env create -f opty-dev-env.yml
> $ conda activate opty-dev
> $ python -m pip install --no-deps -e .  # installs version in the current directory, i.e. the development version
> ```
> 
> Now as long as I have that conda environment activated it will use the opty version in `/path/to/opty/repo`.
> 
> So then, if I want to build the docs, just:
> 
> ```shell
> $ cd docs
> $ make html
> ```
> 
> All of the constraint violations plots should generate using your edits to the direct_collocation.py file.

I think I finally understood! Had to think about it for a good while.

Peter230655:
Prints a warning if axes are given, with sharex = True (which makes no sense here).
If sharex = True it will generater plots, but they are nonsense.

Peter230655:
I rotate the labels -90°. This way, they are always lined up with the ticks and never smear into each other.
If the state variables are defined as functions, their argument is also displayed as mantissa * 10^(exponent). If the argument is _exactly_ 10, the display looks odd sometimes line q(--1.0 * 10^1) instead of q(1.0 * 10^1). I do not know why. 
@moorepants 
- which rotation do you prefer?
- should I keep the numbers as mantissa * 10^(exponent), or better like rounded numbers? Of course the mantissa is also rounded. I have a slight preference for rounded numbers, but this is your program!
What I mean ist this: 50.12345 displayed as 5.01*10^1 or as 50.1

Peter230655:
The failed check with cycling and drone is that they do not give the right number of axes, as before.

opty/direct_collocation.py
```diff
@@ -448,6 +450,35 @@ def plot_constraint_violations(self, vector, axes=None):
...
+        rotation = -90
+
+        # ensure that len(axes) is correct, raise ValuError otherwise
+        if axes is not None:
+            warner = False
+            for i in range(len(axes.ravel())):
+                if axes.ravel()[i]._sharex is not None:
+                    warner = True
+            if warner == True:
+                print('Set sharex=False or remove, it makes no sense here')
```
moorepants, line 471: We typically don't have print statements in library code because their display can't be controlled. Best to avoid warnings and just raise errors.

moorepants: I also don't think we need to warn about things like this. The best approach here is to explain in the docstring what the user must do if they want to pass in their own axes.

moorepants:
> which rotation do you prefer?

I don't know because we need to see how it looks on all the examples and any other test cases you are using.

moorepants:
> The failed check with cycling and drone is that they do not give the right number of axes, as before.

You have to correct that.

Peter230655:
> > The failed check with cycling and drone is that they do not give the right number of axes, as before.
> 
> You have to correct that.

Do you mean I correct the drone example? If Yes, same would be true with the cycling example (?)

opty/direct_collocation.py
Peter230655: > I also don't think we need to warn about things like this. The best approach here is to explain in the docstring what the user must do if they want to pass in their own axes.

If sharex = True, the program will not interrupt, but the plots are meaningless. Do you mean, I just let this happen, and explain in the method, that it should be set to False or removed? Should this be mentioned in the opty Documentation?
If YES, I have to find out how to do this.

moorepants:
> Do you mean I correct the drone example? 

Yes.

> If Yes, same would be true with the cycling example (?)

Yes.

> If yes, would I give the corrrect number of axes, or remove the keyword ax when the function is called?

Give correct number.

> Would I change drone and cycling in the same PR as this one, so the check will run, or change with separate PRs?

All should happen in this PR.

opty/direct_collocation.py
moorepants: > Do you mean, I just let this happen, and explain in the method, that it should be set to False or removed?

Yes, but I don't think you need to explain anything about sharex. `plt.subplots()` has a very long list of arguments and it is not our job to explain them all here. You simply have to state to the user the minimal information they need to know to pass in a correct axes object.

moorepants:
You have mistakenly committed a skateboard example:

image

Peter230655:
> You have mistakenly committed a skateboard example:
> 
> image

This is what I meant with my point one. I remvoed it. I will try again to commit direct_collocation. This git is still difficult for me! :-(

moorepants:
> Shoud I make a PR with normal numbers, not exponential, so you can see the difference and decide which one you prefer?

No, we just iterate on this pr. You make changes, we look at it (and complain :smile:) and you update until we get it right.

Peter230655:
> > Shoud I make a PR with normal numbers, not exponential, so you can see the difference and decide which one you prefer?
> 
> No, we just iterate on this pr. You make changes, we look at it (and complain 😄) and you update until we get it right.

I am glad you complain - this is how I will get better by the time I will be 75! :-)

Peter230655:
Again the same mistake?

Peter230655:
Now o.k.? :-)

tjstienstra:
> Now o.k.? :-)

Yes, you can see that the skateboard example file doesn't show [here](https://github.com/csu-hmc/opty/pull/218/files) anymore.

@moorepants an option is to consider squashing 53b5661cf329c2aab99d11bb1192e7b13822f595 to 458909e27e5430adc11ac636bd555b88868c37f0 resulting in a cleaner commit tree. If preferred there is an easy way to explain how to do so in VS code.

Peter230655:
> > Now o.k.? :-)
> 
> Yes, you can see that the skateboard example file doesn't show [here](https://github.com/csu-hmc/opty/pull/218/files) anymore.
> 
> @moorepants an option is to consider squashing [53b5661](https://github.com/csu-hmc/opty/commit/53b5661cf329c2aab99d11bb1192e7b13822f595) to [458909e](https://github.com/csu-hmc/opty/commit/458909e27e5430adc11ac636bd555b88868c37f0) resulting in a cleaner commit tree. If preferred there is an easy way to explain how to do so in VS code.

Thanks!
Has my direct_collocation.py arrived at the right location?
(I thoght I was getting better with git - and then the most stupid thing happen to me again)

moorepants:
> an option is to consider squashing

Please don't mention the word squashing in Peter's PRs :smile:

Peter230655:
> > an option is to consider squashing
> 
> Please don't mention the word squashing in Peter's PRs 😄

:-)   :-)
I do not get discouraged easily! :-)

moorepants:
Here is the drone example plot:



The 90 rotation will make the spacing between plots a bit extreme, especially if there are more instance constraints. We probably do need an option to replace with `IC0` or something.

There is also an issue with double negative signs.

moorepants:
Here is cycling example:



I would have expected to have equal number of instance constraints on each plot row. Any reason you did not do that?

moorepants:
Here is the double pendulum swing up:
image

A bit odd with the single super wide bar in the second row of instance constraints.

moorepants:
For the pendulum swing up, `h` is available inside `plot_constraint_violations` so `299h` could be replaced with the time value, for example. Might be more useful to have that print as a value of time.

Peter230655:
> Here is the drone example plot:
> 
> image
> 
> The 90 rotation will make the spacing between plots a bit extreme, especially if there are more instance constraints. We probably do need an option to replace with `IC0` or something.
> 
> There is also an issue with double negative signs.

- What is IC0?
- I mentioned these double negatives, which only appear if the number is exactly 10. If you change it to 10.001 it goes away. I have no idea where it is coming from. Both are reasons, I prefer the 'normal number' display.

Peter230655:
> Here is cycling example:
> 
> image
> 
> I would have expected to have equal number of instance constraints on each plot row. Any reason you did not do that?

I simply did not think about it, I will try this right away. I believe, I will need some range of bars_per_plot, I will take 6<= bpp <= 10, can be changed easily.

Peter230655:
> For the pendulum swing up, `h` is available inside `plot_constraint_violations` so `299h` could be replaced with the time value, for example. Might be more useful to have that print as a value of time.

h = solution[-1], correct. I hope I can change this,

moorepants:
> What is IC0?

Just a made up label, we have to come up with something to make it clear what instance constraint it is.

moorepants:
> h = solution[-1], correct? I hope I can change this,

Yes, but only if the problem is variable time. This is not necessary in this PR, but could be done if you like (or in a later PR).

Peter230655:
> > What is IC0?
> 
> Just a made up label, we have to come up with something to make it clear what instance constraint it is.

What I am trying to do is this:
 - 6 <= bars_per_plot <= 10
 - then I get bar_per_plot is this range, so the last plot has as many plots as possible. 
The double pendulum looks like this, it has 11 instance_constraints
image

I am nearly finished, checking with sumulations.

Peter230655:
> > What is IC0?
> 
> Just a made up label, we have to come up with something to make it clear what instance constraint it is.

I can easily set a keyword rotation - but your are not too keen on this. I can switch to 'normal numbers, not these exponentials, they are mistly shorter.

moorepants:
I think the rotation should be back at 45 deg or whatever it was. If there are never more than like 6 bars, then rotated is better because it will not take as much space between plots. I think 90 deg will break down.

moorepants:
> I can easily set a keyword rotation

I'd prefer if we avoid setting matplotib kwargs via these plot commands. If someone wants to do that, they should just create their figure manually.

Peter230655:
> I think the rotation should be back at 45 deg or whatever it was. If there are never more than like 6 bars, then rotated is better because it will not take as much space between plots. I think 90 deg will break down.

it was -10° originally, which I believe was too small.
I will set it to -45°, and look what it looks like.
I allow maximally 10 bars per plot rght now.

Peter230655:
drone with -45°.
I will switch to normal number and show you, too
image

Peter230655:
Now with 'normal numbers', no exponentials. I think in the range of numbers I have seen in the examples, this is better than exponentials - where I also do not know how to solve the 'double minus' problem.
image

moorepants:
Those both look good. I'm fine with the normal numbers as long as the number of figures after the decimal are reduced (like you have it).

Peter230655:
> Those both look good. I'm fine with the normal numbers as long as the number of figures after the decimal are reduced (like you have it).

I will switch to normal numbers, and play around with exponentials and variable h next.
I will make a PR later today or tomorrow morning.

examples-gallery/plot_drone.py
```diff
@@ -230,7 +230,7 @@
 
 # %%
 # Plot the constraint violations.
-fig, axes = plt.subplots(2, figsize=(12.8, 9.6),
```
moorepants, line 233: These were ratio 3:4 aspect, I suspect on purpose. Is there a reason to change that?

opty/direct_collocation.py
```diff
@@ -458,23 +513,51 @@ def plot_constraint_violations(self, vector, axes=None):
         con_nodes = range(1, self.collocator.num_collocation_nodes)
 
         plot_inst_viols = self.collocator.instance_constraints is not None
+        num_inst_viols = len(instance_violations)
+
         if axes is None:
-            fig, axes = plt.subplots(1 + plot_inst_viols, squeeze=False,
-                                     layout='compressed')
+            fig, axes = plt.subplots(1 + num_plots, squeeze=False,
+                layout='compressed', figsize=(8, 2.0*(num_plots+1)))
```
moorepants, line 520: It is better to leave the fig size for the user to create manually. Setting figsize inside a function like this can result in giant figures.

examples-gallery/plot_drone.py
Peter230655: > These were ratio 3:4 aspect, I suspect on purpose. Is there a reason to change that?

No, I did not give any thought. I just changed to make it run with my proposal

moorepants:
I'll merge if you remove the figsize kwarg inside the function and try to simplify the docstring explanation to how many rows they'll need in the axes.

opty/direct_collocation.py
```diff
@@ -438,6 +440,13 @@ def plot_constraint_violations(self, vector, axes=None):
         axes : ndarray of AxesSubplot
             A matplotlib axes with the constraint violations plotted.
 
+            If axes are given by the user:
+            len(axes) = 2 if len(self.collocator.instance_constraints) <= bars_per_plot.
+            len(axes) = len(self.collocator.instance_constraints) // bars_per_plot + 2
+                        if len(self.collocator.instance_constraints) % bars_per_plot != 0.
+            len(axes) = len(self.collocator.instance_constraints) // bars_per_plot + 1
+                        if len(self.collocator.instance_constraints) % bars_per_plot = 0.
```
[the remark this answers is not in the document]
Peter230655: > There is no way for the user to know what the value of `bars_per_plot` is in the docstring. Just state the number. This could be more simply stated in general. Something like, "six instance constraints will be plotted per axes, so you'll need to pass in enough rows to accomodate the number of instance constraints".

My approach was this: I select between 6 to 10 bars per plot to prevent situations like with the double pendulum., where there is only on bar on the last plot. If the wrong number of axes is given, the method will tell the user how many are needed - unless he only gives one, which is always wrong.

opty/direct_collocation.py
```diff
@@ -458,23 +513,51 @@ def plot_constraint_violations(self, vector, axes=None):
...
         axes[0].set_xlabel('Node Number')
         axes[0].set_ylabel('EoM violation')
 
+        # reduce the instance constrtaints to 2 significant digits.
+        instance_constr_plot = []
+        for exp1 in self.collocator.instance_constraints:
+            for a in sm.preorder_traversal(exp1):
+                if isinstance(a, sm.Float):
+                    exp1 = exp1.subs(a, round(a, 2))
+            instance_constr_plot.append(exp1)
```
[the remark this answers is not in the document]
Peter230655: > Python has printing controls that work like this:
> 
> ```
> In [7]: '{:1.3f}'.format(1.2394359593939)
> Out[7]: '1.239'
> ```
> 
> is there a reason to not use something like that instead of literally rounding the numerical values?

Timo suggested this to me, so I used it, since ti worked. I had no idea how to do such things.

Peter230655:
> I'll merge if you remove the figsize kwarg inside the function and try to simplify the docstring explanation to how many rows they'll need in the axes.

Great!
Removing the kwarg is easy, frankly I do not recall why I did it.
As regards the number of axes, would this be o.k.:
"Select at least 3 axes. If more are needed, the method will tell the use how many are needed." ?

opty/direct_collocation.py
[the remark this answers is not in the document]
moorepants: Ok, I see now there may not be a straight forward way to use Python printing methods on SymPy Floats in an expression. I thought there should be a setting on the StrPrinter that would control display of decimals, but there doesn't seem to be one.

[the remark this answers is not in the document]
moorepants: Maybe Timo already showed this (I struggle to find his comment), but this is a way to control printing of floats:

```
In [47]: from mpmath.libmp import prec_to_dps, to_str as mlib_to_str

In [48]: class MyPrinter(sm.StrPrinter):
    ...:     _default_settings = {
    ...:         "order": None,
    ...:         "full_prec": "auto",
    ...:         "sympy_integers": False,
    ...:         "abbrev": False,
    ...:         "perm_cyclic": True,
    ...:         "min": None,
    ...:         "max": None,
    ...:         "dps": None,
    ...:     }
    ...: 
    ...:     def _print_Float(self, expr):
    ...:         prec = expr._prec
    ...:         if prec < 5:
    ...:             dps = 0
    ...:         else:
    ...:             dps = prec_to_dps(expr._prec)
    ...:         if self._settings['dps']:
    ...:             dps = self._settings['dps']
    ...:         if self._settings["full_prec"] is True:
    ...:             strip = False
    ...:         elif self._settings["full_prec"] is False:
    ...:             strip = True
    ...:         elif self._settings["full_prec"] == "auto":
    ...:             strip = self._print_level > 1
    ...:         low = self._settings["min"] if "min" in self._settings else None
    ...:         high = self._settings["max"] if "max" in self._settings else None
    ...:         rv = mlib_to_str(expr._mpf_, dps, strip_zeros=strip, min_fixed=low, max_fixed=high)
    ...:         if rv.startswith('-.0'):
    ...:             rv = '-0.' + rv[3:]
    ...:         elif rv.startswith('.0'):
    ...:             rv = '0.' + rv[2:]
    ...:         if rv.startswith('+'):
    ...:             # e.g., +inf -> inf
    ...:             rv = rv[1:]
    ...:         return rv
    ...: 

In [49]: MyPrinter(settings={'dps': 3}).doprint(f(1.329294))
Out[49]: 'f(1.33)'
```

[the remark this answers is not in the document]
moorepants: Hah, I found Timo's solution which was the same.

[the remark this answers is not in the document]
Peter230655: > Maybe Timo already showed this (I struggle to find his comment), but this is a way to control printing of floats:
> 
> ```
> In [47]: from mpmath.libmp import prec_to_dps, to_str as mlib_to_str
> 
> In [48]: class MyPrinter(sm.StrPrinter):
>     ...:     _default_settings = {
>     ...:         "order": None,
>     ...:         "full_prec": "auto",
>     ...:         "sympy_integers": False,
>     ...:         "abbrev": False,
>     ...:         "perm_cyclic": True,
>     ...:         "min": None,
>     ...:         "max": None,
>     ...:         "dps": None,
>     ...:     }
>     ...: 
>     ...:     def _print_Float(self, expr):
>     ...:         prec = expr._prec
>     ...:         if prec < 5:
>     ...:             dps = 0
>     ...:         else:
>     ...:             dps = prec_to_dps(expr._prec)
>     ...:         if self._settings['dps']:
>     ...:             dps = self._settings['dps']
>     ...:         if self._settings["full_prec"] is True:
>     ...:             strip = False
>     ...:         elif self._settings["full_prec"] is False:
>     ...:             strip = True
>     ...:         elif self._settings["full_prec"] == "auto":
>     ...:             strip = self._print_level > 1
>     ...:         low = self._settings["min"] if "min" in self._settings else None
>     ...:         high = self._settings["max"] if "max" in self._settings else None
>     ...:         rv = mlib_to_str(expr._mpf_, dps, strip_zeros=strip, min_fixed=low, max_fixed=high)
>     ...:         if rv.startswith('-.0'):
>     ...:             rv = '-0.' + rv[3:]
>     ...:         elif rv.startswith('.0'):
>     ...:             rv = '0.' + rv[2:]
>     ...:         if rv.startswith('+'):
>     ...:             # e.g., +inf -> inf
>     ...:             rv = rv[1:]
>     ...:         return rv
>     ...: 
> 
> In [49]: MyPrinter(settings={'dps': 3}).doprint(f(1.329294))
> Out[49]: 'f(1.33)'
> ```

He did show this or something similar, but I liked the simple round(a, 3) much better! :-)

Peter230655:
My idea, based on your (very justified!) complaint about the double pendulum is this: 
bars_per_plot is between 6 and 10.
Say, there are 11 instance constraints, as in the double pendulum. Then there will be two plots for the instance constraints, the first has 6 bars, the second one has 5 bars
Say, there are 16 instance constraints, then both plots have 8 bats
That is, I select the bars_per_plot so, that (bars_per_plot - num_instance_constraints % bars_per_plot) is minimal.
(Of course, I have to separately consider when (bars_per_plot - num_instance_constraints % bars_per_plot) = 0, etc, hence my program looks maybe a bit confusing. Also my execution is likely not the most elegant one)

- I found no way to get the plots right unless I give subspace = (..., ...)
- If a user wants to give his own axes, he should give at least two axes, the program will tell him how many are needed.

Peter230655:
I added figsize(...) to plot_trajectories, so it gives nice pictures without the user giving axes. (I see no way to get by without this)
I changed the description on how the number of axes is determined in plot_constraint_violations a little bit. With the way I determine the number of axis, I see no simple way to tell the user how many are needed.

Peter230655:
> > h = solution[-1], correct? I hope I can change this,
> 
> Yes, but only if the problem is variable time. This is not necessary in this PR, but could be done if you like (or in a later PR).

I may have found a solution to print the correct time with variable h.
I do it in a different branch. ( I am much more sure I found a solution for variable h than I am about this different branch...)

moorepants:
One possible way around it is not to use the `layout='compressed'` flag. That flag breaks down if there are too many subplots in the default size of the figure.

Let's keep the automated figsize and the layout=compressed. I typically do not muck with figsize in library code due to figures becoming too large, but I think this is as good as we can get now.

moorepants:
Thanks!

moorepants:
> Thank YOU! I bet you spent more time coaching me compared to having it done yourself! :-)

I'm a teacher by trade, so that's just the way it always is :)

>  This is now part of opty 1.4.0dev0 ?

Yes

>  I tested it the best I could, but if users find problems I will learn about it so I can correct them?

They will open issues on this repository.

>  Now that you merged it do I delete this branch?

Yes, you can safely delete, but make sure to pull the master from this repository in your local master to get the updates.

Peter230655:
> > Thank YOU! I bet you spent more time coaching me compared to having it done yourself! :-)
> 
> I'm a teacher by trade, so that's just the way it always is :)
> 
> > This is now part of opty 1.4.0dev0 ?
> 
> Yes
> 
> > I tested it the best I could, but if users find problems I will learn about it so I can correct them?
> 
> They will open issues on this repository.
> 
> > Now that you merged it do I delete this branch?
> 
> Yes, you can safely delete, but make sure to pull the master from this repository in your local master to get the updates.

Yeah, part of your job! :-)
I may have found a way to give the time with variable h. Seems to work so far, but still not all exceptions caught. I will make a PR when I feel much more confident.

Peter230655:
I think, I found a pretty good way to print the rounded number with variable h. I simply look around the symbols tree (I guess that is what it is called). It works for many ways of giving the duration, like
duration = (num_nodes - 1) * h
duration = h * (num_nodes - 1)
duration = (num_nodes - 1) * h / 2.0

It does NOT work for
duration = (num_nodes - 1) * h / 2   In this case it prints  125h/2 if num_nodes is 126

The variable inteval may have any name, not just h

I'd like to make a PR to see if it passes all the test (I think it would..) but I am afraid, I am swamping you with PRs!

moorepants:
> The variable inteval may have any name, not just h

h is stored in Problem, so you'll always have the symbol inside problem.

Peter230655:
> > The variable inteval may have any name, not just h
> 
> h is stored in Problem, so you'll always have the symbol inside problem.

At the top of class Problem, I set
self.zeit = node_time_interval   (I used zeit = German for time, to ensure I would not mess up anything, as presumably you not use German names )

moorepants:
>  self.zeit = node_time_interval (I used zeit = German for time, to ensure I would not mess up anything, as presumably you not use German names )

presumably not, but you are teaching me new words every day :)

Peter230655:
> > self.zeit = node_time_interval (I used zeit = German for time, to ensure I would not mess up anything, as presumably you not use German names )
> 
> presumably not, but you are teaching me new words every day :)

:-)  :-) German is very similar to Dutch.
It seems to work. Should I make a PR so you can look at it at your convenience?

moorepants:
Just a note for future, in the next sympy version you can do:

```python
StrPrinter(settings={"dps": 3}).doprint(Function('f')(1.329294))
```

which returns

```
f(1.33)
```

Peter230655:
> Just a note for future, in the next sympy version you can do:
> 
> ```python
> StrPrinter(settings={"dps": 3}).doprint(Function('f')(1.329294))
> ```
> 
> which returns
> 
> ```
> f(1.33)
> ```

Thanks!
I did not know this at all, never seen StrPrinter.
This works with the 'development version' of sympy?

moorepants:
Yes, we just merged the change.

moorepants:
>  I did not know this at all, never seen StrPrinter.

Both Timo and I showed you this above at least twice.

Peter230655:
> > I did not know this at all, never seen StrPrinter.
> 
> Both Timo and I showed you this above at least twice.

Seems I am getting old! :-(

PR #221, opened 2024-09-07
Handle variable h

Peter230655:
I case of variable h its value is used in the plots. e.g. if, say. num_nodes = 251 and $h_{optimal} = 0.1$ and duration =(num_nodes - 1) * h it will print, say, $x_0(25.0)$ instead of $x_0(250h)$
The times in the instance constraints should be given as t * h, where t is an integer or a float to ensure proper printing.
t * h / 2.0 will also work, but r * h / 2 will not always work. In case it does not work the 'old' way is given.
It will be rounded to max. two digits after the decimal point. (I assume in sympy the 'mul objects' are handled differently for floats and integers, but I did not study this, just trial and error)

### DIFF
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -151,7 +151,7 @@ class Problem(cyipopt.Problem):
             ``{x(t): (-1.0, 5.0)}``.
 
         """
-        # needed for plot_constraint_violations method below
+        # needed for the plot_constraint_violations method below
         self.zeit = node_time_interval
 
         self.collocator = ConstraintCollocator(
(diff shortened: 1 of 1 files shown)


opty/direct_collocation.py
```diff
@@ -151,6 +151,8 @@ def __init__(self, obj, obj_grad, equations_of_motion, state_symbols,
             ``{x(t): (-1.0, 5.0)}``.
 
         """
+        # needed for the plot_constraint_violations method below
+        self.zeit = node_time_interval
```
moorepants, line 155: This should be stored already on ` self.collocator.node_time_interval`.

Peter230655: > This should be stored already on ` self.collocator.node_time_interval`.

I did not see this, I will change it.

opty/direct_collocation.py
```diff
@@ -459,13 +462,18 @@ def plot_constraint_violations(self, vector, axes=None):
         - r : number of unknown parameters
         - s : number of unknown time intervals
 
+        If node_time_interval is a sympy.Symbol, the times in the instance
+        constraints should be given as alpha * h, where alpha is an integer
+        or a float and h is the name of the node_time_interval symbol. This
+        will ensure that the time is given correctly in the plot.
```
moorepants, line 468: This doesn't belong in the docstring of `plot_constraint_violations`. If you think more explanation is needed on how to work with a variable time interval, that should be documented in the `ConstraintCollocator` primary docstring.

Peter230655: > This doesn't belong in the docstring of `plot_constraint_violations`. If you think more explanation is needed on how to work with a variable time interval, that should be documented in the `ConstraintCollocator` primary docstring.

I can move it there, or delete it. What is your preference? After all, YOUR program!

moorepants: I welcome improvements to the documentation and you think it needs better or more explanation you can add it. But maybe just do it in another PR as it is unrelated to this change.

Peter230655: > I welcome improvements to the documentation and if you think it needs better or more explanation you can add it. But maybe just do it in another PR as it is unrelated to this change.

I will remove it. What I did is a minuscule thing in opty, and explanation likely not needed.

moorepants:
This looks good. Thanks.

Peter230655:
> This looks good. Thanks.

Thank YOU! I must admit it feels good to contribute even in smaller ways!

Should I make a new PR to remove the figsize(.., ...) ?

PR #226, opened 2024-09-10
Try Python 3.13 in CI.

moorepants:


moorepants:
Can't test with Python 3.13 in our CI setup yet:

```
  Warning: 
  LibMambaUnsatisfiableError: Encountered problems while solving:
    - nothing provides _python_rc needed by python-3.13.0rc1-h3b4618f_0_cp313t
  
  Could not solve for environment specs
  The following package could not be installed
  └─ python 3.13**  is not installable because it requires
     └─ _python_rc, which does not exist (perhaps a missing channel).
  
  
  
  LibMambaUnsatisfiableError: Encountered problems while solving:
    - nothing provides _python_rc needed by python-3.13.0rc1-h3b4618f_0_cp313t
  
  Could not solve for environment specs
  The following package could not be installed
  └─ python 3.13**  is not installable because it requires
     └─ _python_rc, which does not exist (perhaps a missing channel).
```

moorepants:
CI means "continuous integration".

Peter230655:
> CI means "continuous integration".

The set up that does the testing for PRs for opty?

moorepants:
Yes.

PR #230, opened 2024-09-13
Add example of parameter estimation from multiple measurement trials

Peter230655:
In a simple spring / damper model the spring constant and the dampening coefficient are to be estimated using noisy measurements.
The model is set up four times corresponding to four noisy measurements.
I create these noisy measurements by integrating the four systems and adding random numbers to each point of the solutions and a random shift.
As the method plot_trajectories does not work if no input trajectories are given, I plot them 'manually'.
(I hope I understood the idea correctly)


examples-gallery/plot_parameter_estimation.py
```diff
@@ -0,0 +1,227 @@
...
+
+instance_constraints = (
+    x1.subs({t: t0}) - x0[0],
+    x2.subs({t: t0}) - x0[1],
+    x3.subs({t: t0}) - x0[2],
+    x4.subs({t: t0}) - x0[3],
+    u1.subs({t: t0}) - x0[4],
+    u2.subs({t: t0}) - x0[5],
+    u3.subs({t: t0}) - x0[6],
+    u4.subs({t: t0}) - x0[7],
```
moorepants, line 164: You generally don't want instance constraints in parameter estimation problems. This overly constrains the solution.

Peter230655: > You generally don't want instance constraints in parameter estimation problems. This overly constrains the solution.

Sorry, I missed this in the PR I just started. Will correct it.

moorepants:
This is the right idea and very cool to see that it works!

examples-gallery/plot_parameter_estimation.py
moorepants: My bet is that if you remove these, you'll get better parameter estimates.

Peter230655:
> This is the right idea and very cool to see that it works!

Removing the instance control did two things:
- it made the estimates worse (I know their true value :-) )
- it gives a mistake in plot_constraint_violations
I will fix the error in the method over the weekend and fix it.
I think, no much merging this simulation until the method plot_constraint_violation is fixed?

examples-gallery/plot_parameter_estimation.py
```diff
@@ -0,0 +1,205 @@
...
+Parameter Estimation
+====================
+
+Four noisy measurements of the location of a simple system consisting of a mass
+connected to a fixed point by a spring and a damper. The movement is in a
+horizontal direction. The the spring constant and the damping coefficient
+are to be estimated.
+
+The idea is to set up four sets of eoms, one for each of the measurements, with
+identical parameters, and let opty estimate the parameters.
```
moorepants, line 13: Maybe we should add an explanation of why we have to do such a thing. Something like:

For parameter identification, it is common to collect measurements of a system's trajectories from distinct expeirments. For example, if you are identifying the parameters of a mass-spring-damper system you will excite the system with different initial conditions multiple times. The data cannot simply be stacked and the identification run because the measurement data would be discontinuous between trials. A work around in opty is to create a set of differential equations with unique state variables for each measurement trial that all share the same constant parameters. You can then identify the parameters from all measurement trials simultaneously by passing the uncoupled differential equations to opty.

examples-gallery/plot_parameter_estimation.py
```diff
@@ -0,0 +1,205 @@
...
+
+**State Variables**
+
+- :math:`x_1`: position of the mass of the first system [m]
+- :math:`x_2`: position of the mass of the second system [m]
+- :math:`x_3`: position of the mass of the third system [m]
+- :math:`x_4`: position of the mass of the fourth system [m]
```
moorepants, line 20: These are really all the same system. I would call $x_1$ the state variable associated with the first measurement trial.

examples-gallery/plot_parameter_estimation.py
```diff
@@ -0,0 +1,205 @@
...
+forces = [(P1, -k * (x1 - l0) * N.x - c * u1 * N.x),
+    (P2, -k * (x2 - l0) * N.x - c * u2 * N.x), (P3, -k * (x3 - l0) * N.x - c * u3 * N.x),
+    (P4, -k * (x4 - l0) * N.x - c * u4 * N.x)]
+
+kd = sm.Matrix([u1 - x1.diff(), u2 - x2.diff(), u3 - x3.diff(), u4 - x4.diff()])
+
+q_ind = [x1, x2, x3, x4]
+u_ind = [u1, u2, u3, u4]
+
+KM = me.KanesMethod(N, q_ind, u_ind, kd_eqs=kd)
```
moorepants, line 79: I think the example would be clearer if you didn't use KanesMethod. You can write these EoMs directly with SymPy and it is readable.

```python
Matrix([
    x1.diff(t) - u1,
    m*u1.diff(t) + c*u1 + k*(x1 - l0),
])
```

examples-gallery/plot_parameter_estimation.py
```diff
@@ -0,0 +1,205 @@
...
+fr, frstar = KM.kanes_equations(bodies, forces)
+eom = kd.col_join(fr + frstar)
+sm.pprint(eom)
+
```
moorepants, line 83: make a new cell here and print the eoms

examples-gallery/plot_parameter_estimation.py
```diff
@@ -0,0 +1,205 @@
...
+        np.random.randn(1)*2)
+
+fig, ax = plt.subplots(figsize=(10, 5))
+for i in range(4):
+    ax.plot(times, noisy[i], label=f'measurement {i+1}', lw=0.5)
+plt.xlabel('Time')
+ax.set_title('Measurements of the position of the mass')
+ax.legend();
+
+plt.show()
```
moorepants, line 119: show is only needed once at the very end

examples-gallery/plot_parameter_estimation.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+# %%
+# Set up the Estimation Problem.
+# --------------------------------
+#
+# If some measurement is considered more reliable, its weight w can be increased.
+#
+# objective = :math:`\int_{t_0}^{t_f} (w_1 (x_1 - noisy_{x_1})^2 + w_2 (x_2 - noisy_{x_2})^2 + w_3 (x_3 - noisy_{x_3})^2 + w_4 (x_4 - noisy_{x_4})^2)\, dt`
```
moorepants, line 128: Using $x_1 - x_1^m$ where "m" is for measured would make these equations simpler

examples-gallery/plot_parameter_estimation.py
Peter230655: > make a new cell here and print the eoms

I will incorporate the changes.
What I do not fully understand: why not use Kane's method? After all I will print the eoms.

examples-gallery/plot_parameter_estimation.py
```diff
@@ -0,0 +1,205 @@
...
+    interval_value,
+    known_parameter_map=par_map,
+    instance_constraints=instance_constraints,
+    time_symbol=me.dynamicsymbols._t,
+)
+
+# %%
+# Initial guess.
+#
+initial_guess = np.array(list(np.random.randn(8*num_nodes + 2)))
```
moorepants, line 182: In parameter identification problems you have the measured state values, so there is really no reason to not use the measurements as the initial guess for the state trajectories.

examples-gallery/plot_parameter_estimation.py
```diff
@@ -0,0 +1,205 @@
...
+# %%
+# Solve the Optimization Problem.
+#
+solution, info = problem.solve(initial_guess)
+print(info['status_msg'])
```
moorepants, line 187: new cell before the plot command, basically a new cell should be after anything you print or plot

examples-gallery/plot_parameter_estimation.py
moorepants: If you can write the equations directly in 8 lines of code, there is no need to obscure that behind all the kanesmethod stuff. The purpose of KanesMethod is to give you the equations of motion, but if we already know the equations of motion, why use it?

moorepants: For example the EoM of a single penudulum is dead simple, so I just write them here:

https://opty.readthedocs.io/stable/examples/plot_pendulum_swing_up_fixed_duration.html#sphx-glr-examples-plot-pendulum-swing-up-fixed-duration-py

and a reader doesn't have to know anything about KanesMethod to understand that example. Not all systems need an advanced multibody dynamcis tool to find the EoMs.

examples-gallery/plot_parameter_estimation.py
```diff
@@ -0,0 +1,205 @@
...
+
+t0, tf = 0, 20
+num_nodes = 500
+times = np.linspace(t0, tf, num_nodes)
+t_span = (t0, tf)
+
+x0 = np.array([3, 3, 3, 3, 0, 0, 0, 0])
+pL_vals = [1.0, 0.25, 1.0, 1.0]
+
+resultat1 = solve_ivp(gradient, t_span, x0, t_eval = times, args=(pL_vals,))
```
moorepants, line 103: A more realistic measurement data set would be collected from trials that have different initial conditions. You often don't have control of the initial conditions in the experiments, you simply measure whatever the state happens to be.

examples-gallery/plot_parameter_estimation.py
moorepants: You also typically have some reasonable guess for the parameters and the parameters needs to be bounded.

moorepants:
I removed instance constraints, added bounds to the unknown parameters, set the initial guess of x to the measured states and set the initial guess for u's to zeros, and gave reasonable guesses for the unknown parameters and I get a consistent optimal solution. I think making those updates would make this more realistic and solve well. Additionally, we should use different initial conditions for each of the four measurement trials.

Peter230655:
> I removed instance constraints, added bounds to the unknown parameters, set the initial guess of x to the measured states and set the initial guess for u's to zeros, and gave reasonable guesses for the unknown parameters and I get a consistent optimal solution. I think making those updates would make this more realistic and solve well. Additionally, we should use different initial conditions for each of the four measurement trials.

I will do this. But could you plot the constraint violations?

moorepants:
It would be helpful to make the final plot be a comparison of the optimal trajectories vs the measured trajectories (like done in the other parameter id examples). You can do this by first plotting the measurements on the same number of axes that the `plot_trajectories()` method produces and then pass that axes set into the plot method.

moorepants:
> I will do this. But could you plot the constraint violations?

No there is a bug.

moorepants:
There should be a `Problem.num_instance_constraints` that returns 0 when there are none.

examples-gallery/plot_parameter_estimation.py
```diff
@@ -0,0 +1,205 @@
+# %%
+"""
+
+Parameter Estimation
```
moorepants, line 4: The title and filename are too generic.

1. We use the terms "parameter identification" in opty not "parameter estimation", we should be consistent.
2. This is a special parameter id example, in that it is showing how to use multiple measurement trials that constitute a set of discontinuous data. So the title should reflect this uniqueness (relative to the other simpler parameter id examples)

Peter230655:
I just pushed this to let you know where I am. Not yet been able to incorporate all your suggestions.
NB: I assumed only the positions of the mass are measured, not the speeds. So the measurement has only four plots, the plot_trajectories has 8.
(I have a hell of a time setting up eoms without sympy mechanics! Got the signs wrong all the time!)

Peter230655:
> > I will do this. But could you plot the constraint violations?
> 
> No there is a bug.

I will try to fix the bug in the next few days. I might have caused it when I changed this method a bit.

moorepants:
Already fixed: https://github.com/csu-hmc/opty/pull/232

moorepants:
>  (I have a hell of a time setting up eoms without sympy mechanics! Got the signs wrong all the time!)

Equations are here: https://en.wikipedia.org/wiki/Mass-spring-damper_model

Peter230655:
> > (I have a hell of a time setting up eoms without sympy mechanics! Got the signs wrong all the time!)
> 
> Equations are here: https://en.wikipedia.org/wiki/Mass-spring-damper_model

Thanks!
I studied engineering 100 years ago! I should still be able to do it without Wikipedia! :-(

Peter230655:
> Already fixed: #232

Unbelievable! I would have taken me 2 days!! :-(

examples-gallery/plot_parameter_estimation.py
moorepants: >  Question: Would you not know the initial conditionsd of the experiment, like im my case you would know how much you extended the mas, and at what initial speed it started?

You know the initial conditions only as good as your error filled measurement might say. If you force the initial condition to the value of a measurement that has error, then you are not allowing for solutions that minimize the overall error.

Peter230655: Makes sense!
When I give reasonable bounds it finds good values.
I will try to incorporate all you suggested tomorrow.

Peter230655: Question: does it make sense to estimate continuous parameters this way? What I mean: say, I apply a time varying force to this sytem and try to estimate what it looks like?

moorepants: If I understand you, I think you can.

examples-gallery/plot_parameter_estimation.py
```diff
@@ -0,0 +1,209 @@
+# %%
+"""
+
+Parameter Identification of a Mass-Spring-Damper System.
+========================================================
```
moorepants, line 5: I'd prefer "Parameter Identification from Non-Continuous Measurements" of something similar. The unique aspect of this example relative to what we already have is that we have these independent measurements of the system's motion.

moorepants: Maybe "non-contiguous" is the right word.

moorepants: Parameter Identification from Non-Contiguous Measurements

Peter230655: > I'd prefer "Parameter Identification from Non-Continuous Measurements" of something similar. The unique aspect of this example relative to what we already have is that we have these independent measurements of the system's motion.

Will do so and push

moorepants: Also change the filename to `plot_non_contiguous_parameter_estimation.py`

examples-gallery/plot_non_continuous_parameter_estimation.py
```diff
@@ -0,0 +1,209 @@
+# %%
+"""
+
+Parameter Identification from Non-Continuous Measurements.
```
moorepants, line 4: You didn't agree that "contiguous" was the correct word?

contiguous /kən-tĭg′yoo͞-əs/
adjective

    Sharing an edge or boundary; touching.
    Neighboring; adjacent.
    Connecting without a break.

moorepants: I guess it can be either, but we have continuous measurements, they are just not contiguous in time.

Peter230655: > You didn't agree that "contiguous" was the correct word?
> 
> contiguous /kən-tĭg′yoo͞-əs/ adjective
> 
> ```
> Sharing an edge or boundary; touching.
> Neighboring; adjacent.
> Connecting without a break. 
> ```

:-)  :-)
I thought is was a typo. Knowing you I should have known better! Will change and push.

moorepants: It doesn't' really matter, I just get nit picky on wording. I'm working an journal article this morning doing the final editing and my mind is in the mode of being extremely precise with wording.

Peter230655: > It doesn't' really matter, I just get nit picky on wording. I'm working an journal article this morning doing the final editing and my mind is in the mode of being extremely precise with wording.

Can I get a copy when you are done with it?

moorepants: Sure.

Peter230655: Thanks!

Peter230655:
I modified this example a bit: I added a force = F1*sin(omegat*t) and also estimate F1 and omega.
Also I made the number of measurements variable.
Would you prefer this 'enlarged' simulation?
If yes, would I simply push to this PR, or close this PR and make a new one?
Thanks!

moorepants:
> I modified this example a bit: I added a force = F1 * sin(omega * t) and also estimate F1 and omega.

I'd prefer if we didn't add this to this example and simply demonstrate how to do parameter identification from multiple non-contiguous measurements. If you want it to be something different I can make an example that focus on the parameter id approach.

Peter230655:
> > I modified this example a bit: I added a force = F1 * sin(omega * t) and also estimate F1 and omega.
> 
> I'd prefer if we didn't add this to this example and simply demonstrate how to do parameter identification from multiple non-contiguous measurements. If you want it to be something different I can make an example that focuses on the parameter id approach.

**NO**, not at all, we leave it as it is!
I play around a lot, the come up with something else, which I feel is interesting, so I let you know.
I do have a tendency to complicate my programs, I want to see how far I can go - and this has a tendency to cover up what should be shown.
NB: I do not know, why it failed on Github test.

moorepants:
>  NB: I do not know, why it failed on Github test.

You have to read the error message.

Peter230655:
> > NB: I do not know, why it failed on Github test.
> 
> You have to read the error message.
Of course I tried to read it - problem is understanding :-(
image

examples-gallery/plot_non_contiguous_parameter_estimation.py
```diff
@@ -0,0 +1,209 @@
...
+
+measurements = []
+np.random.seed(123)
+for i in range(4):
+    measurements.append(resultat[:, i] + np.random.randn(resultat.shape[0]) * 0.5 +
+        np.random.randn(1)*2)
+
+# %%
+# Set up the Identification Problem.
+# --------------------------------
```
moorepants, line 114: The error is here and it is exactly what it says in the error message.

Peter230655: > The error is here and it is exactly what it says in the error message.

Fair enough! I only looked at the 'big' title. I will correct and push

PR #231, opened 2024-09-13
make plot_trajectories work if no input trajectories are given

Peter230655:
plot_trajectories tries to put the title _Input Trajectories_ on a non existing axis if no input trajectories are present.
I check whether they are present.

### DIFF
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -418,7 +418,8 @@ class Problem(cyipopt.Problem):
             ax.set_ylabel(sm.latex(symbol, mode='inline'))
         ax.set_xlabel('Time')
         axes[0].set_title('State Trajectories')
-        axes[self.collocator.num_states].set_title('Input Trajectories')
+        if self.collocator.num_input_trajectories > 0:
+            axes[self.collocator.num_states].set_title('Input Trajectories')
 
         return axes
 


moorepants:
LGTM

Peter230655:
Thanks!!
Since you merged this, let me push a version of plot_estimate_parameters which uses plot_trajectories.
I will push tomorrow morning.

moorepants:
Are you suggesting to plot the solutions to the unknown constants? Why would you plot single values like that? I can't think of a way to plot them that would have any utility.

Peter230655:
Maybe part habit, partly I wanted to see how well the solutions with the estimated parameters would fit the 'true'
solution. That fit was actually better than the accuracy of the estimation of the dampening coefficient.
Accuracy of the solutions is within 2%, dampening is of by 15% as I recall. No idea why.
But you are right, trajectories not needed. I will remove and push again.

Peter230655:
> Are you suggesting to plot the solutions to the unknown constants? Why would you plot single values like that? I can't think of a way to plot them that would have any utility.

At least like this I found the small problem in plot_trajectories. :-)

moorepants:
In general, there is no "true" solution for the unknown constant parameters unless you artificially create the data used for the parameter identification.

Peter230655:
> In general, there is no "true" solution for the unknown constant parameters unless you artificially create the data used for the parameter identification.

I artificially created the measurement data as described above, how else should I get 'measurement data'?
Is the basic approach correct, that is did I understand you correctly?
The convergence is quite fast, actually.

moorepants:
> how else should I get 'measurement data'?

You would have to take measurements of something in real life.

Peter230655:
> > how else should I get 'measurement data'?
> 
> You would have to take measurements of something in real life.

I am a retired salesman, I do not have measurement instruments. :-)
To me it is interesting how well opty found trajectories from very noisy 'measurements', which are close to the 'true' trajectory - known in the case on hand.

moorepants:
I know that, I'm not implying you do so, but the point I'm trying to make is that a parameter identification problem, in general, has no known solution, so it does not make sense to add a plot function to opty to compare it against a nonexistent true solution.

Peter230655:
> I know that, I'm not implying you do so, but the point I'm trying to make is that a parameter identification problem, in general, has no known solution, so it does not make sense to add a plot function to opty to compare it against a nonexistent true solution.

Now I think, I got your point. I will change things a bit a push.

PR #232, opened 2024-09-14
Ensure number of instance constraints is valid and use inside plot me…

moorepants:
…thod. Remove ill formated rst.

### DIFF
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -438,18 +438,9 @@ class Problem(cyipopt.Problem):
         Returns
         =======
         axes : ndarray of AxesSubplot
-            A matplotlib axes with the constraint violations plotted.
-
-            The number of axes needed is calculated like this:
-            len(axes) = 2 if len(self.collocator.instance_constraints) <= bars_per_plot.
-            len(axes) = len(self.collocator.instance_constraints) // bars_per_plot + 2
-                        if len(self.collocator.instance_constraints) % bars_per_plot != 0.
-            len(axes) = len(self.collocator.instance_constraints) // bars_per_plot + 1
-                        if len(self.collocator.instance_constraints) % bars_per_plot = 0.
-
-            If the uses gives at least two axis, the method will tell the user
-            how many are needed, unless the correct amount is given.
-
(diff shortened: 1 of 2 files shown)


Peter230655:
Strange: I still get an error with plot_constraint_violations when I do not give instance_constraints.

moorepants:
You'll have to post error message and reproducing code for us to evaluate it then.

Peter230655:
> You'll have to post error message and reproducing code for us to evaluate it then.

After you merged the fix, I updated my fork, and then I wrote _git pull origin master_. 
So, my master should contain the fix?
Thrn I simply ran the parameter identification simulation without instance constrains, and I got this error

```
ile c:\users\peter\github_repos\opty\opty\direct_collocation.py:463, in Problem.plot_constraint_violations(self, vector, axes)
    [460](file:///C:/users/peter/github_repos/opty/opty/direct_collocation.py:460) # find the number of bars per plot, so the bars per plot are arroximately
    [461](file:///C:/users/peter/github_repos/opty/opty/direct_collocation.py:461) # the same on each plot.
    [462](file:///C:/users/peter/github_repos/opty/opty/direct_collocation.py:462) hilfs = []
--> [463](file:///C:/users/peter/github_repos/opty/opty/direct_collocation.py:463) len_constr = len(self.collocator.instance_constraints)
    [464](file:///C:/users/peter/github_repos/opty/opty/direct_collocation.py:464) for i in range(6, 11):
    [465](file:///C:/users/peter/github_repos/opty/opty/direct_collocation.py:465)     hilfs.append((i, i - len_constr % i))

TypeError: object of type 'NoneType' has no len()
```
Maybe I did not update correctly?

Peter230655:
Now I do have the version with the fix you made in line 511.
My error occurs in line 463.
Strange...

moorepants:
It is because your code doesn't handle the fact that self.instance_constraints can be None or a dictionary.

Peter230655:
> It is because your code doesn't handle the fact that self.instance_constraints can be None or a dictionary.

I agree, I did not think about it when I made the changes. I will try to fix it in the next few days and make a PR.

moorepants:
I've fixed it already.

PR #236, opened 2024-09-15
check that equations_of_motion are not AEs

Peter230655:
Check whether the equations of motions contain differential equation, and rais a ValueError if they are AEs.
I used @tjstienstra idea, which seems to be VERY fast!
Should take care of issue # 228

### DIFF
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -153,7 +153,10 @@ class Problem(cyipopt.Problem):
         """
 
         if equations_of_motion.has(sm.Derivative) == False:
-            raise ValueError('The equations of motion must be DEs or DAEs, not AEs.')
+            raise ValueError('No time derivatives are present.' +
+                ' The equations of motion must be ordinary ' +
+                'differential equations (ODEs) or ' +
+                'differential algebraic equations (DAEs).')
 
         self.collocator = ConstraintCollocator(
             equations_of_motion, state_symbols, num_collocation_nodes,
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -152,6 +152,9 @@ class Problem(cyipopt.Problem):
 
         """
 
+        if equations_of_motion.has(sm.Derivative) == False:
+            raise ValueError('The equations of motion must be DEs or DAEs, not AEs.')
+
         self.collocator = ConstraintCollocator(
             equations_of_motion, state_symbols, num_collocation_nodes,
             node_time_interval, known_parameter_map, known_trajectory_map,
--- a/opty/tests/test_direct_collocation.py
+++ b/opty/tests/test_direct_collocation.py
@@ -1555,14 +1555,14 @@ def test_for_algebraic_eoms():
                             omega(0.0),
                             omega(duration))
 
-    # This will test that a compilation works.
+    # This will test that a ValueError is raised.
+    with raises(ValueError) as excinfo:
+        prob = Problem(
+            obj, obj_grad, eom, state_symbols, num_nodes, interval_value,
+            known_parameter_map=par_map,
+            instance_constraints=instance_constraints,
+            time_symbol=t,
+            bounds={T(t): (-2.0, 2.0)},
+        )
 
-    prob = Problem(
-        obj, obj_grad, eom, state_symbols, num_nodes, interval_value,
-        known_parameter_map=par_map,
-        instance_constraints=instance_constraints,
-        time_symbol=t,
-        bounds={T(t): (-2.0, 2.0)},
-    )
-    with raises(ValueError):
-        eom = 'algebraic equations'
+    assert excinfo.type is ValueError
--- a/opty/tests/test_direct_collocation.py
+++ b/opty/tests/test_direct_collocation.py
@@ -1556,17 +1556,13 @@ def test_for_algebraic_eoms():
                             omega(duration))
 
     # This will test that a compilation works.
-    try:
-        prob = Problem(
-            obj, obj_grad, eom, state_symbols, num_nodes, interval_value,
-            known_parameter_map=par_map,
-            instance_constraints=instance_constraints,
-            time_symbol=t,
-            bounds={T(t): (-2.0, 2.0)},
+
+    prob = Problem(
+        obj, obj_grad, eom, state_symbols, num_nodes, interval_value,
+        known_parameter_map=par_map,
+        instance_constraints=instance_constraints,
+        time_symbol=t,
+        bounds={T(t): (-2.0, 2.0)},
     )
-    except ValueError:
-        print('ValueError found correctly')
-    else:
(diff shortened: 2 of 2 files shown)


moorepants:
You'll need to check this on something at least the size of the bicycle model to make sure it doesn't slow down things. I'm generally fine having a "garbage in, garbage out" design, especially when there can be significant performance costs for checking inputs.

Peter230655:
> You'll need to check this on something at least the size of the bicycle model to make sure it doesn't slow down things. I'm generally fine having a "garbage in, garbage out" design, especially when there can be significant performance costs for checking inputs.

The bicycle model does not run with me, and I still could not lokate the error.
I tested this two liner:
```
zeit = time.time()
if eom.has(sm.Derivative == False):
    raise ValueError(.....)
duration = time.time() - zeit
```
with the eom having 23,000 operations, and duration was less than 10^(-15).
Is this testing of duration inadequate?
Should my eom have many more operations?

moorepants:
>  (Of course my eom had some sm.Derivative very 'soon', and I believe has(...) turns true as soon as it has found the first one.)

If that is true, then it should be pretty fast.

opty/direct_collocation.py
```diff
@@ -152,6 +152,9 @@ def __init__(self, obj, obj_grad, equations_of_motion, state_symbols,
 
         """
 
+        if equations_of_motion.has(sm.Derivative) == False:
+            raise ValueError('The equations of motion must be DEs or DAEs, not AEs.')
```
moorepants, line 156:
```suggestion
            raise ValueError('No time derivatives are present, the equations of motion must be ordinary differential equations (ODEs) or differential algebraic equations (DAEs),')
```

moorepants:
This needs a unit test to make sure the error raises.

This will isn't completely robust in the sense that it may find a `Derivative` that is not a time derivative. In the case a user passes in equations with derivatives of variables other than time, the error message is misleading.

Peter230655:
> > (Of course my eom had some sm.Derivative very 'soon', and I believe has(...) turns true as soon as it has found the first one.)
> 
> If that is true, then it should be pretty fast.

As per chatGPT it is true, that it stops as soon as it has found the first term...

moorepants:
I'm not the type that listens to anything ChatGPT may say...

Peter230655:
> This needs a unit test to make sure the error raises.
> 
> This will isn't completely robust in the sense that it may find a `Derivative` that is not a time derivative. In the case a user passes in equations with derivatives of variables other than time, the error message is misleading.

I tested it with a small example, where I simply deleted the .diff(t) in all the terms. This worked, it raised the error.
So, I should look for ```sm.Derivative(time_symbol``` ?

Peter230655:
> I'm not the type that listens to anything ChatGPT may say...

likely smart - but it has helped me a lot with git commands. :-)

moorepants:
> Adjusted the wording of the ValueError, just in case you want to merge this.

It would still need a unit test.

Peter230655:
> > Adjusted the wording of the ValueError, just in case you want to merge this.
> 
> It would still need a unit test.

Would this be a program without time derivaties so it gives the error? (of course I did test it)
If YES, would I push it with this PR?
Any special 'form' required?
**Sorry about all these questions!**

moorepants:
Yes to all three questions.

Peter230655:
> Yes to all three questions.

I wrote a function, which when run will cause the ValueError to be raised - and of course will stop running. 

- Would I add it to opty/opty/tests?
- Does this function need to import all the libraries it needs, or just a function which will cause the ValueError to be raised?

Thanks!

moorepants:
I would recommend having a look at the existing unit tests and using the patterns there to come up with how your test will look.

Peter230655:
> I would recommend having a look at the existing unit tests and using the patterns there to come up with how your test will look.
This is what I did, and there I found this:
image
Would I paste my function at the end of test_direct_collocation.py? (There are many functions there)

moorepants:
Yes.

opty/tests/test_direct_collocation.py
```diff
@@ -1513,3 +1513,60 @@ def test_known_and_unknown_order():
...
+            obj, obj_grad, eom, state_symbols, num_nodes, interval_value,
+            known_parameter_map=par_map,
+            instance_constraints=instance_constraints,
+            time_symbol=t,
+            bounds={T(t): (-2.0, 2.0)},
+    )
+    except ValueError:
+        print('ValueError found correctly')
+    else:
+        raise Exception()
```
moorepants, line 1570: You should use the `raises()` function from pytest. See where it is used elswhere in the file.

Peter230655:
Is this how you meant it? I simply copied it from the only other place it is used in test_direct_collocation.py
Frankly unsure how it works.

opty/tests/test_direct_collocation.py
```diff
@@ -1556,17 +1556,13 @@ def test_for_algebraic_eoms():
...
 
+    prob = Problem(
+        obj, obj_grad, eom, state_symbols, num_nodes, interval_value,
+        known_parameter_map=par_map,
+        instance_constraints=instance_constraints,
+        time_symbol=t,
+        bounds={T(t): (-2.0, 2.0)},
+    )
+    with raises(ValueError):
+        eom = 'algebraic equations'
```
moorepants, line 1568: You'd have to run the line that raises the error in the with statement.

moorepants: https://docs.pytest.org/en/4.6.x/reference.html#pytest-raises

Peter230655:
You ARE a teacher! Your method of not telling me exactly what to do helps me to learn! 

Just curiosity: Would my original way have worked and just been very ugly compared to this way, or plain wrong?

moorepants:
> Just curiosity: Would my original way have worked and just been very ugly compared to this way, or plain wrong?

Probably would have worked. We use pytest for the tests, so no reason not to use these helper functions that are more robust than things we write.

Peter230655:
THANKS!
Stupid question - as often: 
Did this merge the test only, or also the change in direct_collocation.py?
(You were not so sure you wanted it)

Peter230655:
> > Just curiosity: Would my original way have worked and just been very ugly compared to this way, or plain wrong?
> 
> Probably would have worked. We use pytest for the tests, so no reason not to use these helper functions that are more robust than things we write.

Makes sense, plus more 'elegant', I think!

moorepants:
Everything in the PR is merged. You can see all changes in the "files changed" tab.

Peter230655:
> Everything in the PR is merged. You can see all changes in the "files changed" tab.

Thanks! Makes me feel good to contribute - even if not exactly in gigantic ways! :-)
By the time I will be 75 I hope I will know much more, fewer dumbe questions!° :-)

PR #242, opened 2024-09-18
reduced number of frames in animation form 300 to about 75

Peter230655:
By reducing the number of frames, running time reduces to 19.5 sec from 35 sec.
This is the double pendulum on a cart already in examples-gallery, just sped up.


Peter230655:
Here I used slicing to reduce the number of frames in the animation. 
Also use improved initial conditions.

moorepants:
LGTM

Peter230655:
> LGTM

Thanks! :-)

PR #243, opened 2024-09-18
Pendulum swing up fixed duration

Peter230655:
Reducing the frames from 500 to 100 reduced the running time from 38 sec to 15 sec on my PC


Peter230655:
Fewer frames for the animation as per Jason's and Timo's suggestions.

Peter230655:
Something very similar is on examples-gallery, so I close it.

PR #244, opened 2024-09-18
reduced frames in animation, better initial guess

Peter230655:
Reduced number of frames in the animation fromm around 500 to around 167.
Gave better initial guess.
Made the animation perform faster than real time. I believe it still give a good idea what opty does.
This reduced running time from 44 sec to 21 sec on my PC


examples-gallery/plot_parallel_park.py
```diff
@@ -1,3 +1,4 @@
+# %%
```
moorepants, line 1: Not sure why this is added as the first line, should be removed.

examples-gallery/plot_parallel_park.py
```diff
@@ -223,8 +228,18 @@
...
+coords1 = []
+time1 = []
+for i in range(coords.shape[0]):
+    if i % 3 == 0:
+        coords1.append(coords[i, :, :])
+        time1.append(time[i])
+coords1.append(coords[-1, :, :])
+time1.append(time[-1])
+time = np.array(time1)
+coords = np.array(coords1)
```
moorepants, line 242: This is an overly complicated way to skip frames (if that is what it does). It isn't so favorable for such things to be in examples for beginners.

examples-gallery/plot_parallel_park.py
```diff
@@ -185,10 +186,14 @@
 
 # %%
 # Find the optimal solution.
+initial_guess = np.load('parallel_park_solution.npy')
```
moorepants, line 189: If we add these to examples, we need to explain to the reader why we are doing so.

examples-gallery/plot_parallel_park.py
tjstienstra: Just use slicing in `numpy` that would be a need approach. Though an easier approach in this case would be providing a range like `range(0, len(time), 3)` to `FuncAnimation`.

Peter230655: > This is an overly complicated way to skip frames (if that is what it does). It isn't so favorable for such things to be in examples for beginners.

Yes, I wanted to skip frames, and I could not think of a simpler way.

examples-gallery/plot_parallel_park.py
Peter230655: > If we add these to examples, we need to explain to the reader why we are doing so.

I will correct. Just slipped.

examples-gallery/plot_parallel_park.py
moorepants: I thought one of the examples already has exactly what Timo suggests.

Peter230655: > Just use slicing in `numpy` that would be a need approach. Though an easier approach in this case would be providing a range like `range(0, len(time), 3)` to `FuncAnimation`.

Not sure, I understand what slicing means - my Python skill are limited. Could you explain, please.
This "range thing" I understand, I will try!

Peter230655: > I thought one of the examples already has exactly what Timo suggests.
I will look at the examples for this feature, I did not look before.

Peter230655: > Internet search of "numpy slicing" points to: https://numpy.org/doc/stable/user/basics.indexing.html#slicing-and-striding

Now clear! This I even know, just did not think about it :-(

examples-gallery/plot_parallel_park.py
Peter230655: > Not sure why this is added as the first line, should be removed.

I have to add this in VSC so I can run the simulation  from the begining.
I will remove before I push

examples-gallery/plot_parallel_park.py
tjstienstra: @Peter230655 in this case it is even easier to use `ani = animation.FuncAnimation(fig, animate, range(0, len(time), 3),` than numpy slicing. Though learning more about numpy slicing is definitely useful.

Peter230655: > @Peter230655 in this case it is even easier to use `ani = animation.FuncAnimation(fig, animate, range(0, len(time), 3),` than numpy slicing. Though learning more about numpy slicing is definitely useful.

I will try this first thing tomorrow morning!! :-)

Peter230655:
The way you suggested to reduce the number of frames is really MUCH simpler and more elegant.
Will do change on the other PRs, too.

examples-gallery/plot_parallel_park.py
```diff
@@ -252,6 +258,8 @@ def frame(i):
 
 fig, title_text, lines, Pr_path, Pf_path = frame(0)
 
+time = list(time)
+time[-3:] = [time[-1]]*3
```
moorepants, line 262: What are you doing with these lines?

Peter230655: > What are you doing with these lines?

Without them I did not get the time showing on the animation to end at the correct time. It ended ar 29.88 sec (if I recall correctly) instead of at 30 sec.

Peter230655:
There is something identical on examples-gallery, so I close it.

PR #245, opened 2024-09-19
new PR for pendulum swing up variable duration

Peter230655:
I deleted the old PR because there were to many files and I did not know how to handle it.


Peter230655:
Something like this is on examples-gallery, so I close it

PR #246, opened 2024-09-19
New Example: Parameter Identification from Non-Contiguous Measurements

Peter230655:
I made some changes to parameter identification simulation. I deleted the old PR as it contained unwanted files, which I did not know how to remove.


Peter230655:
removed instance constraints. They were not used, but I forgot to remove the code.

moorepants:
I took the liberty to edit the example with the main thing being changing it to have different initial conditions for each measurement. I also removed the bias, as the identification will fail for large bias unless you add such a term in the model.

moorepants:
You can see my edits here: 0961906ea9868a70596adedf098f29fcf7ebcdf4

moorepants:
If you are fine with my edits, I will merge it.

moorepants:
You can do a "git pull origin parameter_identification" on that branch to bring my changes to your computer and then you can run it to see the result.

Peter230655:
> If you are fine with my edits, I will merge it.

It looks MUCH nicer!! Of course I am fine with these changes!

Peter230655:
> I took the liberty to edit the example with the main thing being changing it to have different initial conditions for each measurement. I also removed the bias, as the identification will fail for large bias unless you add such a term in the model.

I wondered, that it worked with the bias. I gues it did because my 'biases' were centered around 0.
Would it be worth to make a model with bias?

Peter230655:
There is a small typo. which I overlooked 50 times:
image

Peter230655:
If I understand this portion correctly:
```
measurements = []
np.random.seed(123)
for i in range(4):
    x0 = 4.0*np.random.randn(8)
    sol = solve_ivp(lambda t, x, args: eval_rhs(*x, *args).squeeze(),
                    (t0, tf), x0, t_eval=times, args=(par_vals,))
    measurements.append(sol.y.T[:, i] + 2.0*np.random.randn(len(sol.t)))
measurements = np.array(measurements)
```
you integrate the complete system four times. Of coure the code looks extremely elegant!
But is there another reason you did it  or am I misunderstanding the code?

moorepants:
>  Would it be worth to make a model with bias?

You could explore that if it interests you. When I was playing with the example I noticed that the identification would fail often if the bias was large relative to the measurements. There are biases in real experiment data, but I think you would need to identify the bias, i.e. include it as a parameter in the model, for things to work consistently.

moorepants:
You simulate the same motion four times and changed the noise applied to the measurements. I'm simulating four times and change both the initial conditions and the noise. It is more typical, in my experience, that real measurements have arbitrary initial states. Only if you have a very controlled experiment is that not the case and I do most experiments with humans who cannot start in the same state.

Peter230655:
I will play around with it, I have lots of time! 

NB: 
I think I told you that I added a force = F1*sin(omega*t)
and tried to get opty to find this unkown input trajectory.
No matter how many measurements I gave, I gave up to 50 or so, it would always get the trajectory to be 'Bang - Bang'.
(of course it would easily find F1 and omega as parameters)

Do you think, this is impossible in principle?
Or would one 'just' need a huge amount of measurements?

moorepants:
> Do you think, this is impossible in principle?

I don't know, I'd have to see how you are constructing the problem.

Peter230655:
I will send you the program I wrote.
How would you prefer it:

- by email?
- as a PR?
- another way?

Thanks!

PR #249, opened 2024-09-22
parameter_identification_biased_measurements

Peter230655:
I try to identify parameters with noisy and biased measurements.
I doubled the set of eoms and made the second set a shifted copy of the first set.
I used the existing simulation for noisy measurements, and basically made these changes:
- made the number of measurements variable
- doubled the set of eoms
-  used sympy mechanics to get the eoms (I feel unsure about speed constraints incorporatring manually). 

My tests fo far seem to show that this will identify c and k closely, independent of the bias.
unclear to me still is this:
- only if I shift the dependent eoms by _twice_ the bias will it estimate the bias correctly.
- the speeds vary rapidly, they certainly do not reflect the 'true' speeds of the system.


Peter230655:
I added friction to be identified. It gets all three parameters accurately. The estimated biases now seem to have no relation to the actually given biases. Somehow it seems the 'freedom of the biases is needed to get the parameters accurately.

Peter230655:
My tests (many of them) seem to show this:
- the parameters are identified closely, comparable to the simulation without bias in examples-gallery. Accuracy dos not seem to depend on the biases of the measurements.
- the biases are not identified correctly, however the ratio "estimated bias of measurement i / actual bias of measurement i" is close to constant (errors in the range of the error of the identified parameters). This makes sense in my opinion.

So, if biases in the measurements cannot be outruled this may be a viable way of doing parameter estimation.

Peter230655:
What does not give me full confidence in this model is this: While the parameters are identified as good as in the model without bias, and the results seem totally independent of the bias, the **speeds** are not at all realistic. They fluctuate rapidly.
Yet, no mater what I tried, the parameters are identified correctly, and the positions fit nicely into the measurements, regardless of bias.

PR #259, opened 2024-11-02
Betts 10 58

Peter230655:
I think, this may be a candidate for examples-gallery for these reasons:
- it deals with the 'discretization' of a partial differential equation. I do not think, there is such an example in examples-gallery.
- in additon to the 11 DEs for the 11 state variables, two AEs are given. opty needs exactly 11 DAEs for 11 state variables. So, I solve the AEs for one of the state variables each (luckily they can be solved symbolically, likely unusual in general) and substitute into the DEs.
- the objective function depends on the state variables at the final point in time only. Other than minimizing the time interval h, I do not think, there is such an objective function in the examples.
- I iterate over the solution three times, using the solution as initial guess for the next turn. I have done this with another example, but here I can actually do it, running time still short, and it shows how the optimization improves. (Betts gives $2.45 \cdot 10^{-3}$ as optimum, but I see no reason, why it should not be zero.)


Peter230655:
I close this PR as I replaced it with PR #309

PR #261, opened 2024-11-05
Betts 10 103 104

Peter230655:
This may be _mildly_ interesting for examples-gallery:
It compares a simple system with four ODEs and one AE for five state variables vs. the same system except the AE is differentiated w.r.t time.
I would have expected that the system with the AE gives better results, and this is confirmed at least in this example.
(I think, there is a typo in these two problems: It states, that the objective function should be maximized, but I only got close to his results when I minimized it, Update: John Betts confirmed this typo by email)

An aside: 
In Betts' book, the examples are calculated to around 7 significant digits (as stated in appendix A2 of his book) whereas opty normally meets the eoms much closer, here around 11 digits. This may be a reason (other than different machines) that the few examples I have set up always run much slower than the running time Betts gives.


Peter230655:
This PR was replaced by PR #312, as it does not pass the latest test. Therefore I close it.

Peter230655:
This PR was replaced by PR #312, as it cannot pass the latest test. Therfore I close it.

PR #262, opened 2024-11-09
Betts 10 47

Peter230655:
- It is a three stage problem, where the results of the previous stage are the initial conditions for the next one. There is no example like this in examples-gallery, to my knowledge.
- The second stage is a singular arc (see section 4.16.1 for some explanation). I do not think, opty allows to incorporate the additional boundary condition needed, so I just ignorted it. As per John Betts ignoring it may make the result less reliable.
- In the first stage opty does not converge, no matter what I tried. Yet the results are good. So, maybe one should not immediately discard a result of opty simply because opty says it did not converge. **Update 1-5-25:** Now that I include this boundary condition, also the first stage converges rapidly
- The results given in the book differ from the results John Betts sent to me by email, by 0.5..0.6 %. So, it seems to me the difference between opty and Betts' results are acceptable.


Peter230655:
I made some mistake with git!

Peter230655:
Just for information: These are the result John Betts sewnt me by email.
[cdsosn_gdrd07.out.zip](https://github.com/user-attachments/files/17686869/cdsosn_gdrd07.out.zip)
[cdsosn_gdrd07_barrier.zip](https://github.com/user-attachments/files/17686870/cdsosn_gdrd07_barrier.zip)

Peter230655:
When I pushed this initially, I did not know how to handle nontrivial instance constraints, depending on state variables and control variables, so I ignored it.
Now that I know how to do it, I included it.

moorepants:

Peter230655:
No, there is not. Only reason I did not close it already was because you were interested in the name.
(all of Betts' problems has some difficulty to make them converge. But it was shown opty can handle it here)

moorepants:
If you want this one in, then you need to place it in beginner, intermediate, or advanced.

moorepants:
A brief scan shows that there are many formatting and structure things that need addressed. I will not nitpick through them all, as it should be apparent from all prior PRs.

Peter230655:
I thnik, it should be intermediate:
- it shows how to solve a multistage problem, given the constraints we presently have. Once ``opty`` can handle them, we can always remove it. We do not have this in examples-gallery to my knowledge.

Peter230655:
- I improved the docstring as per current standard.
- I used flake8 to look for violations of PEP8. I feel with indentations flake8 does not always agree with you.
- The last phase, where there is really nothing to be optimized took by far the most iterations, about 10,000 vs. about 85 for the allegedly most difficult one, phase2. the singular arc. Maybe because there is nothing to be optimized ``opty`` desperately searches?

Peter230655:
Corrected an error in the identation in doscstrings.

moorepants:
> I used flake8 to look for violations of PEP8. I feel with indentations flake8 does not always agree with you.

PEP8 is a guideline and I generally follow it. I believe the indentation that is generally followed in opty is PEP8 compatible.

examples-gallery/intermediate/plot_betts_10_47.py
```diff
@@ -0,0 +1,415 @@
...
+prob = Problem(obj,
+        obj_grad,
+        eom,
+        state_symbols,
+        num_nodes,
+        interval_value,
+        instance_constraints=instance_constraints,
+        bounds=bounds,
+        time_symbol=t,
+        backend='numpy',
```
moorepants, line 133: Indentation in Python is multiples of 4 spaces, but there is no reason to indent 8 spaces unless you are in a second layer of indentation. This passes flake8 because it only checks if indentation is a multiple of 4, but you are over indenting.

examples-gallery/intermediate/plot_betts_10_47.py
```diff
@@ -0,0 +1,415 @@
...
+
+# %%
+# Plot the objective function as a function of optimizer iteration.
+prob.plot_objective_value()
+error_t = (t_phase1*(num_nodes-1) - 13.726485)/13.726485 * 100
+error_v = (solution[2*num_nodes-1] - 791.2744)/791.2744 * 100
+error_h = (solution[num_nodes-1] - 4560.8912)/4560.8912 * 100
+error_m = (solution[3*num_nodes-1] - 1.323901)/1.323901 * 100
+print((f'duration of phase 1 is {(num_nodes-1) * solution[-1]:.3f}, '
+      f'error is   {error_t:.3f} %'))
```
moorepants, line 164: This is also not correct, the "f"'s should align for correct indentation.

examples-gallery/intermediate/plot_betts_10_47.py
```diff
@@ -0,0 +1,415 @@
...
+# Give some rough estimates for the trajectories.
+# The result of a previous run are used to save time.
+initial_guess = np.array((*[solution[num_nodes-1] for _ in range(num_nodes)],
+                          *[solution[2*num_nodes-1] for _ in range(num_nodes)],
+                          *[solution[3*num_nodes-1] for _ in range(num_nodes)],
+                          *[Tm/c for _ in range(num_nodes)], solution[-1]))
+
+# %%
+# Find the optimal solution.
+for _ in range(1):
```
moorepants, line 241: remove loop

examples-gallery/intermediate/plot_betts_10_47.py
Peter230655: would this be over-indented? 4 spaces is lime pressing the space bar on the keyboard 4 times?

examples-gallery/intermediate/plot_betts_10_47.py
Peter230655: will correct this. My old eys do not catch this sort of thing easily.

examples-gallery/intermediate/plot_betts_10_47.py
Peter230655: Will do. A leftover from when I thought I needed it.

Peter230655:
Now I am at a loss with this error. I thought I had found it.

examples-gallery/intermediate/plot_betts_10_47.py
Peter230655: Removed.

Peter230655:
I still cannot find this mistake.
I only have two bullit lists and cannot find the mistale mentioned. :-(

moorepants:
rst lists have to have blank lines above an below:

```rst
a sentence

- item 1
- item 2
- item 3

a sentence
```

Peter230655:
I checked this, and my two bullit list seem to have it, at least I cannot see where I missed the blank lines.

Peter230655:
Just another attempt, as I cannot see where the problem could be.

Peter230655:
Somewhere in the docstrings I had written  ``.....phase - wise.....`` Let's see if this was the issue.

Peter230655:
I have no more ideas what to look for.
The error refers to a file:
/home/runner/work/opty/opty/docs/examples/intermediate/plot_betts_10_47.rst
but I have no idea where tro find it.
@moorepants : whenever you have time could you look at the docstrings, I simply have no more ideas. Thanks!

PR #266, opened 2024-11-16
Betts 10 141

Peter230655:
The only reason I think, this might be a candidate for examples-gallery is this:

It is a variable time problem, the final value of a state variable is to be minimized.
It is only 3 'simple' DEs.
BUT: unless I give the final time prettly close to the result given in the book (final time is 1.196.., so unless I give
$h = \dfrac{1.2 ... 1.8}{\text{num nodes - 1}}$ 
it either does not converge at all, or to some totally different result. 
If I give the correct solution as initial_guess, it is also more tolerant about the final time.

I do not think, I have ever come across such a 'sensitive' set of EOMs


Peter230655:
As this certainly does not meet the guidelines of examples-gallery, I close it.

PR #267, opened 2024-11-19
Betts 10 144 145

Peter230655:
These are two examples solving the same problem with a bit different (but equivalent) sets of equations of motion.
This might be a candidate for examples-gallery for two reasons:
- The second version (10.145) adds one algebraic equation to the eoms, and add one control variable. As opty needs as many state variables as there are eoms, I simply declared one control variable to be a state variable. Admittedly, not a stroke of genius, but I do not think, this is in examples-gallery, yet.
- the second version, though mathematically equivalent to the first one seems to hold the constraints better by a factor on 10...100. I do not know why this is so. I any case it shows, that different, but equivalent formulations may affect the accuracy.


moorepants:
This needs to be put in beginner, intermediate, or advanced.

Peter230655:
> This needs to be put in beginner, intermediate, or advanced.

I think, beginner. In a way, it simply shows that one can convert a unknown trajectory into a state variable if necessary.
But frankly, I am unsure whether it even should go into examples-gallery.
(This morning I closed 2 or 3 where I was sure they have no business being there)

moorepants:
You decide. I will review if you think it fits into opty's gallery.

Peter230655:
> You decide. I will review if you think it fits into opty's gallery.

Then let's close it. Examples-gallery should only have first class simulations, and this is not one of them

moorepants:
Anything that offers a unique thing to learn you can do with opty, is suitable, but they should be simple examples. Does this comparison of two solutions here show anything we haven't already?

Peter230655:
> Anything that offers a unique thing to learn you can do with opty, is suitable, but they should be simple examples. Does this comparison of two solutions here show anything we haven't already?

The comparison (AE vs. AE.diff(t) ) is not in any other example to the best of my knowledge. So, this would be unique.

moorepants:
This example doesn't show that?

https://opty.readthedocs.io/latest/examples/beginner/plot_betts_10_103_104.html#sphx-glr-examples-beginner-plot-betts-10-103-104-py

Peter230655:
> This example doesn't show that?
> 
> https://opty.readthedocs.io/latest/examples/beginner/plot_betts_10_103_104.html#sphx-glr-examples-beginner-plot-betts-10-103-104-py

Minimum, they are VERY similar:
10-103-104 directly converts AE into AE.diff(t)

10-144-145 has a tern: $\dot y_2 = (1 - y_1^2)y_2 - y_1 + u$  and converts this into
```
$\dot y_2 = v - y_1 + 1$
$v = (1 - y_1^2)y_2$
´´´

Peter230655:
moved to pst-notebooks

PR #269, opened 2024-11-24
Betts 10 1

Peter230655:
This example may be a candidate for examples gallery for this reason:

In addition to the four equations of motion for the state variables $y_1,...y_4$ there is an **inequality** these variables must fulfill:
$y_1^2 + y_2^2 + y_3^2 + y_4^2 \geq$ function(time, parameters).
As presently opty does not allow inequalities, I introduced an additional state variable _fact_ and made
$y_1^2 + y_2^2 + y_3^2 + y_4^2$ = fact $\cdot$ function(time, parameters) the fifth eom. I restrict fact $\geq 1.0$.
This should be equivalent to the inequality, as long as **function(time, parameters $\geq 0$** throughout the interval considered, as is the case here.

The minimum I get with opty is actually about 5% lower than the value given in the book, the solutions look similar to the ones given in the book.


Peter230655:
I made a new PR #308 as I could not fix the branch. Hence I close it.

PR #271, opened 2024-11-26
Examples 10.73 and 10.74 from Betts' book

Peter230655:
These two examples solve the same problem, but formulated differently.
- 10.73 has a boundary condition at the final time: *function(state variables, parameters) = 0*. As opty currently does not allow such instance constraints, I introduced an additional state variable, _hilfs_, and an additional equation of motion as _hilfs - function(state variables, parameters)_. Then I set an instance_constraint as _hilfs($t_f$) = 0_.
I believe there is no such example in examples-gallery.
- while 10.74 converged very rapidly, 10.73 only converged with very close initial guesses. This shows, that the formulation of a problem has a (big) influence on the ease of solving it.

### DIFF
--- a/examples-gallery/plot_betts_10_73_74.py
+++ b/examples-gallery/plot_betts_10_73_74.py
@@ -1,0 +1,280 @@
+"""
+Linear Tangent Steering
+=======================
+
+These are example 10.73 and 10.74 from Betts' book "Practical Methods for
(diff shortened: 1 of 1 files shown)


Peter230655:
There is nothing in this simulation, which is not in other examples.
I close this PR and move the simulation to pst-notebooks

PR #272, opened 2024-11-28
Betts 10 50

Peter230655:
- Let $t_0, t_f$ be the starting and the final times of the simulation. There are instance constraints $x_i(t_0) = x_j(t_f)$ which to my knowledge *opty* cannot handle at the moment. So I iterate over the problem and reset the instance constraints at each iteration. At least in this example (and another one I did) this converged to fulfill these instance constraints.
- there are inequalities in the equations of motion. So, I define additional state variables to handle this. This is not new in the sense that I did this in PR # 269

The minimum given in Betts' book is missed by over 12%. So, I guess this "iteration - solution" is only a 'crutch' until *opty* can handle this type of situation.


Peter230655:
Instance constrains like $x_2(t_f) - x_1(t_0)$ are possible with opty, so I changed the simulation accordingly.

Peter230655:
Mostly cosmetic changes. With the instance constraints like $x_i(t_f) - x_j(t_0)$ which opty allows, it converges rapidly amd gets very close to Betts' result.
This is a simple example of a time delay differential system, described in Betts' book page 269 ff.

moorepants:
Yes, we need this example to show the more complex instance constraints.

Peter230655:
> Yes, we need this example to show the more complex instance constraints.

Thanks! :-)

PR #273, opened 2024-12-01
Detailed eoms

Peter230655:
At present all EoM violations a printed in one plot when using _plot_constraint_violations(..)_. There may be occasions when one would want to see each eom violation in a separate plot.
I added an optional keyword **detailed_eoms** with default value False. If set to True, each eom violation will be put in a separate  plot, If not set, or set to False, all is like it has been, that is all EoM violations in one plot.


Peter230655:
This was replaced by PR #310, as this one cannot pass the doc/readthedoc.... test and I did not know how to fix it.

PR #276, opened 2024-12-10
Car in garage

Peter230655:
A car should park in a garage without hitting any walls.
To avoid hitting any walls, I form a (differentiable) trough in the shape of the garage and the walls. *number* of points along the body of the car must always be above the trough. This leads to *number* of **inequalities** which I handle by introducing new state variables.
I have handled inequalities in 1 or 2 previous PRs. Maybe this one is *nicer* as it is applied to 'real' eoms, instead of just using the equations given in Betts' book.

As *a priori* it is not clear whether the care will back in or drive in straight, I look at the 'lower' part of the car, either fron or back for the final location, again a state variable.

NB: depending on the starting location of the car, it can take a very long time for the program to converge.


Peter230655:
**NO** rush! Github does not loose anything! :-)
NB: Instance constraints like $x_2(t_0) = x_1(t_0)$ seem to work. I modiefied betts_10_50 accordingly, and tried one more, which I will not push, as similar only larger.
Bounds like {x: (y($t_0, y(t_f)$} do not seem to work. If tzhey worked this would bre 'close' to an inequality. I think.

Peter230655:
Better starting position of the car, showing more clearly that it avoids the walls to get into the garage.
Better sphinx thum nail picture

Peter230655:
I pushed this from the worng branch. Now I pushed it correctly (#289), so I close this PR.

PR #277, opened 2024-12-12
Betts 10 113

Peter230655:
This is a **very** simple example how to handle inequalities of the form: _function(state_variables, control_variables, parameters) >= 0_
I simply define a new state variable, say, J, and set J = function(state_variables, control_variables, parameters) as additional equation in the equations of motion.
Then, in the bounds I set: { J: (-np.inf, 0.0) }

I made similar PRs before, but none this simple.
I guess, a matter of taste, which, if any at all, should make it into examples-gallery.


Peter230655:
I pushed this from a wrong branch.
I pushed it agaon, s I close this

PR #278, opened 2024-12-14
Car on racecourse

Peter230655:
A car must move from start to finish on a windy road as fast as possible.

The road is modelled as two functions with $f(x) > g(x) \forall{a}$.
I take *number* of points $P(x_i | y_i)$ evenly spaced on the body of the car and ensure that $f(x_i) > y_i > g(x_i)$
This is done by using additional state variables $P_{upper_i}, P_{lower_i}$, set, say, $P_{upper_i} = f(x_i) - y_i$ and bound $P_{upper_i} > 0.0$
This exactly like my car parking in a garage, just a bit more elaborate, so I think at most one of them should be merged.

I also limit the maximum accelerations for the front and the end of the car, to mimick slowing down in curves.

NB: opty always arranged the car to "drive backwards" ,like a fork lift. Apparently better to solve this problem.


Peter230655:
I must have made some mistake pushing this

Peter230655:
New attempt.

Peter230655:
I psuhed this again (#290) from the correct branch, so I close this one.

Issue #255, opened 2024-10-30
opty does not accept only one scalar differential equation


Peter230655:
Problem 10.7 in Chapter 10 of John T. Betts' _Practical Methods for..._.., 3rd edition, only has one scalar differential equation:
$\dot y = - y^3 + u$, where u is the control.
When I run it, I get this error:
![image](https://github.com/user-attachments/assets/3dc12339-c50c-4fc0-81cf-f16ae64d8cfc)

When I 'artificially' enlarge the system:
$uy = \dfrac{d}{dt} y$
$uy = - y^3 + u$
all works fine and it find the solution.
This is definitely not a big thing, it is probably a rare case that there is only one scalar DE, but I wanted to point it out.

moorepants:
Yes, opty assumes you are working with a physics system that has a mass and forces, thus you always get at least one second order differential equation.

Peter230655:
> Yes, opty assumes you are working with a physics system that has a mass and forces, thus you always get at least one second order differential equation.

Makes sense.
Betts gives the dynamic equations as a set of first order DEs, at least in the three examples I have calculated.
In example 10.6 it is a set of 10 first order DEs, which do not look like mechanical equations - yet the results look fine.
So, as long as it is more than one first order DEs in the eoms, opty does not care how they 'originated?

moorepants:
>  So, as long as it is more than one first order DEs in the eoms, opty does not care how they 'originated?

I think that the only assumption is that there are n states and n equations, where n is > 1.

Two things:

- An example in the gallery that shows how to use a single differential equation with your hack is useful as a temporary solution.
- We should relax the assumption opty makes, because we also would like to add extra constraint equations where n states != num constraint equations (e.g. in https://github.com/csu-hmc/opty/issues/156).

Peter230655:
> This issues is sufficient for the request for support for single first order differential equations.

Not sure I understand. Raise a new issue or not?
(Sorry, my command of the English language...)

moorepants:
This is the issue, no new issue is needed.

moorepants:
You could try removing https://github.com/csu-hmc/opty/blob/master/opty/direct_collocation.py#L1184 and just see what happens.

Peter230655:
> You could try removing https://github.com/csu-hmc/opty/blob/master/opty/direct_collocation.py#L1184 and just see what happens.

I will try in the next couple of days and report.

Peter230655:
```
# %%
"""
[...]

PR #279, opened 2024-12-16
One eom enough

Peter230655:
This should solve issue #255.
I simply removed two lines each, which tested that there are at least two eoms. Around line 1200 and around line 1600 in the code.
I also added a test function with only one eom.


Peter230655:
test functions seem to have some "assert".
In my case I do not know, what I should assert. It has one eom only.
Also unsure, whether at the correct location. I put in in the folder tests.

Peter230655:
I corrected my test function as suggested. I removed it from the wrong location and added it to test_direct_collocation.py. 
The typo in the documentation was confirmed by @moorepants, so I used this.

Peter230655:
Here again, I pushed from the wrong branch.
I pushed it correctly(#287), so I close this PR.

Issue #282, opened 2024-12-20
Wrong time step width in plots generated by Problem.plot_trajectories()


chris-konrad:
[The time array for trajectory plotting](https://github.com/csu-hmc/opty/blob/aae009e48f1c1e9d19f43b12c650929622b0a510/opty/direct_collocation.py#L399) created by Problem.plot_trajectories() covers the interval [0, `num_collocation_nodes * node_time_interval`] including the limits while forcing the array size to be `num_collocation_nodes`. As a result, the time spacing is not as expected. This is the default behaviour of `np.linspace()`. Should be fixed by using `np.arange()` instead.

Example for `num_collocation_nodes = 5`, `node_time_interval=1`:
```
In[2]: np.linspace(0, num_collocation_nodes * node_time_interval, num=num_collocation_nodes)
Out[2]: array([0.  , 1.25, 2.5 , 3.75, 5.  ])

In[3]: np.arange(0, num_collocation_nodes * node_time_interval, node_time_interval)
Out[3]: array([0, 1, 2, 3, 4])
```

In the linspace version, the time steps of the resulting array are not 1, which however would be expected because of `node_time_interval=1`.

Peter230655:
As I know it, the final time is defined as tf = node_time_interval * (num_collection_nodes **- 1**),
you defined it as tf = node_time_interval * num_collection_nodes

chris-konrad:
That would fix the problem as well. But it is currently not implemented that way (see: [/opty/direct_collocation.py#L400](https://github.com/csu-hmc/opty/blob/aae009e48f1c1e9d19f43b12c650929622b0a510/opty/direct_collocation.py#L400))

Peter230655:
Maybe you are rigt, but all the examples I ever saw, the final time was defined a interval * (num-nodes **- 1**), assuming the starting time is zero.

chris-konrad:
Have a look at [np.arange(start, stop, step)](https://numpy.org/doc/2.2/reference/generated/numpy.arange.html). It defines the interval as [start,stop). In `np.arange(0, num_collocation_nodes * node_time_interval)`,  the end of the interval is `stop = num_collocation_nodes * node_time_interval = 5`. But this will not be included in the time array. It stops at 4: 
```
In[3]: np.arange(0, num_collocation_nodes * node_time_interval, node_time_interval)
Out[3]: array([0, 1, 2, 3, 4])
```

PR #283, opened 2024-12-20
Correcting time step width of trajectory plots to equal node_time_int…

chris-konrad:
Fixing #282

### DIFF
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -396,10 +396,10 @@ class Problem(cyipopt.Problem):
                 variable_duration=self.collocator._variable_duration)
             node_time_interval = self.collocator.node_time_interval
 
-        time = np.linspace(0,
-                           self.collocator.num_collocation_nodes *
-                           node_time_interval,
-                           num=self.collocator.num_collocation_nodes)
+        time = np.arange(0,
+                         self.collocator.num_collocation_nodes *
+                         node_time_interval,
+                         node_time_interval)
 
         num_axes = (self.collocator.num_states +
                     self.collocator.num_input_trajectories)
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -396,10 +396,10 @@ class Problem(cyipopt.Problem):
                 variable_duration=self.collocator._variable_duration)
             node_time_interval = self.collocator.node_time_interval
 
-        time = np.arange(0,
-                         self.collocator.num_collocation_nodes *
-                         node_time_interval,
-                         node_time_interval)
+        time = np.linspace(0,
+                           (self.collocator.num_collocation_nodes-1) *
+                           node_time_interval,
+                           num=self.collocator.num_collocation_nodes)
 
         num_axes = (self.collocator.num_states +
                     self.collocator.num_input_trajectories)


chris-konrad:
I don't think this can happen. The results of arrange and linspace with (num_collocation_nodes-1) are equivalent:

```
In[30]: np.arange(0, num_collocation_nodes * node_time_interval, node_time_interval)
Out[30]: array([0., 1., 2., 3., 4.])

In[31]: np.linspace(0, (num_collocation_nodes-1) * node_time_interval, num=num_collocation_nodes)
Out[31]: array([0., 1., 2., 3., 4.])
```

Peter230655:
> I don't think this can happen. The results of arrange and linspace with (num_collocation_nodes-1) are equivalent:
> 
> ```
> In[30]: np.arange(0, num_collocation_nodes * node_time_interval, node_time_interval)
> Out[30]: array([0., 1., 2., 3., 4.])
> 
> In[31]: np.linspace(0, (num_collocation_nodes-1) * node_time_interval, num=num_collocation_nodes)
> Out[31]: array([0., 1., 2., 3., 4.])
> ```
> 
> It will not really matter which correction we choose. But the current behaviour is definitely not correct :)

...but if an existing simulation uses tf = interval * (num_nodes - 1), then your change will matter?

chris-konrad:
I am not sure I understand you correctly. Where would a current simulation use `tf = interval * (num_nodes - 1)`?

Peter230655:
Virtually all the simulations I know or wrote define the starting time t0, mostly t0 = 0 and the final time tf.
Then interval = (tf - t0) / (num_nodes - 1) [you may look at the simulations in examples-gallery]
I guess the advantage is, that you can play around with num_nodes for convergence without messing up the 'physics' of the problem.

As regards compatibility, I do not know for sure.

Peter230655:
I have no preference, it is up to Jason to decide about this.

chris-konrad:
Actually, @Peter230655, turns out there is a good reason to prefer linspace over arange. The latter is [less numerically stable](https://numpy.org/doc/stable/reference/generated/numpy.arange.html) when using non-integer steps. This actually caused two of the examples to fail after all.

So, we should keep using linspace and add the correction by -1 instead. 
```
        time = np.linspace(0,
                           (self.collocator.num_collocation_nodes-1) *
                           node_time_interval,
                           num=self.collocator.num_collocation_nodes)
```

chris-konrad:
But yes indeed. The "error" is purely cosmetic and won't really be visible for large sample numbers.

Peter230655:
> Actually, @Peter230655, turns out there is a good reason to prefer linspace over arange. The latter is [less numerically stable](https://numpy.org/doc/stable/reference/generated/numpy.arange.html) when using non-integer steps. This actually caused two of the examples to fail after all.
> 
> So, we should keep using linspace and add the correction by -1 instead.
> 
> ```
>         time = np.linspace(0,
>                            (self.collocator.num_collocation_nodes-1) *
>                            node_time_interval,
>                            num=self.collocator.num_collocation_nodes)
> ```

I had no idea about the instabilitiy of arange. I am not the biggest Python expert, so I always use linspace.
Looks to me your suggestion makes sense, based on the definition of tf = interval * (num_nodes - 1) as it is done in all examples I know.

moorepants:
LGTM!

Thanks for your first contribution :)

PR #284, opened 2024-12-23
Drone in tube

Peter230655:
This is the drone from examples-gallery, but now it must stay inside a tube of a given radius, and a curve as centerline of the tube.
To get the distance of the drone to the centerline of the tube, I intersect a certain plane (described in more detail in the program) with the centerline. 
This will lead to a nonlinear equation for the point of intersection. I declare this equation an additional eom, and the parameter as an additional state variable.  I do not think, this method in in examples gallery already.
The distance drone-centerline also becomes a state vatiable, so it can be bound 0 <= distance <= radius of the tube.

It turned out, that it is very difficult to get it to converge. My suspicion is that because the nonlinear equation in general has many solutions, this causes trouble with convergence. For example I never managed to get the drone to fly as fast as possible.


Peter230655:
I do not see, how the error is related to my code: it seems to pass the test with older an with newer versions of python - if I read it correctly.
image
Unlcear, what this means.

Peter230655:
Small obervation: when I switched the description of the rotation of the drone from quaternions to 'normal' angles, this seemed to help 'speed optimization'. But it more seemed to hurt 'minimum energy optimozation'. All in all it seems to be a difficult problem for opty, unless I committed some basic blunder.

Peter230655:
I did not change anything in the code. Just trying if the error given earlier has gone.

Peter230655:
I pushed this one from the wrong branch.
I pushed it again, so I will close this PR.

PR #285, opened 2024-12-29
Particle in tube

Peter230655:
A particle in §D must move from A to B and must stay in a tube definded by a parametric curve and a radius.
(Essentially like the drone, but I switched to a particle to get simple (=faster) EOMs)
Here one can see the following:
- trying to get the distance from the particle to the curve results in a nonlinear equation for the parameter which gives the point of the curve with smallest distance. I make this an additional rom, and the parameter n additional state variable.
- to ensure the particle stays inside the tube, I add another eom: dist = distance, and bound dist <= radius of tube.
- at some intermediate time tm the particle should move through a gate, modelled as a circle with center on the curve. To do this, I introduce yet another state variable and another specified variable, gate_h, and add gate = gate_h * distance. I bound gate_h >= 1 and set the instance constraint gate(tm) = radius of the gate.

Most of this is in other PRs, but I find maybe nicely summarized here.


Peter230655:
I did the branching all wrong here, this is why so many files are changed.
I pushed the same example again (#286), no with correct branchin as @tjstienstra explained to me.
So, I close this PR.

PR #286, opened 2024-12-30
particle-in-tube-new

Peter230655:
I hope now I did the branching correctly.

A particle in 3D must move from A to B and must stay in a tube definded by a parametric curve and a radius.
(Essentially like the drone, but I switched to a particle to get simple (=faster) EOMs)

Here one can see the following:
Trying to get the distance from the particle to the curve results in a nonlinear equation for the parameter which gives the point of the curve with smallest distance. I make this an additional rom, and the parameter n additional state variable. to ensure the particle stays inside the tube, I add another eom: dist = distance, and bound dist <= radius of tube.at some intermediate time tm the particle should move through a gate, modelled as a circle with center on the curve. To do this, I introduce yet another state variable and another specified variable, gate_h, and add gate = gate_h * distance. I bound gate_h >= 1 and set the instance constraint gate(tm) = radius of the gate.

Most of this is in other PRs, but I find maybe nicely summarized here.


Peter230655:
Now it traces the particle in the tube. I think, this shows nicely, how the particle does go through the gate, at the place of the gate closest to the wall of the tube - which makes sense as the energy should be minimized.

Peter230655:
worked again! :-)

examples-gallery/plot_particle_in_tube.py
```diff
@@ -0,0 +1,513 @@
...
+- :math:`radius` : radius of the tube [m]
+- :math:`max_z` : maximum height of the particle [m]
+
+
+**States**
+
+- :math:`x, y, z` : position of the particle [m]
+- :math:`v_1, v_2, v_3` : speed of particle [m]
+- :math:`dist` : distance of the particle from the curve [m]
+- :math:`cut_{param}` : parameter of the curve where the distance is closest.
```
moorepants, line 51:
```suggestion
- :math:`\textrm{cut}_\textrm{param}` : parameter of the curve where the distance is closest.
```

Words in latex math should have `\textrm{}` for correct rendering. There are many instances here where it is needed.

examples-gallery/plot_particle_in_tube.py
```diff
@@ -0,0 +1,513 @@
+"""
```
moorepants, line 1:
```suggestion
r"""
```

If you have latex in a python string that includes backslashes, you should make the string a raw string so you don't have to escape the backslashes.

examples-gallery/plot_particle_in_tube.py
```diff
@@ -0,0 +1,513 @@
...
+The particle must not leave a tube defined by a curve (centerline) in space and a
+radius. At one point during the motion, it must pass through a narrow gate,
+modelled as a circle.
+
+Interesting maybe this:
+( In what follows all components are w.r.t. the inertial frame N.)
+The curve is given as :math:`X(r) = (f(r, params), g(r, params), h(r, params))`,
+where :math:`r` is the parameter of the curve. Let :math:`cut_{param}` be the parameter
+of the curve where the distance of the particle from the curve is closest. Then
+:math:`( \\dfrac{df}{dr}, \\dfrac{dg}{dr}, \\dfrac{dh}{dr} ) |_{r = cut_{param}}`
```
moorepants, line 16: When the raw string is add all backslash escapes should be removed. `\\` -> `\`

examples-gallery/plot_particle_in_tube.py
```diff
@@ -0,0 +1,513 @@
...
+- :math:`max_z` : maximum height of the particle [m]
+
+
+**States**
+
+- :math:`x, y, z` : position of the particle [m]
+- :math:`v_1, v_2, v_3` : speed of particle [m]
+- :math:`dist` : distance of the particle from the curve [m]
+- :math:`cut_{param}` : parameter of the curve where the distance is closest.
+- :math:`cut_{dt} = \\dfrac{d}{dt} cut_{param}` : :math:`cut_{dt} \\geq 0`, so the particle never flies backwards.
```
moorepants, line 52: wrap lines to 79 characters

Peter230655:
This is one of my favorite simulations.
As I recall, it was not easy to find initial conditions which worked. I had to add final conditions one by one, and use the solutions obtained as initial guess for the solution with an additional final constraint.

Peter230655:
I am unclear about this error:
image
1.
I did not touch any code, and before it passed all tests
2.
The objective function is an indefitie integral

moorepants:
I've already discussed that error in other PRs. I have not been able to debug yet.

Peter230655:
Now I got that error you addressed before:
image
I will just push the same in a little while.

Peter230655:
I do not understand how this new error relates to my simulation.
The previous error did not show up.
My only change was in the docstrings, where I moved a word, which slightly extended the max. allowable with, to a new line.

Peter230655:
Anyway, now it seems it has passed alltests. :-)

Peter230655:
**Thanks!!**
1. I see that ``textrm`` can be used much more widely than I thought. I did not know it until a few days ago.
2.  You change may indentations of formulae of of function wrapping around into the next line. What is the correct amount of indentation?
3. In line 312 / 313 you gave different indentations.  Any reason for this?

moorepants:
Yes, if you want a word in a math formula, you generally need to use textrm so it isn't the same font as a math variable.

I try to follow the guidelines in pep8: https://peps.python.org/pep-0008/ for indentation, etc. But I have a tool in my IDE that tells me if i'm not following pep8. The easiest thing is to enable a pep8 checker for your IDE.

examples-gallery/plot_particle_in_tube.py
```diff
@@ -0,0 +1,533 @@
...
+
+prob.add_option('max_iter', 3000)
+
+# %%
+# Give a guess of a plausible route with constant thrust. Here the
+# solution of a previous run is used as initial guess.
+initial_guess = np.zeros(prob.num_free)
+x_guess, y_guess, z_guess = eval_curve(np.linspace(0.0, max_z/par_map[a3],
+                                                   num=num_nodes), par_map[a1],
+                                       par_map[a2], par_map[a3])
```
moorepants, line 313: The difference indentation matches with the leading parentheses for the function the arguments are related to.

PR #289, opened 2024-12-30
park a car in a garage

Peter230655:
I did this from the wrong branch earlier. Now corrected:

A car should park in a garage without hitting any walls.
To avoid hitting any walls, I form a (differentiable) trough in the shape of the garage and the walls. number of points along the body of the car must always be above the trough. This leads to number of inequalities which I handle by introducing new state variables.
I have handled inequalities in 1 or 2 previous PRs. Maybe this one is nicer as it is applied to 'real' eoms, instead of just using the equations given in Betts' book.

As a priori it is not clear whether the care will back in or drive in straight, I look at the 'lower' part of the car, either fron or back for the final location, again a state variable.

NB: depending on the starting location of the car, it can take a very long time for the program to converge.


Peter230655:
There is **nothing** new here compared to #285 here.
Since @moorepants seemed to like it some time ago, I will not close it but let him decide.

Peter230655:
Worked again!! :-)   :-)

examples-gallery/plot_car_in_garage.py
```diff
@@ -0,0 +1,474 @@
...
+# %%
+# Plot the constraint violations.
+prob.plot_constraint_violations(solution)
+
+# %% [markdown]
+# Plot generalized coordinates / speeds and forces / torques
+prob.plot_trajectories(solution)
+
+# %%
+# Aminate the Car
```
moorepants, line 350:
```suggestion
# Animate the Car
```

examples-gallery/plot_car_in_garage.py
```diff
@@ -0,0 +1,474 @@
...
+
+ax[4].plot(XX, step_r_diff_lam(XX, b, gr))
+ax[4].axvline(b, color='k', linestyle='--')
+ax[4].set_title('differentiable step_r')
+
+ax[5].plot(XX, in_0_1_lam(XX))
+ax[5].axvline(0, color='k', linestyle='--')
+ax[5].axvline(1, color='k', linestyle='--')
+ax[5].set_title('differentiable in_0_1')
+prevent_print = 1.
```
moorepants, line 231: If you end the set_title call with a `;` it shouldn't print the output.

moorepants:
Very nice!

examples-gallery/plot_car_in_garage.py
Peter230655: > If you end the set_title call with a `;` it shouldn't print the output.

I forgot -:(

Peter230655:
> Very nice!

Thanks! :-)

PR #290, opened 2024-12-30
car on a racecourse, limited acceleration

Peter230655:
I had pushed this earlier, from wrong branch.

A car must move from start to finish on a windy road as fast as possible.

The road is modelled as two functions with f ( x ) > g ( x ) ∀ x .
I take number of points P ( x i | y i ) evenly spaced on the body of the car and ensure that $f ( x_i ) > y_i > g ( x_i )$
This is done by using additional state variables $Pupper_ i , Plower_i$ , set, say, $Pupper_i = f ( x_i ) − y_i$ and bound $Pupper_i > 0.0$
This exactly like my car parking in a garage, just a bit more elaborate, so I think at most one of them should be merged.

I also limit the maximum accelerations for the front and the end of the car, to mimick slowing down in curves.

NB: opty always arranged the car to "drive backwards" ,like a fork lift. Apparently better to solve this problem.


Peter230655:
Nothing new here compared to #285 , so I close it.
As I thing, it is quite nice, I moved it to pst-notebooks.

PR #291, opened 2024-12-30
betts 10 113, new

Peter230655:
I pushed this from a wrong branch earlier.

This is a very simple example how to handle inequalities of the form: function(state_variables, control_variables, parameters) >= 0
I simply define a new state variable, say, J, and set J = function(state_variables, control_variables, parameters) as additional equation in the equations of motion.
Then, in the bounds I set: { J: (-infinity , 0.0) }

I made similar PRs before, but none this simple.
I guess, a matter of taste, which, if any at all, should make it into examples-gallery.

### DIFF
--- a/examples-gallery/plot_betts_10_113.py
+++ b/examples-gallery/plot_betts_10_113.py
@@ -1,0 +1,141 @@
+"""
+Mixed State-Control Constraints
+===============================
+
+This is example 10.113 from Betts' book "Practical Methods for Optimal Control
+Using NonlinearProgramming", 3rd edition, Chapter 10: Test Problems.
+
+This simulation shows how to handle inequality constraints, here of the form:
+:math:`0 \\geq u + \\dfrac{y_1}{6}`.
+
+I  set:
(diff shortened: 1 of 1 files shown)


moorepants:
You successfully merged the upstream master and pushed this. And the build is complete. Nice job!

Peter230655:
> You successfully merged the upstream master and pushed this. And the build is complete. Nice job!

I had a LOT of help!
Question: Do you know a person dumber than me in git / github?
Answer;   NO!! :-)

1. In one or two PRs I proposed changes to the class Problem. Would this work the same way, or would I have to manually correct?
2.  Should I push all PRs for examples-gallery this way, now that I know how to do it?

Peter230655:
I managed to successfully push two simulations, but now I am stuck again!

image

Peter230655:
This 'hack' how to handle inequality constraints is shown in various simulations in examples-gallery, so there is no need for one more. I will move it to pst-notebooks and clse this PR

PR #292, opened 2024-12-30
drone in tube, new

Peter230655:
I pushed this earlier, from a wrong branch.

This is the drone from examples-gallery, but now it must stay inside a tube of a given radius, and a curve as centerline of the tube.
To get the distance of the drone to the centerline of the tube, I intersect a certain plane (described in more detail in the program) with the centerline.
This will lead to a nonlinear equation for the point of intersection. I declare this equation an additional eom, and the parameter as an additional state variable. I do not think, this method in in examples gallery already.
The distance drone-centerline also becomes a state vatiable, so it can be bound 0 <= distance <= radius of the tube.

It turned out, that it is very difficult to get it to converge. My suspicion is that because the nonlinear equation in general has many solutions, this causes trouble with convergence. For example I never managed to get the drone to fly as fast as possible.


Peter230655:
This PR is very similar to #286 but #286 shows one additional feature: The particle has to pass a gate on its way.
So, I close it.

Issue #224, opened 2024-09-09
Feature request: Time delay parameters


chrismo-schmidt:
For our work on human control, we would be interested in introducing EOMs with an unknown time shift parameter $t_d$ that should be a free parameter for the optimization. 

For example an unknown input delay, aka "reaction time":
$\dot x(t) = f(x(t)) + g(u(t-t_d))$ 

If one introduces this in the symbolic expressions of the EOM, the problem does not compile. It seems like this feature is currently not supported:
`eom = x(t).diff - (f(t) +  g(t-t_d)) `

I am trying workarounds with explicit time variables or an allpass. But generally it would be great if opty could support time delay out of the box. 

Any other ideas for workarounds are of course also greatly appreciated. 

Regards,
Christoph

tjstienstra:
Could you perhaps write up code of a simple problem you would like to see working? That would make it easier to provide suggestions for workarounds and for implementations.

I expect that it would be something like this:
```python
import sympy as sm
import sympy.physics.mechanics as me
from opty import Problem, create_objective_function

N = 101
dt= 0.1

t = me.dynamicsymbols._t
td = sm.Symbol("td")
x, y, u = me.dynamicsymbols("x y u")
u_delayed = u.subs({t: t - td})
eoms = sm.Matrix([
  x.diff() - u * y + u_delayed,
  y - sm.sin(x),
])

obj, obj_grad = create_objective_function(u**2, (x, y), (u,), (), N, dt)
prob = Problem(obj, obj_grad, eoms, (x, y), N, dt, instance_constraints=[x.subs({t: 0.0}) - 3])
prob.solve()
```
Note, it is important to choose a good representative equation of motion, e.g. `x.diff() - u_delayed` could just be solved by offsetting the solution of `x.diff() - u`.

moorepants:
opty assumes all time varying symbols are given as `x(t)` and then it discretizes that to `x(t_i)`. If you provide `x(t - 5.0)` as a symbol in an expression, that continuous expression needs to be replace internally with something like `x(t_i - 5.0//h)` where `h` is the constant time step and that would then map to `x(t_{i-j}` where `j=5.0/h`. Once it made it to that form, the derivatives wrt these shifted variables could then be taken as it is already done internally and things would generally work. We'd have to deal with not having values in the past before the initial time in some way.

moorepants:
[...]

PR #294, opened 2025-01-03
Enabeling opty to estimate input time shifts. 

chris-konrad:
This is work in progress addressing https://github.com/csu-hmc/opty/issues/224. The code currently does not run without exceptions and added functions are untested. 

See https://github.com/chris-konrad/opty/issues/1 for status and plans.

### DIFF
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -727,6 +727,7 @@ class ConstraintCollocator(object):
 
         self._sort_parameters()
         self._check_known_trajectories()
+        self._substitute_timeshift_trajectories()
         self._sort_trajectories()
         self.num_free = ((self.num_states +
                           self.num_unknown_input_trajectories) *
@@ -843,6 +844,32 @@ class ConstraintCollocator(object):
             if len(v) != N:
                 msg = 'The known parameter {} is not length {}.'
                 raise ValueError(msg.format(k, N))
+                
+                
+    def _substitute_timeshift_trajectories(self):
+        """Identify and substitute time-shifted trajectories in the eom. Creates a dictionary 
+        linking the substitutes with the original trajectories and the timeshift parameters."""
+        
+        t_set = {self.time_symbol}
+        
+        trajs = self.eom.atoms(sm.core.backend.AppliedUndef, sm.core.backend.Derivative)
+        self.timeshift_traj_substitutes = {}
+        
+        for traj in trajs:
+            if not len(traj.free_symbols.intersection(t_set)) == 1:
+                continue
+            if not len(traj.free_symbols) == 2:
+                continue
+            
+            traj_subs = sm.symbols(traj.name+"_shift", cls=sm.Function)(self.time_symbol)
+            timeshift_parameter = list(traj.free_symbols - t_set)[0]
+            
+            self.timeshift_traj_substitutes[traj_subs] = (traj, timeshift_parameter)
+            
+            self.eom = self.eom.subs(traj, traj_subs)
+            
+        self.num_timeshift_traj_substitutes = len(self.timeshift_traj_substitutes)
+        
 
     def _sort_trajectories(self):
         """Finds and counts all of the non-state, time varying parameters in
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -1096,18 +1096,16 @@ class ConstraintCollocator(object):
         subbed_constraints = self.instance_constraints
         
         # make def map
-        def_map = {}
+        self.instance_constraint_free_map = {}
         for k,v in self.instance_constraints_free_index_map.items():
             if isinstance(v, sm.core.operations.AssocOp):
                 v = v.subs(timeshift_param_map)
-            def_map[k] = free[v,0]             
+            self.instance_constraint_free_map[k] = free[v,0]             
             
         subbed_constraints = [None]*self.num_instance_constraints
         for i in range(self.num_instance_constraints):
-            subbed_constraints[i] = self.instance_constraints[i]
-            funcs = list(self.instance_constraints[i].atoms(sm.Function))
-            for func in funcs:
-                subbed_constraints[i] = subbed_constraints[i].subs(func, def_map[func])
+            subbed_constraints[i] = me.msubs(self.instance_constraints[i], 
+                                            self.instance_constraints_free_index_map)
 
         f = sm.lambdify(([free] + list(self.known_parameter_map.keys())),
                         subbed_constraints, modules=[{'ImmutableMatrix':
@@ -1138,10 +1136,7 @@ class ConstraintCollocator(object):
     def _instance_constraints_jacobian_values_func(self):
         """Returns the non-zero values of the constraint Jacobian associated
         with the instance constraints."""
-        free = sm.DeferredVector('FREE')
-
-        def_map = {k: free[v] for k, v in
-                   self.instance_constraints_free_index_map.items()}
+        free = sm.MatrixSymbol('FREE', self.num_free, 1)
 
         funcs = []
         num_vals_per_func = []
@@ -1149,7 +1144,7 @@ class ConstraintCollocator(object):
             partials = list(con.atoms(sm.Function))
             num_vals_per_func.append(len(partials))
             jac = sm.Matrix([con]).jacobian(partials)
-            jac = jac.subs(def_map)
+            jac = jac.subs(self.instance_constraint_free_map)
             funcs.append(sm.lambdify(([free] +
                                       list(self.known_parameter_map.keys())),
                                      jac, modules=[{'ImmutableMatrix':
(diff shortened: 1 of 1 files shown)


moorepants:
> See https://github.com/chris-konrad/opty/issues/1 for status and plans.

This would be nice if were an issue here, because forks are often temporary or just may disappear at some point.

chris-konrad:
> This would be nice if were an issue here, because forks are often temporary or just may disappear at some point.

I added it to the first message in this PR

chris-konrad:
I added the new indices of the new timeshift contraint partials to the jacobian indices. This assumes the following layout of the Jacobian. Can you confirm that this is correct, @moorepants? Green is the existing Jacobian, yellow is the extension with timeshift constraints. 

The constraint definition is as above: 
$G_i = [\frac{\partial c_i}{\partial u[i]}, \frac{\partial c_i}{\partial \tau}]$

Creating a new problem with `Problem()` now runs through without exceptions. I did not test it to any extend so far though.

Peter230655:
@chris-konrad : Curiosity: Once this works will opty be able to handle time delayed eoms, like $\dfrac{sx}{dt = f(x(t - \tau), u(x - \tau}$

chris-konrad:
@Peter230655: No, unfortunately this will not directly be possible. The implemented approach requires that the timeshifted trajectory is known trajectory.

Peter230655:
> @Peter230655: No, unfortunately this will not directly be possible. The implemented approach requires that the timeshifted trajectory is a known trajectory.

Thanks!
Still good for me: I am mostly a user of sympy mechanics and of opty, playing around with them to pass time in retirement.
This, when implemented will enlarge the playing field for me. :-)

chris-konrad:
All previously existing unittests run fine but it seems to break some examples with variable times. This is little surprising as I did not pay any attention to that during implementation. Before addressing these issues, I will focus on testing if the new functions work at all.

opty/tests/test_direct_collocation.py
```diff
@@ -1566,3 +1566,51 @@ def test_for_algebraic_eoms():
...
+class TestConstraintCollocatorTimeshiftConstraints():
+
+    def setup_method(self):
+        
+        T_0, a_0, a_1, t, t_d = sym.symbols('T_0, a_0, a_1, t, t_d')
+        x_1, x_2, u, u_shift = sym.symbols('x_1, x_2, u, u_shift', cls=sym.Function)
+        self.eom = sym.Matrix(((x_1(t).diff() - x_2(t),),
+                          (x_2(t).diff() + a_0 * T_0 * x_1(t) 
+                                              + a_1 * T_0 * x_2(t) 
+                                              - T_0 * u(t-t_d),)))
```
moorepants, line 1580: Why would u_shift not be added here instead of u(t-t_d)? Shouldn't the eom's not have t_d present?

opty/tests/test_direct_collocation.py
```diff
@@ -1566,3 +1566,51 @@ def test_for_algebraic_eoms():
...
+                          (x_2(t).diff() + a_0 * T_0 * x_1(t) 
+                                              + a_1 * T_0 * x_2(t) 
+                                              - T_0 * u_shift(t),)))
+        
+        
+
+        self.state_symbols = (x_1, x_2)
+        self.constant_symbols = (a_0, a_1)
+        self.constant_values = (1, np.sqrt(2))
+        self.timeshift_inputs = {u_shift(t): (u(t - t_d), t_d)}
```
moorepants, line 1592: Why did you choose adding a new attribute here, instead of passing u_shift - u(t - t_d) directly to instance constraints?

opty/tests/test_direct_collocation.py
chris-konrad: This is the eom that I pass to the constructor of `Problem`. It has `t_d` present in `u(t-t_d)`. The `ConstraintCollocator` then looks through this eom, finds all expressions of the form U(time_symbol +/- symbol) and substitutes them with U_shift(time_symbol.

opty/tests/test_direct_collocation.py
chris-konrad: The `ConstraintCollocator` automatically appends instance constraints representing u_shift(t) - u(t - t_d) to the other instance constraints, I just didn't write a test for this yet. The attribute `ConstraintCollocator.timeshift_trajectory_substitutes` that I check in this test is just a map from the substitutes to the original trajectory and the timeshift parameter that I occasionally need throughout the code to look up which functions and parameters relate to the timeshift.

chris-konrad:
> > This assumes the following layout of the Jacobian. Can you confirm that this is correct, @moorepants?
> 
> Are you treating τ the same as any unknown parameter? i.e. is it in the list of `unknown_parameters`? This would have some bearing on the construction of the jacobian (order of rows).

Thats what I do. Indeed there seems to be a problem in the Jacobian index function related to this. I will have a look.

chris-konrad:
Added unittest to check eom discretization and the jacobian matrix. Should be fine, but IPOPT/cyipopt still complains about invalid jacobian row indices: 

```
This is Ipopt version 3.14.16, running with linear solver MUMPS 5.7.1.

Exception of type: INVALID_TNLP in file "D:\bld\ipopt_1717602782538\work\src\Interfaces\IpTNLPAdapter.cpp" at line 1100:
 Exception message: eval_jac_g returned false in GetSpaces

EXIT: Some uncaught Ipopt exception encountered.
ERROR:cyipopt:b'Invalid row indices returned from jacobianstructure'
Some uncaught Ipopt exception encountered.
objective value:    0.000000
identified T_0: 1.000000
true T_0:       10.000000

```

opty/direct_collocation.py
```diff
@@ -1680,6 +1958,9 @@ def constraints(free):
             eom_con_vals = func(free_states, all_specified, all_constants,
                                 time_interval)
 
+            if all_specified.shape[0] == 1:
+                all_specified.squeeze()
```
moorepants, line 1962: Should be safe to call squeeze() without the if statement, as it will only affect arrays with an empty dimension.

Peter230655:
I may have found a simple hack to do this:
In the time delayed driving force $u(t - \tau)$ I replaced t with a variable T, and in the eom I set $\dfrac{d}{dt} T(t) = 1.0$

moorepants:
Yes, this sounds like a useful example to have.

moorepants:
Are you doing `u(T(t)  - tau)`?

Peter230655:
> Are you doing `u(T(t) - tau)`?

Yes, this is exactly what I do. So far looks o.k.

moorepants:
I don't think the internals of opty support anything but `t` in the function `Function('u')(t)`.

Peter230655:
Of course only you know this best!
I used ``sm.sin(T(t) - $\tau)$. My simple thinking was this: opty does accept functions of state variables and treats them correctly.
So if I introduce a state valiable with the smae numerical values as ``t`` it might work.
So far looks o.k., but I will play around more.
Actually I used torque = $F \cdot \sin(\omega \cdot (T(t) - \tau)$ and let it estimate F, $\omega, \tau$.

moorepants:
` s i n ( T ( t ) − τ ) ` will work, but not if the function is a state variable, e.g. `x(T(t) - t)` or input variable `r(T(t) - t)`. I think all of the instance constraint code only works with `x(t)` or `r(t)` and should break (or give incorrect results) otherwise.

moorepants:
Feel free to prove me wrong though!

Peter230655:
I do not think, you are wrong! :-)
I also do not think, a state variable will work with anything but ``t``, but this is not what I need - if I understood Chris' problem correctly.

moorepants:
>  I also do not think, a state variable will work with anything but t, but this is not what I need - if I understood Chris' problem correctly.

Christoph precisely needs an unknown input to have a delay, so `r(t - tau)` instead of `r(t)`.

Peter230655:
Then I missunderstood, was too easy to be true! 😊😊
So is opty expected to estimate the delay and what function r is?

Peter230655:
> > I also do not think, a state variable will work with anything but t, but this is not what I need - if I understood Chris' problem correctly.
> 
> Christoph precisely needs an unknown input to have a delay, so `r(t - tau)` instead of `r(t)`.

@moorepants This has been dormant for a while. I am still confused: @chris-konrad wants to estimate the unknown delay of an unknown input function, so estimate $u(t - \tau)$ where both u and $\tau$ are unknown. How is this possible in principle?
$v(t) := u(t - \tau)$ would also work. so how could opty distinguish between the functions $u$ and $v$ ?
Not urgent, I just stumbled across this looking for something else.

chris-konrad:
I want to find the timeshift tau of an otherwise known function u(t-tau). I know the shape of the input, just not the timeshift. 

To make it a little more concrete: I have a scenario where a dynamic system is exposed to an input step. I know the input level it started with and I know the input level it ended with, but I don't know when the jump occured. 

It is the classic case of trying to find an unknown input delay. This could be for example a network latency, some fixed processing delay or a human reaction time.

PR #298, opened 2025-01-12
car around pylons

Peter230655:
I do not believe, such an example as described below is in examples-gallery.
A car in the X/Y plane should drive close to two points and return to its starting point.
I want *opty* to decide at what intermediate times it reaches the points.
To the best of my knowledge, opty does not allow variable intermediate times, but they must be of the form:
$t_{intermediate} = integer \cdot h$, with $0 < integer < num_{nodes}$ and h = interval time.

So, I proceeded as follows:
- I make a differentiable function hump(x, a, b, gr) with hump = 1 if $a \leq x \leq b$ and hump = 0 otherwise. (gr detertmines the steepness of the hump). So, if (xb, yb) is a point to be reached, and $\epsilon$ an small positive number then $punkt_{dt}$ := $hump(x, xb-\epsilon, xb+\epsilon, 5) \cdot hump(y, yb-\epsilon, yb+\epsilon, 5) \approx 1$ when the car is close to the point, and = 0 otherwise.
- In order to know whether the car has come close, I integrate over the run, that is punkt = $\int_{t0}^{t_f} punkt_{dt} \  dt$. So, punkt > 0 if the car has come close to the point. punktdt and punkt become state variables.
- As I do not know the exact value of punkt, I need one more state variable and one more specified variable.

This is explained in detail in the simulation.
It works, unless the hump is too steep. Mathematically it is differentiable infinitely often, but I guess numerically it ceases to be so, if too steep (?).

This _crutch_ comes at the expense of 2 additional state variables and one specified variable per point, but I could not come up with a better way to do it.


Peter230655:
Just info: I added a very similar example, just a sailboat instead of a car, to pst-notebooks. It is called sailboat_around_buoy_opty.py

moorepants:
This is very nice too. It shows how to manage the intermediate time values, which we don't have. You are becoming a pro!

Peter230655:
> This is very nice too. It shows how to manage the intermediate time values, which we don't have. You are becoming a pro!

Thanks! :-)
I believe it is less of a pro and more of an old man who has time to play around and is stubburn as you undobtedly must have noticed!

PR #301, opened 2025-01-19
docstrings changed for solve to match opty

Peter230655:
Added a function solve to direct_collocation as suggested by JM above to get docstrings which better match opty.
In my tests in ran fine.
1. I do not know how the check the visual appearance of what I did, so I do not know, how it will look in the opty documentation
2. It seems to me, that the output of info['g'] has shape (n*(N-1) + o, ) not (n*N + o, ). This is only based on numerical trials, no understanding behind it.

### DIFF
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -202,7 +202,7 @@ class Problem(cyipopt.Problem):
     def solve(self, free):
         """
 
-        **solve**(x)
+        **solve(x)**
 
         Returns the optimal solution and an info dictionary.
 
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -197,6 +197,65 @@ class Problem(cyipopt.Problem):
 
         self.obj_value = []
 
+    # this is added only to correct the docstring, which is a bit different with
+    # opty compared to cyipopt.
+    def solve(self, free):
+        """
+        **solve**(x)
+
+        Returns the optimal solution and an info dictionary.
+
+        Solves the posed optimization problem starting at point x.
+
+        Parameters:
+
+            x (array-like, shape(n*N + q*N + r + s, ))
+
+                Initial guess.
+
+        Returns:
+
+            x (array, shape(n*N + q*N + r + s, ))
+
+                optimal solution.
+
+            info (dictionary)
+
+                x: ndarray, shape(n*N + q*N + r + s, )
+
+                    optimal solution
+
+                g: ndarray, shape(n*(N-1) + o, )
+
+                    constraints at the optimal solution
+
+                obj_val: float
+
+                    objective value at optimal solution
+
+                mult_g: ndarray, shape(n*(N-1) + o, )
+
+                    final values of the constraint multipliers
+
+                mult_x_L: ndarray, shape(n*N + q*N + r + s, )
+
+                    bound multipliers at the solution
+
+                mult_x_U: ndarray, shape(n*N + q*N + r + s, )
+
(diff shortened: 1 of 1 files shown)


moorepants:
The docstrings need to follow the numpydoc format, see here: https://numpydoc.readthedocs.io/en/latest/format.html

moorepants:
I would recommend copying the cyipopt docstring and then edit that so that the formatting and information match. We don't really want to remove any info that is in the cyipopt docstring, but just change the parts that are needed for opty:

https://github.com/mechmotum/cyipopt/blob/master/cyipopt/cython/ipopt_wrapper.pyx#L636

Peter230655:
I did copy - but from the part of the opty documentation instead of from the code directly. :-( 
I will change it, it is easy.
Question: is my point 2. above, concerning the shape of info['g'] correct? My numerical trials seem to say so.

moorepants:
It doesn't look like you copied this: https://github.com/mechmotum/cyipopt/blob/master/cyipopt/cython/ipopt_wrapper.pyx#L636

You are also missing the option arguments to solve.

Peter230655:
I did copy it. I did not push it, yet, because I am unsure about my point 2 above.
Should I just push it?

Peter230655:
Now I copied the docstring directly from the cyipopt code, and change the parts to make the opty docs consistent.
Based on some tests, I assumed the **shape of info['g'] to be (n*(N-1) + o)**. Should this be wrong, I can easily correct it.

opty/direct_collocation.py
```diff
@@ -197,6 +197,54 @@ def __init__(self, obj, obj_grad, equations_of_motion, state_symbols,
...
+            ``mult_x_U``: :py:class:`numpy.ndarray`, shape `(n*N + q*N + r + s, )`
+                bound multipliers at the solution
+            ``status``: :py:class:`int`
+                gives the status of the algorithm
+            ``status_msg``: :py:class:`str`
+                gives the status of the algorithm as a message
+
+        """
+
+        return super().solve(free)
```
moorepants, line 246: This lien is not correct because it will not handle the optional keyword arguments.

opty/direct_collocation.py
```diff
@@ -197,6 +197,54 @@ def __init__(self, obj, obj_grad, equations_of_motion, state_symbols,
 
         self.obj_value = []
 
+    # this is added only to correct the docstring, which is a bit different with
+    # opty compared to cyipopt.
```
moorepants, line 201: These two lines are not need. If is common to overwrite methods in sub classes for this reason and it needs no explanation.

opty/direct_collocation.py
```diff
@@ -197,6 +197,54 @@ def __init__(self, obj, obj_grad, equations_of_motion, state_symbols,
...
+            ``mult_x_U``: :py:class:`numpy.ndarray`, shape `(n*N + q*N + r + s, )`
+                bound multipliers at the solution
+            ``status``: :py:class:`int`
+                gives the status of the algorithm
+            ``status_msg``: :py:class:`str`
+                gives the status of the algorithm as a message
+
+        """
+
+        return super().solve(free, lagrange=[], zl=[], zu=[])
```
moorepants, line 246:
```suggestion
        return super().solve(free, lagrange=lagrange, zl=zl, zu=zu)
```

Needs that changed, otherwise the arguments that the user passes in will never reach this method.

Peter230655: Shows my weakness in python - after so many years! :-(
Corrected as persuggestions.

PR #302, opened 2025-01-20
AEs only

Peter230655:
opty can solve equations of motion which only consist of algebraic equations. I simply removed the lines which test for differential equations.
Surely this is not the main goal of opty!
Still it might be useful to have this option, say, to get th steady state behaviour of a set of differential equations - in any case it does not hurt to have the option, without stressing it,or even pointing it out in the documentation.

### DIFF
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -157,11 +157,11 @@ class Problem(cyipopt.Problem):
 
         """
 
-        if equations_of_motion.has(sm.Derivative) == False:
-            raise ValueError('No time derivatives are present.' +
-                ' The equations of motion must be ordinary ' +
-                'differential equations (ODEs) or ' +
-                'differential algebraic equations (DAEs).')
+    #    if equations_of_motion.has(sm.Derivative) == False:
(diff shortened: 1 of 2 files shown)


Peter230655:
I think, it is bound to fail **test_for_algebraic_eoms** because I precisely removed the lines in direct_collocation.py testing for this.
This additional possible feature is not very important, one can always add a dummy differential equation to the eoms.
@moorepants Best I simply close this PR?

moorepants:
The purpose of opty has always been strictly for solving optimal control problems that involve differential equations. There are many other tools that solve other optimization problems, including optimization of a set of algebraic equations. So unless you have a convincing argument as to why opty should support this (e.g. no other tool in Python does it), then I don't think we should support algebraic-only equations.

moorepants:
Secondly, if we merge this support, you will have to rewrite much of the documentation to explain the theory and all the details of this support. The documentation is all currently written from the perspective that this tool solves optimal control problems in the time domain.

Peter230655:
I agree! It would distract from opty's main purpose without really adding anything substantial.

Issue #203, opened 2024-08-04
Make parse_free a method on Problem


moorepants:
It would be helpful to have something like:

```
Problem().parse_free(solution)
```

then you wouldn't have to pass in all the extra args to the on in utils.py.

moorepants:
Sure, unless you can think of something that is more informative.

Peter230655:
**No,** just wanted to make sure there are no 'conflicts'. My python skills are limited.
NB: about the many things I pushed I feel the one "detailed eoms" may be the most interesting one.

Peter230655:
#303 was merged, so this can be closed, I think.

moorepants:
I've reopened because it looks like the HTML rendering was not checked:

![Image](https://github.com/user-attachments/assets/26b0f35e-8093-4062-96ba-262099ee597a)

It does not render properly.

https://opty.readthedocs.io/latest/api.html#opty.direct_collocation.Problem.parse_free

Peter230655:
How can I check the HTML rendering?
My suscpicion is that the problem is the "where", which is the only thing I changed.

Peter230655:
Like I check the rendering for my examples-gallery things (?)
Let me try!

moorepants:
Yes

Peter230655:
It worked! I changed the docstrings, I hope I did not mess up the pushing.

Peter230655:
I think, maybe now it can be closed (?)

moorepants:
Looks good!

Peter230655:
> Looks good!

Thanks!
This stuff is more difficult for me than making sympy mechanics / opty make a drone fly inside a tube and go through some narrow gate on its course! :-)

PR #303, opened 2025-01-21
add parse_free method to Problem class, add test fubction

Peter230655:
I added parse_free(solution) to the Problem class, and added a test function comparing the new method to the one existing in utils. This should take care of issue #203

### DIFF
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -633,26 +633,7 @@ class Problem(cyipopt.Problem):
         q = self.collocator.num_unknown_input_trajectories
         variable_duration = self.collocator._variable_duration
 
-        len_states = n * N
-        len_specified = q * N
-
-        free_states = free[:len_states].reshape((n, N))
-
-        if q == 0:
-            free_specified = None
-        else:
-            free_specified = free[len_states:len_states + len_specified]
-            if q > 1:
-                free_specified = free_specified.reshape((q, N))
-
-        if variable_duration:
-            free_time_interval = free[-1]
-            free_constants = free[len_states + len_specified:-1]
-            return free_states, free_specified, free_constants, free_time_interval
-        else:
-            free_constants = free[len_states + len_specified:]
-            return free_states, free_specified, free_constants
-
+        return parse_free(free, n, q, N, variable_duration)
 
 class ConstraintCollocator(object):
     """This class is responsible for generating the constraint function and the
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -597,6 +597,62 @@ class Problem(cyipopt.Problem):
 
         return ax
 
+    def parse_free(self, free):
+        """Parses the free parameters vector and returns it's components.
+
+        Parameters
+        ----------
+        free : ndarray, shape(n*N + q*N + r + s)
+            The free parameters of the system.
+
+        where:
+
+        - N : number of collocation nodes
+        - n : number of unknown state trajectories
+        - q : number of unknown input trajectories
+        - r : number of unknown parameters
+        - s : number of unknown time intervals (s=1 if variable duration, else s=0)
+
+
+        Returns
+        -------
+        states : ndarray, shape(n, N)
+            The array of n states through N time steps.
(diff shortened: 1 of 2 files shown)


opty/direct_collocation.py
```diff
@@ -597,6 +597,62 @@ def plot_objective_value(self):
...
+            if q > 1:
+                free_specified = free_specified.reshape((q, N))
+
+        if variable_duration:
+            free_time_interval = free[-1]
+            free_constants = free[len_states + len_specified:-1]
+            return free_states, free_specified, free_constants, free_time_interval
+        else:
+            free_constants = free[len_states + len_specified:]
+            return free_states, free_specified, free_constants
```
tjstienstra, line 654:
```suggestion
        return parse_free(n, q, N, variable_duration)
```
This is indeed a nice feature. You can remove some code duplicaton by just calling the `parse_free` function.

Peter230655: Thanks!
I thought of this but I was not sure util's parse_free is automatically available.
Does it not have to be 'imported'?

tjstienstra: Yes, it needs to be imported, which you can see is done already at the top of this module.

opty/tests/test_direct_collocation.py
```diff
@@ -1566,3 +1566,125 @@ def test_for_algebraic_eoms():
         )
 
     assert excinfo.type is ValueError
+
+
+def test_prob_parse_free():
```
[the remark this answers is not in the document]
tjstienstra: To create a simple equation of `n` symbols, one can use something like `sum(sm.symbols(f"p:{n}"))`.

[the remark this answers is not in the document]
tjstienstra: If you'd like it to have a physical meaning. You can just imagine a set of n independent point masses that can only move in one direction by example.

opty/direct_collocation.py
Peter230655: Now I noticed.
So I am 'allowed' to use util's parse_free?

tjstienstra: Yes, it is even preferable as it reduces code duplication, i.e. same code in multiple locations.

Peter230655: I already changed as per your suggestion.
Somehow, I feel 'more comfortable' with my clumpsy test function. I can't really say why. I know yours is correct.

Peter230655:
Simplified completely as per Timo's suggestion!
I left my test function, somehow I feel more comfortable with it - likely dumb feeling.

opty/direct_collocation.py
Peter230655: > Yes, it is even preferable as it reduces code duplication, i.e. same code in multiple locations.

Thanks!
This really helped my understanding on such packages!

opty/tests/test_direct_collocation.py
```diff
@@ -1566,3 +1566,125 @@ def test_for_algebraic_eoms():
...
+    initial_guess1 = initial_guess
+
+    # check whether same results.
+    statesu, controlsu, constantsu = parse_free(initial_guess1,
+            len(state_symbols), len(control_symbols), num_nodes)
+
+    states, controls, constants = prob.parse_free(initial_guess)
+    assert(np.all(states == statesu))
+    assert(np.all(controls == controlsu))
+    assert(np.all(constants == constantsu))
```
moorepants, line 1652: The better testing for floating point comparison is `np.testing.test_allclose()`.

Peter230655:
I used assert(np.allclose(a, b))
Would np.assert_allclose(a, b) still be better?

opty/tests/test_direct_collocation.py
moorepants: Sorry, it is `np.testing.assert_allclose()`.

Peter230655:
Question:
is _np.testing.assert_allclose(a, b)_ better than _assert(np.allclose(a, b))_ because it is all numpy and hence faster?
Thanks!

moorepants:
This looks good. Thanks!

Peter230655:
> This looks good. Thanks!

:-)

PR #304, opened 2025-01-22
 made HTML rendering a bit better

Peter230655:


Peter230655:
I close this, because I made a mistake in pushing it.

PR #306, opened 2025-01-25
trial push whether layout is o.k.

Peter230655:
This is only to see whether the rendering is as it should be. Not finished, yet
I followed the numpydoc explanation as I understood it, but not sure, whether the rendering is as per expectation.

### DIFF
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -656,9 +656,86 @@ class ConstraintCollocator(object):
     - o : number of instance constraints
     - nN + qN + r + s : number of free variables
     - n(N - 1) + o : number of constraints
+    - Some of the attributes are explained in more detail under Parameters below.
+
+    Attributes
+    ----------
+    ==========
+
+    current_discrete_state_symbols : n-tuple
+        The symbols for the current discrete states.
+    current_discrete_specified_symbols : q-tuple
+        The symbols for the current discrete specified inputs.
+    discrete_eom : sympy.Matrix, shape(n, 1)
+        Discretized equations of motion. Depending on the integration method
+        used.
+    eom: sympy.Matrix, shape(n, 1)
+        The equations of motion used.
+    input_trajectories : tuple
+        known_input_trajectories + unknown_input_trajectories.
+    instance_constraints : o-tuple
+        The instance constraints used in the optimization.
+    integration_method : str
+        The integration method used.
+    known_parameters : tuple
+        The symbols of the known parameters in the problem.
+    known_parameter_map : dict
+        A mapping of known parameters to their values.
+    known_trajectory_map : dict
+        A mapping of known trajectories to their values.
+    known_trajectory_symbols : (m-q)-tuple
+        The known trajectory symbols.
+    next_discrete_specified_symbols : q-tuple
+        The symbols for the next discrete specified inputs.
+    next_discrete_state_symbols : n-tuple
+        The symbols for the next discrete states.
+    node_time_interval : float or sympy.Symbol
+        The time interval between the collocation nodes. float if the interval
+        is fixed, sympy.Symbol if the interval is variable.
+    num_collocation_nodes : int
+        Number of times spaced evenly between the initial and final time of
+        the optimization = N.
+    num_constraints : int
+        The number of constraints = (num_collection_nodes-1)*num_states +
+        len(instance_constraints).
+    num_free : int
+        Number of variables to be optimized = n*N + q*N + r + s.
+    num_input_trajectories : int
+        The number of input trajectories = len(input_trajectories).
+    num_instance_constraints : int
+        The number of instance constraints = len(instance_constraints).
+    num_known_trajectories : int
+        The number of known trajectories = len(known_trajectory_symbols).
+    num_parameters : int
+        The number of parameters = len(parameters).
+    num_states : int
+        The number of states = len(state_symbols) = n.
+    num_unknown_input_trajectories : int
+        The number of unknown input trajectories =
+        len(unknown_input_trajectories).
+    num_unknown_parameters : int
+        The number of unknown parameters = r.
+    parameters : tuple
+        known_parameters + unknown_parameters.
+    parallel : bool
+        Whether to use parallel processing or not.
+    previous_discrete_state_symbols : n-tuple
+        The symbols for the previous discrete states.
+    show_compile_output : bool
+        Whether to show the compile output or not.
+    state_derivative_symbols : n-tuple
+        symbols for the time derivatives of the states.
+    time_symbol : sympy.Symbol
+        The symbol used to represent time, usually `t`.
+    tmp_dir
+        The temporary directory used to store files generated.
+    unknown_input_trajectories : q-tuple
+        The unknown input trajectories symbols.
+    unknown_parameters : r-tuple
+        The unknown parameters in the problem, in the sequence in which they
+        appear in the solution of the optimization.
 
     """
-
     def __init__(self, equations_of_motion, state_symbols,
                  num_collocation_nodes, node_time_interval,
                  known_parameter_map={}, known_trajectory_map={},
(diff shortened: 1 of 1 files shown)


Peter230655:
I think I added all attributes of the class ConstraintCollocator.
I added a line in the class Problem on how to get the attributes.
Following questions:
- I described the attributes to the best of my knowledge - but I am not the resident expert on opty. Probably worth to have a look at them.
- I follwed the numpy docstring instructions the best I knew, but I do not know whether the rendering is what is expected.

opty/direct_collocation.py
```diff
@@ -727,6 +729,77 @@ def __init__(self, equations_of_motion, state_symbols,
             If True, STDOUT and STDERR of the Cython compilation call will be
             shown.
 
+        Attributes
+        ----------
+        Some of the attributes are explained in more detail under Parameters above.
+
+        eom: sympy.Matrix
```
moorepants, line 736:
```suggestion
        eom: sympy.Matrix, shape(2*n, 1)
```

It is helpful to add shape information to all Matrices and/or arrays.

opty/direct_collocation.py
```diff
@@ -727,6 +729,77 @@ def __init__(self, equations_of_motion, state_symbols,
...
+            The equations of motion used.
+        state_symbols : tuple
```
moorepants, line 738:
```suggestion
        state_symbols : 2*n-tuple
```

You can indicate the length of the tuple like so. Or 3-tuple, n-tuple, etc.

Peter230655: > You can indicate the length of the tuple like so. Or 3-tuple, n-tuple, etc.

I will so this.
Rendering o.k.?

moorepants: I don't  know, you will have to work out the size of any iterables.

Peter230655: > I don't know, you will have to work out the size of any iterables.

I am not sure I understand:
if, by definition, n is the number of state variables, the would state_symbols not have to be an n-tuple and eom would have to be an (n,1) sympy Matrix?

moorepants:
>  Rendering o.k.?

You have to check that yourself on your computer.

opty/direct_collocation.py
moorepants: I think so, but I don't remember off the top of my head.

Peter230655:
> > Rendering o.k.?
> 
> You have to check that yourself on your computer.

I did check on my computer, but I do not know whether this is the expected/desired rendering

opty/direct_collocation.py
Peter230655: > I think so, but I don't remember off the top of my head.

These are the definitions in the docstring of the class, so I take them.

opty/direct_collocation.py
```diff
@@ -727,6 +729,77 @@ def __init__(self, equations_of_motion, state_symbols,
             If True, STDOUT and STDERR of the Cython compilation call will be
             shown.
 
+        Attributes
+        ----------
+        Some of the attributes are explained in more detail under Parameters above.
```
moorepants, line 734: This line should be removed or moved. It curently renders as an attribute (in bold font).

opty/direct_collocation.py
```diff
@@ -727,6 +729,77 @@ def __init__(self, equations_of_motion, state_symbols,
             If True, STDOUT and STDERR of the Cython compilation call will be
             shown.
 
+        Attributes
+        ----------
```
moorepants, line 733:
```suggestion
        ----------
        ==========
```

match heading level of parameters

opty/direct_collocation.py
```diff
@@ -727,6 +729,77 @@ def __init__(self, equations_of_motion, state_symbols,
...
+        num_free : int
+            Number of variables to be optimized
+        time_symbol : sympy.Symbol
+            The symbol used to represent time, usually `t`
+        tmp_dir
+            The temporary directory used to store files generated
+        parallel : bool
+            Whether to use parallel processing or not.
+        show_compile_output : bool
+            Whether to show the compile output or not.
```
moorepants, line 801: since there are so many I think it would be best to put them into alphabetical order

opty/direct_collocation.py
```diff
@@ -727,6 +729,77 @@ def __init__(self, equations_of_motion, state_symbols,
...
+        state_symbols : tuple
+            The state symbols used. Functrion of time.
+        state_derivative_symbols : tuple
+            symbols for the time derivatives of the states.
+        num_states : int
+            The number of states = len(state_symbols).
+        num_collocation_nodes : int
+            Number of times spaced evenly between the initial and final time of the optimization.
+        node_time_interval : float or sympy.Symbol
+            The time interval between the collocation nodes. float if the time is fixed, sympy.Symbol if the time is a variable.
```
moorepants, line 747: all lines need to be wrapped to 79 characters.

opty/direct_collocation.py
```diff
@@ -727,6 +729,77 @@ def __init__(self, equations_of_motion, state_symbols,
...
+        num_parameters : int
+            The number of parameters = len(parameters).
+        known_trajectory_map : dict
+            A mapping of known trajectories to their values, see also above in Parameters
+        known_trajectory_symbols : tuple
+            The known trajectory symbols.
+        num_known_trajectories : int
+            The number of known trajectories = len(known_trajectory_symbols).
+        unknown_input_trajectories : tuple
+            The unknown input trajectories symbols.
```
moorepants, line 767: Some statements have periods, some do not.

moorepants:
Some of the attributes in the list should not be changed after creation of a `ConstraintCollocator`.

I was wondering if we should some how indicate that. I guess the Pythonic way to do that is to create property style attributes that only have getter methods. We can think about this later.

opty/direct_collocation.py
moorepants, line 733: This section is not supposed to be in the `__init__` docstring. It is supposed to be in the class's primary docstring.

Peter230655: I will make these changes and push again.

Peter230655:
> Some of the attributes in the list should not be changed after creation of a `ConstraintCollocator`.
> 
> I was wondering if we should some how indicate that. I guess the Pythonic way to do that is to create property style attributes that only have getter methods. We can think about this later.

_create property style attributes that only have getter methods._
My python is too limited to understand this on the spot.

opty/direct_collocation.py
moorepants: But you missed that it should be below the parameters in the primary class docstring.

Peter230655: Fair enough! :-)  :-)

moorepants: I typically put parameters in the `__init__` docstring, but over the years the numpydoc format converged to it being recommended to put it in the class docstring but I never changed my ways. We can move it, but I think there are some related sphinx settings we should adjust and we should do it for all classes.

Peter230655: > I typically put parameters in the `__init__` docstring, but over the years the numpydoc format converged to it being recommended to put it in the class docstring but I never changed my ways. We can move it, but I think there are some related sphinx settings we should adjust and we should do it for all classes.

Just curiosity: Why would numpy give such recommendstions, rather than the python organisation?

moorepants: Python does have a style guide for docstrings but historically the scientific projects found that insufficient, so they created a style guide for NumPy. Almost all scientific python software uses the numpydoc format. There are many things like this, for example conda is not the Python org official package manager, it is pip. But pip doesn't solve all the needs of the scientific community, so conda was created.

moorepants:
> > Some of the attributes in the list should not be changed after creation of a `ConstraintCollocator`.
> > I was wondering if we should some how indicate that. I guess the Pythonic way to do that is to create property style attributes that only have getter methods. We can think about this later.
> 
> _create property style attributes that only have getter methods._ My python is too limited to understand this on the spot.

Here is a lesson on the topic: https://realpython.com/python-getter-setter/

Peter230655:
As I said before, I tried to explain the attributes to the best of my knowledge - but I am not the resident opty expert. :-(

moorepants:
> but I am not the resident opty expert. :-(

I would beg to differ :)

opty/direct_collocation.py
```diff
@@ -654,9 +656,86 @@ class ConstraintCollocator(object):
     - o : number of instance constraints
     - nN + qN + r + s : number of free variables
     - n(N - 1) + o : number of constraints
+    - Some of the attributes are explained in more detail under Parameters below.
+
+    Attributes
+    ----------
+    ==========
```
moorepants, line 663: The "Attributes" section is supposed to be above the "Notes" section, as per the numpydoc format.

The underline for the headers, i.e. "====" should be a single line. See the reStructuredText documentation, e.g. https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections

opty/direct_collocation.py
```diff
@@ -654,9 +656,86 @@ class ConstraintCollocator(object):
     - o : number of instance constraints
     - nN + qN + r + s : number of free variables
     - n(N - 1) + o : number of constraints
+    - Some of the attributes are explained in more detail under Parameters below.
```
moorepants, line 659: This line does not belong in the list, but you can add it in the notes section as a normal paragraph line.

Peter230655:
> > but I am not the resident opty expert. :-(
> 
> I would beg to differ :)

:-)
Maybe I am an o.k. user of opty by now - I enjoy what it can do.

Peter230655:
No matter what I tried, I do not get it to show the heading Attributes. I must have tried every combination and permutation I could think of.

moorepants:
I can only assume that the numpydoc formatter does not display it on purpose.

opty/direct_collocation.py
```diff
@@ -128,6 +128,8 @@ class Problem(cyipopt.Problem):
                       eomn2, ... eomnN,
                       c1, ..., co]
 
+    The attributes may be accessed as follows: Problem_instance.collocator.name_of_attribute
```
moorepants, line 131:
```suggestion
    The attributes may be accessed as follows: ``Problem_instance.collocator.name_of_attribute``
```

opty/direct_collocation.py
```diff
@@ -641,6 +643,82 @@ class ConstraintCollocator(object):
...
+        A mapping of known trajectories to their values.
+    known_trajectory_symbols : (m-q)-tuple
+        The known trajectory symbols.
+    next_discrete_specified_symbols : q-tuple
+        The symbols for the next discrete specified inputs.
+    next_discrete_state_symbols : n-tuple
+        The symbols for the next discrete states.
+    node_time_interval : float or sympy.Symbol
+        The time interval between the collocation nodes. float if the interval
+        is fixed, sympy.Symbol if the interval is variable.
```
moorepants, line 678:
```suggestion
        is fixed, ``sympy.Symbol`` if the interval is variable.
```

moorepants:
LGTM!

Peter230655:
> LGTM!

Thanks! :-)

PR #307, opened 2025-01-27
Testing the RTD preview builds.

moorepants:


moorepants:
@Peter230655 I just discovered this new feature from read the docs:

image

It builds a preview of the documentation from the pull request. Just click "Details" when the check is green.

Peter230655:
> @Peter230655 I just discovered this new feature from read the docs:
> 
> image
> 
> It builds a preview of the documentation from the pull request. Just click "Details" when the check is green.

Neat! I will try as soon as the tests are finished.
By the way: it seems you have added a test, now 22 tests.

Peter230655:
Works !!
I saw this:
image
1. It is in bold face. Was this inteded? (You disliked it with me before)
2.

moorepants:
No, it isn't intended. I guess that numpydoc treats everything in that section as attributes and you can't put arbitrary text.

moorepants:
> With your 2nd sentence you clearly state not to change attributes. My feel is a user of opty should be smart enough not to change attributes after this warning, so maybe 'protecting' them as you suggested earlier may not be necessary?

Protecting them is just a further courtesy we have the option of extending to the user..

Peter230655:
> > With your 2nd sentence you clearly state not to change attributes. My feel is a user of opty should be smart enough not to change attributes after this warning, so maybe 'protecting' them as you suggested earlier may not be necessary?
> 
> Protecting them is just a further courtesy we have the option of extending to the user..

Let me finish reading the RealPython article you suggested, and I will see what I understood.

moorepants:
> Let me finish reading the RealPython article you suggested, and I will see what I understood.

It is not a necessity, just a possible thing we could do.

Peter230655:
> No, it isn't intended. I guess that numpydoc treats everything in that section as attributes and you can't put arbitrary text.

Now regular face. This preview is neat!

moorepants:
>  But it does not show on the github page when I cleack read the docs in the upper right corner. Will this take a while?

The builds take some time. It has to build our examples, which are slow.

moorepants:
There is also browser caching and cdn caching, things don't update instantly.

Peter230655:
> > Let me finish reading the RealPython article you suggested, and I will see what I understood.
> 
> It is not a necessity, just a possible thing we could do.

Would the following be a 'legal' way to make the attributes read-only?
For example:
- I replace self.eom = equations_of_motion with self._eom = equations_of_motion.
- I add this  @property
                   def eom(self):
                        return self._eom

moorepants:
Yes, that is the general approach.

Peter230655:
> Yes, that is the general approach.

Thanks!
I will do this, and then push it.

PR #308, opened 2025-01-28
same betts 10 1 simulation but new branch

Peter230655:
This is the same betts 10 1 as before. As I could not fix the branch, I put it into a new one and made a new PR. I will close the old PR.


moorepants:
I'll merge this one as the demonstration of how to handle inequalities (until we have native support).

examples-gallery/plot_betts_10_1.py
```diff
@@ -0,0 +1,165 @@
+"""
+Alp Rider
+=========
+
+This is example 10.1 from Betts' book "Practical Methods for Optimal Control
+Using NonlinearProgramming", 3rd edition, Chapter 10: Test Problems.
+More details are in sectionn 4.11.7 of the book.
```
moorepants, line 7: Can you cite Bett's book using RST citation notation, i.e. `[Betts2010]_`.

examples-gallery/plot_betts_10_1.py
```diff
@@ -0,0 +1,165 @@
...
+
+The main issue here is an inequality constraint for the state variables:
+:math:`y_1^2 + y_2^2 + y_3^2 + y_4^2 \\geq function(time, parameters)`
+
+As presently *opty* does not support inequality constraints, I introduced a new
+state variable `fact` to take care of the inequality. The inequality is then
```
moorepants, line 13:
```suggestion
state variable ``fact`` to take care of the inequality. The inequality is then
```

double back ticks give monspace font

examples-gallery/plot_betts_10_1.py
```diff
@@ -0,0 +1,165 @@
...
+
+- :math:`y_1, .., y_4` : state variables as per the book
+- :math:`fact`: additional state variable to take care of the inequality
+
+**Specifieds**
+
+- :math:`u_1, u_2` : control variables
+
+Note: I simply copied the equations of motion, the bounds and the constants
+from the book. I do not know their meaning.
```
moorepants, line 29: The why is the name of the problem "Alp Rider"? what does that mean? Please clarify.

examples-gallery/plot_betts_10_1.py
```diff
@@ -0,0 +1,165 @@
...
+        state_symbols,
+        num_nodes,
+        interval_value,
+        instance_constraints=instance_constraints,
+        known_trajectory_map={T(t): times},
+        bounds=bounds,
+        integration_method=integration_method,
+)
+
+#prob.add_option('nlp_scaling_method', 'gradient-based')
```
moorepants, line 127: remove commented lines

examples-gallery/plot_betts_10_1.py
```diff
@@ -0,0 +1,165 @@
...
+# %%
+# Plot the objective function as a function of optimizer iteration.
+prob.plot_objective_value()
+
+# %%
+# Verify that the inequality is always kept, i.e. that fact >= 1.0 always.
+print(f'minimum value of fact = {np.min(solution[4*num_nodes: 5*num_nodes]):.4e}')
+
+# %%
+# sphinx_gallery_thumbnail_number = 2
```
moorepants, line 164: This is displaying in the rendered notebook, so I guess it not correctly specified. It should not show up in the rendered version.

examples-gallery/plot_betts_10_1.py
Peter230655: I simply took the name give to this problem by Dr. Betts.
Below it says:
Stiff ODE, terrain following `[Betts2010]_` Sect. 4.7.6

But my 3rd edition does not seem to have this section.

moorepants: We should just name it something that has a meaning to the reader of opty documentation, as they may not have the book.

Peter230655: In chapter 4.11.7 it is described what the system does:
It must follow steep peaks as closely as possible, see fig 4.6 there.
Maybe: _System following uneven terrain_ ?

examples-gallery/plot_betts_10_1.py
Peter230655: > This is displaying in the rendered notebook, so I guess it not correctly specified. It should not show up in the rendered version.

corrected.

Peter230655:
[Betts_10_1_peaks.pdf](https://github.com/user-attachments/files/18632766/Betts_10_1_peaks.pdf)

These are the peaks the system (whatever it is, I do not know) has to follow.

moorepants:
I am not so concerned with what Bett's has, but I just would like the examples to be understandable without having his book. Can you make sure to provide enough information so that is the case. The name "Alp Rider" was no were else in your text so I don't know what it means.

Peter230655:
> I am not so concerned with what Bett's has, but I just would like the examples to be understandable without having his book. Can you make sure to provide enough information so that is the case. The name "Alp Rider" was no were else in your text so I don't know what it means.

As I understand it from section 4.11.7, this was designed to be a stiff system, without apparent physical meaning. It was proposed to check optimization routines. The inequality constraint forces the norm of the state vector has to be larger than a wildly fluctuating function (the function in the picture), and the objective forces it to be close to this function.

More I cannot find out about it.
Maybe like this:
Title: Stiff Set of Equations
Below something like the 'explanation' I just gave  ?

moorepants:
Yes, perfect!

moorepants:
We're probably not comparing apples to apples though. Opty does not do many of the fancy things (like mesh refinement, variable node spacing, higher order integrators, etc.)

Peter230655:
> We're probably not comparing apples to apples though. Opty does not do many of the fancy things (like mesh refinement, variable node spacing, higher order integrators, etc.)

Fair enough, but if opty's minimum is lower and all the constraints are met very closely, does this not mean it works very well?

moorepants:
I think it still depends on the integrator and steps you use, as the constraints are a function of those choices.

Peter230655:
> I think it still depends on the integrator and steps you use, as the constraints are a function of those choices.

I think you are right: If I understand opty correctly, the results are "correct" only at the points in time given by num_nodes, and by continuity one hopes they are about right in between, too. And this is a tricky system as I now remember again. It took me a while to get it to converge.

moorepants:
Yes, this suffers from the same effects that any integration routine faces and the accuracy is effected by truncation error and floating point error. opty uses very simple integration algorithms relative to what Bett's recommends. Maybe one day I'll have time to include better integrators, but we could also move to pycollo and build that out instead.

Peter230655:
> Yes, this suffers from the same effects that any integration routine faces and the accuracy is effected by truncation error and floating point error. opty uses very simple integration algorithms relative to what Bett's recommends. Maybe one day I'll have time to include better integrators, but we could also move to pycollo and build that out instead.

I just looked around a bit. It seems, pycollo has been kinda dormant for more than a year?

moorepants:
Sam wrote pycollo for his 2021 dissertation. He worked on it some when he was a post doc with me (we used it for the skateboard ollie) and then he moved on to other things. So, yes, it is dormant but it has all of the more advanced collocation methods and it works with SymPy.

moorepants:

Peter230655:
No, there is not. Only reason I did not close it earlier was because you were interested in the name.

Peter230655:
Moved to pst-notebooks

PR #309, opened 2025-01-28
same betts 10 58, but new branch

Peter230655:
I had originally, out of ignorance!, branched off not from master but from some other branch. Hence I could not fix it with the method JM had told me. Consequently I make a new PR and close the old one, which was #259

NB: I think, in this case opty's solution is more accurate than Betts'.

### DIFF
--- a/examples-gallery/plot_betts_10_58.py
+++ b/examples-gallery/plot_betts_10_58.py
@@ -1,0 +1,149 @@
+"""
+Heat Equation
+=============
+
+This is example 10.58 from Betts' book "Practical Methods for Optimal Control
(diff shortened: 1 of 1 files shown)


moorepants:
What is unique about this problem relative to the other examples?

Peter230655:
> What is unique about this problem relative to the other examples?

As far as opty is concerned nothing new.
It just shows how a PDE may be tackled in principle.
I think, o.k. that you close it.

Peter230655:
There is nothing new here compared to other examples.
I move it to pst-notebooks and close this PR

PR #125, opened 2024-02-14
Improve constraint violation visualization

tjstienstra:
Fixes #123 and #53

Changes
---------
Correct the slicing of the constraints_violation vector.
The constraint vector is formatted as (refer to [the creation of the constraint function](https://github.com/csu-hmc/opty/blob/6a23d17493341ad5698e61c6a2dd6e1353b47440/opty/direct_collocation.py#L886) and [the wrapper](https://github.com/csu-hmc/opty/blob/6a23d17493341ad5698e61c6a2dd6e1353b47440/opty/direct_collocation.py#L1440):
```
[con_1_2, ..., con_1_N,
con_2_2, ..., con_2_N,
...,
con_n_2, ..., con_n_N,
*instance_constraints]
```
where `con_1_2, ..., con_1_N` the constraint violation is at every node for the first discrete equation of motion.

Only create a single axis for showing the EoM violation, as they are not officially coupled to specific states. Yes, it is correct to say that most will specify the first n//2 as kinematic differential equations, but that is not required.

Only plot the constraint violations of the instance constraints if there are any instance constraints.

PR #310, opened 2025-01-28
Allow separate eom violation plots. replaces PR #273

Peter230655:
This allows to see the violations of the equations of motion in detail, each in a separate plot, if the new optional keyword _detailed_eoms_ is set to True. If set to False or not set, the eom violations will be displayed as before.
This replaces PR #273 which does not meet the last test.
I have tried it on many simulations, and at times found substantial differences in the violations of the eoms, say 10^(-8) for the worst to 10^(-18) for the best is no unusual.


moorepants:
This puts each equation on a new row:
image

Note that we switched to the single axis deliberately here: https://github.com/csu-hmc/opty/pull/125

moorepants:
In general, one is needing to know if any constraints are violated *at all*. What is the reason to make the default plot show the relative differences in the equation of motion violations?

Secondly , have you tested this with very large # of EoMs? Does it always make a plot that isn't extremely squished vertically?

I prefer that the plot functions just work and have few to no knobs to turn. Anyone can create their own plotting functions or manipulate the returned axes. I'd like these to give a default average plot to quickly view with simple function calls.

Peter230655:
> In general, one is needing to know if any constraints are violated _at all_. What is the reason to make the default plot show the relative differences in the equation of motion violations?
> 

As I remember, with some simulation I wanted to see which eom was violated severely, but I must admit I do not remember exactly. I just thought it does not hurt having it.

> Secondly , have you tested this with very large # of EoMs? Does it always make a plot that isn't extremely squished vertically?

I tested it with many of my simulations, and never noticed this. Max number maybe 15 or 20 probably not more.

> 
> I prefer that the plot functions just work and have few to no knobs to turn. Anyone can create their own plotting functions or manipulate the returned axes. I'd like these to give a default average plot to quickly view with simple function calls.

The optional detailed_eoms is set to false. So, the knob need not be turned. Then everything is exactly like the current version.

Peter230655:
For info:
The drone example in examples-gallery has 14 eoms.
Detailed printout looks like this:
image

Peter230655:
> This puts each equation on a new row: image
> 
> Note that we switched to the single axis deliberately here: #125

This only happens if the eoms should be printed individually, otherwise nothng changes.

Peter230655:
> > This puts each equation on a new row: image
> > Note that we switched to the single axis deliberately here: #125

But they are related to the eoms, so one can see the violation of each eom?
> 
> This only happens if the eoms should be printed individually, otherwise nothng changes. What I am saying now is **not** a proof, just an observation: Since I made this PR for detailed eoms, I must have run 50 or more simulations with detailed eoms and without - and never had a problem except that in rare cases (I forgot what caused it) the rounding of the errors in the instance constraints did not work. I pushed a simple PR to fix it.

moorepants:
Can you demonstrate that this works when the # eoms is large and teh # of instance constraints is large.

opty/direct_collocation.py
```diff
@@ -525,16 +531,43 @@ def plot_constraint_violations(self, vector, axes=None):
...
         if axes is None:
-            fig, axes = plt.subplots(1 + num_plots, 1,
-                                     figsize=(6.4, 1.50*(1 + num_plots)),
-                                     layout='compressed')
+            if detailed_eoms == False or self.collocator.num_states == 1:
+                num_eom_plots = 1
+            else:
+                num_eom_plots = self.collocator.num_states
+            fig, axes = plt.subplots(num_eom_plots + num_plots, 1,
+                            figsize=(6.4, 1.75*(num_eom_plots + num_plots)),
```
moorepants, line 539: Why have you made the plots taller? It is already problematic if there many rows of constraints.

opty/direct_collocation.py
```diff
@@ -438,7 +438,7 @@ def plot_trajectories(self, vector, axes=None):
         return axes
 
     @_optional_plt_dep
-    def plot_constraint_violations(self, vector, axes=None):
+    def plot_constraint_violations(self, vector, axes=None, detailed_eoms=False):
```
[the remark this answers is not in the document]
Peter230655: > I think `subplots=True/False` is a better name for this.

I agree, will change it.

opty/direct_collocation.py
```diff
@@ -525,16 +531,43 @@ def plot_constraint_violations(self, vector, axes=None):
...
-            fig, axes = plt.subplots(1 + num_plots, 1,
-                                     figsize=(6.4, 1.50*(1 + num_plots)),
-                                     layout='compressed')
+            if detailed_eoms == False or self.collocator.num_states == 1:
+                num_eom_plots = 1
+            else:
+                num_eom_plots = self.collocator.num_states
+            fig, axes = plt.subplots(num_eom_plots + num_plots, 1,
+                            figsize=(6.4, 1.75*(num_eom_plots + num_plots)),
+                            constrained_layout=True)
```
moorepants, line 540: I thought his is `layout='constrained'`

opty/direct_collocation.py
```diff
@@ -525,16 +531,43 @@ def plot_constraint_violations(self, vector, axes=None):
...
+                    msg = 'st'
+                elif k % 10 == 2:
+                    msg = 'nd'
+                elif k % 10 == 3:
+                    msg = 'rd'
+                else:
+                    msg = 'th'
+
+                axes[i].plot(con_nodes, state_violations[i])
+                axes[i].set_ylabel(f'{str(k)}-{msg} EOM violation')
```
moorepants, line 568: Just use `Eq. 1`, `Eq. 2`.

Peter230655:
> Can you demonstrate that this works when the # eoms is large and the # of instance constraints is large (together)?

I am not sure, I understand. Do you mean, if we have, say, 35 eoms, the plots will look ugly?

moorepants:
Yes, you can find the example in our gallery that has the most eoms and most instance constraints and show what it looks like.

opty/direct_collocation.py
Peter230655: > I thought his is `layout='constrained'`

I do not recall why I did it, it has been a while.
I assumed I did it like this so the plots would not get too small with large number of eoms. Need to double check.

Peter230655:
> Yes, you can find the example in our gallery that has the most eoms and most instance constraints and show what it looks like.

I used your drone, which likely has the mostr eoms. It worked as I recall.

opty/direct_collocation.py
Peter230655: > Just use `Eq. 1`, `Eq. 2`.

MUCH better, will change.

opty/direct_collocation.py
Peter230655: > Why have you made the plots taller? It is already problematic if there many rows of constraints.

Will try to make them smaller, maybe 1.5 will work.

moorepants:
I only see one row of instance constraints in what you show above. We need to prove it works for many eoms + many rows of instance constraints.

moorepants:
please show the entire plot

opty/direct_collocation.py
```diff
@@ -448,6 +448,12 @@ def plot_constraint_violations(self, vector, axes=None):
             The initial guess, solution, or any other vector that is in the
             canonical form.
 
+        detailed_eoms : boolean, optional. If True, the equations of motion
+            will be plotted in a separate plot for each state. Default is False.
+
+        axes : ndarray of AxesSubplot, optional. If given, it is the user's
+            responsibility to provide the correct number of axes.
```
moorepants, line 455: You need to give a clue of what the number of subplots show be if subplots is true or false

opty/direct_collocation.py
```diff
@@ -525,16 +531,43 @@ def plot_constraint_violations(self, vector, axes=None):
...
+                if k in (11,12,13):
+                    msg = 'th'
+                elif k % 10 == 1:
+                    msg = 'st'
+                elif k % 10 == 2:
+                    msg = 'nd'
+                elif k % 10 == 3:
+                    msg = 'rd'
+                else:
+                    msg = 'th'
```
moorepants, line 565: What is this elaborate if statement? It looks like something that would be prone to fail for arbitrary inputs.

Peter230655:
> I only see one row of instance constraints in what you show above. We need to prove it works for many eoms + many rows of instance constraints.

Should work, as they are plotted independently, as I recall. Will test it.

opty/direct_collocation.py
Peter230655: > What is this elaborate if statement? It looks like something that would be prone to fail for arbitrary inputs.

Just optics: I did not want to write 1th, 2th 22th, etc. Taken care of with your suggestion to write Eq 1, Eq 2, etc.

opty/direct_collocation.py
Peter230655: > You need to give a clue of what the number of subplots show be if subplots is true or false

Simple here, I did no think: about it: # of axes = eom.shape[0]
In the case of instance constraints it was much more difficult, so at the time we decided like this:
Let the user not give axes on the first run. Then he can count the plots and give the righe number of axes.
But her can be done easily.

moorepants: Then write in the docstring that you should call the function once and count the subplots.

Peter230655: > Then write in the docstring that you should call the function once and count the subplots.

Will add this:

moorepants:
Wouldn't the gait example have at least two rows of instance constraints?

Peter230655:
> Wouldn't the gait example have at least two rows of instance constraints?



> Wouldn't the gait example have at least two rows of instance constraints?

I will have to look. I thought of my n-link pendulum (not in the examples). There I think, I can make as many eoms and as many instance constraints as I want.

moorepants:
The thing is we don't write unit tests for the plots, so if we don't check them thoroughly manually, we may have edge cases.

Peter230655:
A first attempt. 
[plot_sit_to_stand.pdf](https://github.com/user-attachments/files/19397104/plot_sit_to_stand.pdf)
You got a much better feel for aestetics.

(I was surprised badly the eoms are violated)

moorepants:
That show it works, even for large sets of constraints. Thanks for checking.

Peter230655:
I will test more simulations. Aestetics o.k.?
Thanks!

moorepants:
What you have shown is sufficient, as it is the worst case scenario.

Peter230655:
> What you have shown is sufficient, as it is the worst case scenario.

Thanls!
Does it **look** o.k. in your opinion? I mean the size of the plots, etc.

Peter230655:
I tested on many 15 - 20 excamples in examples-gallery.
[plot_sit_to_stand_subplots=False.pdf](https://github.com/user-attachments/files/19401524/plot_sit_to_stand_subplots.False.pdf)
[plot_sit_to_stand_subplots=True.pdf](https://github.com/user-attachments/files/19401525/plot_sit_to_stand_subplots.True.pdf)

I noticed, that figsize = 1.25 worked fine with subplots=True, but gave sometimes 'ugly' plots with it.
figsize = 1.75 works fine with subplots=False, but too large for subplots=True.
So, I set it accordingly.

I added a quark ``figsize`` to give the use additional flexibility.
The figsizes I have preset (1.25 / 1.75) worked fine for all my tests.

Peter230655:
I am snsure about the error:
I says a problem with ``pprint(par_map)`` in plot_sit_to_stand.py and in plot_human_gait.py, but I did only changed the method``plot_constraint_violations(..)``

moorepants:
The height of the subplot is mostly governed by how long you make the y label.

Peter230655:
> The height of the subplot is mostly governed by how long you make the y label.

I shortend the y - labels when showing distict eom plots.
After trials, what I implemented gave nice plots in all my tests.

opty/direct_collocation.py
```diff
@@ -573,6 +574,22 @@ def plot_constraint_violations(self, vector, axes=None):
...
             canonical form.
 
+        subplots : boolean, optional.
+            If True, the equations of motion will be plotted in a separate
+            plot for each equation of motion.
+            Default is False.
+            If a user wants to provide the axes, it is recommended to run once
+            without providing axes, to see how many are needed.
+
+        figsize : float, optional.
```
moorepants, line 584: This is not needed because you can set it and pass in the axes before call this method.

moorepants:
update with master to see if it fixes the error

opty/direct_collocation.py
Peter230655: > This is not needed because you can set it and pass in the axes before call this method.

Not sure what you mean. Do you mean lines 581, 582?
Or do you mena the kwarg figsize (maybe I should call it fig_height)

moorepants:
```python
fig, ax = subplots(figsize=(2, 2))
problem.plot_something(ax=ax)
```

moorepants: matplotlib has a million settings, we are not in the business of customizing plots. our plots should make the most basic simple plot that doesn't look too awful. if a user wants to customize their plots they can pass in an axis or make their own plot.

Peter230655:
My origin was up-to-date
I did
- pull origin master. It said my master already was up-to-date
- detailed-eoms-1 merge master. I said already up-to-date

Removed ``figsize``. As per ym tests to really needed.

opty/direct_collocation.py
Peter230655: > matplotlib has a million settings, we are not in the business of customizing plots. our plots should make the most basic simple plot that doesn't look too awful. if a user wants to customize their plots they can pass in an axis or make their own plot.

When I play around with matplotlib, I sometimes get the feel, your number above is way too low! :-)

moorepants:
> detailed-eoms-1 merge master. I said already up-to-date

I don't see how that is possible as there are no merge commits here in this branch.

opty/direct_collocation.py
```diff
@@ -573,6 +573,17 @@ def plot_constraint_violations(self, vector, axes=None):
...
+        subplots : boolean, optional.
+            If True, the equations of motion will be plotted in a separate
+            plot for each equation of motion.
+            Default is False.
+            If a user wants to provide the axes, it is recommended to run once
+            without providing axes, to see how many are needed.
+
+        axes : ndarray of AxesSubplot, optional.
+            If given, it is the user's responsibility to provide the correct
+            number of axes.
```
moorepants, line 585: These are not in the same order as the function signature and there should not be an empty line between kwargs in the docstring.

opty/direct_collocation.py
moorepants: I think the kwarg should be `layout='constrained'`.

Peter230655: > I think the kwarg should be `layout='constrained'`.

I tried, it did not work for me, but I can try again.

Peter230655: > > I think the kwarg should be `layout='constrained'`.
> 
> I tried, it did not work for me, but I can try again.

Works the same way, it seems, so I changed to it.

Peter230655:
> > detailed-eoms-1 merge master. I said already up-to-date
> 
> I don't see how that is possible as there are no merge commits here in this branch.

I did it again, same results.
Maybe easiest to close this PR and do a new one?

opty/direct_collocation.py
moorepants: The official argument is explained here: https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.figure.html#matplotlib.pyplot.figure we should use what is the docs.

moorepants:
>  Maybe easiest to close this PR and do a new one?

No, we figure out what is wrong here.

Peter230655:
> > Maybe easiest to close this PR and do a new one?
> 
> No, we figure out what is wrong here.

This way, I will earn soemthing for sure!

opty/direct_collocation.py
Peter230655: > The official argument is explained here: https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.figure.html#matplotlib.pyplot.figure we should use what is the docs.

I likely got my ``constrained_layout=True`` from chatGPT, I do not recall. I changed anyway.

moorepants:
I did this in pygait2d yesterday: https://github.com/csu-hmc/gait2d/commit/1b86c50154fcc73a206b3f55900f4272365f6b90 it must be causing an issue.

Peter230655:
> I did this in pygait2d yesterday: [csu-hmc/gait2d@1b86c50](https://github.com/csu-hmc/gait2d/commit/1b86c50154fcc73a206b3f55900f4272365f6b90) it must be causing an issue.

...but if this was merged, I should experience the same problem?

moorepants:
Not if you haven't updated gait2d.

opty/direct_collocation.py
```diff
@@ -604,7 +602,7 @@ def plot_constraint_violations(self, vector, axes=None, subplots=False):
         bars_per_plot = None
         rotation = -45
 
-        if subplots == False:
+        if subplots:
```
Peter230655, line 605: Did you not replace the original condition with itsa opposite?

opty/direct_collocation.py
```diff
@@ -668,22 +666,22 @@ def plot_constraint_violations(self, vector, axes=None, subplots=False):
         con_nodes = range(1, self.collocator.num_collocation_nodes)
 
         if axes is None:
-            if subplots == False or self.collocator.num_states == 1:
+            if subplots or self.collocator.num_states == 1:
```
Peter230655, line 669: Did you not replace the original condition with its opposite?

opty/direct_collocation.py
```diff
@@ -668,22 +666,22 @@ def plot_constraint_violations(self, vector, axes=None, subplots=False):
...
+                                                            num_plots)),
+                                     layout='constrained')
 
         else:
             num_eom_plots = len(axes) - num_plots
 
         axes = np.asarray(axes).ravel()
 
-        if subplots == False or self.collocator.num_states == 1:
+        if subplots or self.collocator.num_states == 1:
```
Peter230655, line 684: Didyou not replace the original condition with its opposite?

opty/direct_collocation.py
moorepants: oop, yes, it should be `is False` not `== False`

Peter230655: > oop, yes, it should be `is False` not `== False`

Is ``is False`` more pythonic than ``== False``?

Peter230655:
Thanks! Sorry it caused so much work for you!

PR #287, opened 2024-12-30
removed two lines each at around line 1200 and 1600 to allow one eom …

Peter230655:
I push this before, burt from bad branch. Now hopefully correct.

I removed two lines each around line 1200 and 1600 to allow opty to use one eom (must be differential) only.
I added a test function in test_direct_collocation.py

PR #311, opened 2025-01-28
Allow for one differential equation only. Removed lines near 1300, 1680

Peter230655:
I removed two lines each, so opty allows one _differential equation_ of motion only. This replaces PR #287, which does not meet the last test doc/readthedocs... 

I found nothing in the documentation that says explicitly that more than one eom is required.

### DIFF
--- a/examples-gallery/plot_betts_10_7.py
+++ b/examples-gallery/plot_betts_10_7.py
@@ -2,7 +2,7 @@
 Hypersensitive Control
 ======================
 
-This is example 10.7 from `_[Betts2010]`.
+This is example 10.7 from [Betts2010]_ and demonstrates working with a single differential equation.
 
 **States**
 
--- a/examples-gallery/plot_betts_10_7.py
+++ b/examples-gallery/plot_betts_10_7.py
@@ -1,0 +1,136 @@
+"""
+Hypersensitive Control
+======================
+
+This is example 10.7 from `_[Betts2010]`.
+
+**States**
+
+- y : state variable
+
+**Specifieds**
+
+- u : control variable
+
+"""
+
+import numpy as np
+import sympy as sm
+import sympy.physics.mechanics as me
+from opty.direct_collocation import Problem
+from opty.utils import create_objective_function
+import matplotlib.pyplot as plt
+
+# %%
+# Equations of motion.
+
+t = me.dynamicsymbols._t
+y, u = me.dynamicsymbols('y u')
+
+eom = sm.Matrix([-y.diff(t) - y**3 + u])
+sm.pprint(eom)
+
+# %%
+# I packed the optimization problem and its solution into a function, so that
+# I can easily call it with different parameters.
+
+def solve_optimization(nodes, tf):
+    t0, tf = 0.0, tf
+    num_nodes = nodes
+    interval_value = (tf - t0)/(num_nodes - 1)
+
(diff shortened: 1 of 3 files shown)


moorepants:
I'm fine with merging this, but it needs two things 1) unit test(s) for the single EoM and 2) all documentation to be updated to reflect this (if needed). I'm sure I assumed that there would be 2 or more states when writing everything, so there has to be a pass through all the documentation to make sure that assumption is not built in to any text.

moorepants:
It is also worth noting that you removing these lines exposes that we had no prior unit test checking that these two errors were raised.

Peter230655:
> I'm fine with merging this, but it needs two things 1) unit test(s) for the single EoM and 2) all documentation to be updated to reflect this (if needed). I'm sure I assumed that there would be 2 or more states when writing everything, so there has to be a pass through all the documentation to make sure that assumption is not built in to any text.

1.
I thought I made a test called _test_one_eom_only_, at least my local branch shows it.

2.
I did not find anything in opty Documentation Release 1.4.0.dev0, dated September 09, 2024.
Is there a more detailed documentation I missed?

moorepants:
This PR shows one file changed, I do not see any tests here.

moorepants:
It would be helpful to also include in this same PR a new example with a single EoM (I think you have one already), that way we can see that all the plotting and other functionality works for a single EoM in the same PR (which act as a set of tests in essence).

Peter230655:
> This PR shows one file changed, I do not see any tests here.

My stupidity with git! I made a new PR, because the old one did not pass the doc/writethedoc... test. 
I will add it and push.

Peter230655:
> It would be helpful to also include in this same PR a new example with a single EoM (I think you have one already), that way we can see that all the plotting and other functionality works for a single EoM in the same PR (which act as a set of tests in essence).

Where would I put this example? in opty/examples-gallery?

moorepants:
Yes, you already have a PR for 1 EoM, so instead of merging it (that shows the workaround) we can add that example here, but without the work around.

Peter230655:
> Yes, you already have a PR for 1 EoM, so instead of merging it (that shows the workaround) we can add that example here, but without the work around.

Just not make sure I understand correctly:
In this branch one-eom-enough-new-1 I would add the test function to _test_direct_collocation.py_ and I would add my simulation with one eom to examples-gallery.

moorepants:
Yes

moorepants:
Why are there no bars in the instance constraint plots?

image

examples-gallery/plot_betts_10_7.py
```diff
@@ -0,0 +1,136 @@
+"""
+Hypersensitive Control
+======================
+
+This is example 10.7 from `_[Betts2010]`.
```
moorepants, line 5:
```suggestion
This is example 10.7 from [Betts2010]_.
```

Peter230655:
> Why are there no bars in the instance constraint plots?
> 
> image

As per my experience, if the instance constraints are met with very high accuracy I get this.

moorepants:
ok

Peter230655:
Thanks! :-)

Peter230655:
> It is also worth noting that you removing these lines exposes that we had no prior unit test checking that these two errors were raised.

They WERE! This is how I got started. I did not know that minimum 2 eoms were needed.

PR #312, opened 2025-01-28
betts 10 103 and 104. Replaces PR # 261

Peter230655:
This replaces PR #261 which cannot pass the newest test.
Essentially the same problem, once with only differential equation and once with an algebraic equation are compared.


moorepants:
Nice comparison, thanks!

PR #313, opened 2025-01-30
make attributes of ConstraintCollocator read-only

Peter230655:
I tried to make the attributes of ConstraintCollocator to be read-only, as changing them likely will break opty.
The one I did not mess around with was integration_method as it issues an error anyway if a non-valid method is set.
I also added a test function.

NB: num_free is also an attribute of Problem. As it is often used like this: initial_guess = np.zeros(prob.num_free) I left it as it is.

### DIFF
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -1023,8 +1023,12 @@ class ConstraintCollocator(object):
         return self._num_known_input_trajectories
 
     @property
-    def integration_method(self.elf):
+    def integration_method(self):
         return self._integration_method
+
+    @property
+    def current_discrete_specified_symbols(self):
+        return self._current_discrete_specified_symbols
 
     @integration_method.setter
     def integration_method(self, method):
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -182,16 +182,16 @@ class Problem(cyipopt.Problem):
         self.con_jac_rows, self.con_jac_cols = \
             self.collocator.jacobian_indices()
 
-        self.num_free = self.collocator.num_free
-        self.num_constraints = self.collocator.num_constraints
+        self.num_free = self.collocator._num_free
+        self._num_constraints = self.collocator.num_constraints
 
         self._generate_bound_arrays()
 
         # All constraints are expected to be equal to zero.
-        con_bounds = np.zeros(self.num_constraints)
+        con_bounds = np.zeros(self._num_constraints)
 
         super(Problem, self).__init__(n=self.num_free,
-                                      m=self.num_constraints,
+                                      m=self._num_constraints,
                                       lb=self.lower_bound,
                                       ub=self.upper_bound,
                                       cl=con_bounds,
@@ -206,9 +206,9 @@ class Problem(cyipopt.Problem):
         N = self.collocator.num_collocation_nodes
         num_state_nodes = N*self.collocator.num_states
         num_non_par_nodes = N*(self.collocator.num_states +
-                               self.collocator.num_unknown_input_trajectories)
+                               self.collocator._num_unknown_input_trajectories)
         state_syms = self.collocator.state_symbols
-        unk_traj = self.collocator.unknown_input_trajectories
+        unk_traj = self.collocator._unknown_input_trajectories
         unk_par = self.collocator.unknown_parameters
 
         if self.bounds is not None:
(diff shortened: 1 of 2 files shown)


Peter230655:
I am at a **complete** loss, why it failed test_discrete_symbols.
It ran fine with my test, it ran fine with the examples of examples-gallery.
No idea what I did to fail this test.

Peter230655:
One of the error messages says:
image
If I look at the direct_collocation.py I changed, it does hav such an attribute.

Peter230655:
Question: 
If something in direct_collocation.py is changed then part of the test-suite is that all the examples of examples-gallery are run?
Thanks!

moorepants:
You will need to resolve merge conflicts here, some of the other recent merged PRs must conflict.

opty/tests/test_direct_collocation.py
```diff
@@ -1687,3 +1687,120 @@ def obj_grad(free):
...
+        eom,
+        state_symbols,
+        num_nodes,
+        interval_value,
+        time_symbol=t,
+        known_parameter_map=par_map,
+        instance_constraints=instance_constraints,
+        bounds=bounds,
+        integration_method='midpoint',
+    )
```
moorepants, line 1766: no need to create a problem for this test, just create a `ConstrainCollocator`.

opty/direct_collocation.py
```diff
@@ -810,66 +816,218 @@ def __init__(self, equations_of_motion, state_symbols,
...
             self._find_closest_free_index()
             self.eval_instance_constraints = self._instance_constraints_func()
             self.eval_instance_constraints_jacobian_values = \
                 self._instance_constraints_jacobian_values_func()
         else:
-            self.num_instance_constraints = 0
+            self._num_instance_constraints = 0
+
+    @property
+    def eom(self):
```
moorepants, line 876: I think if you make properties, then numpydoc's guidance is to put the docstring for each property here at the property. Then they become more self documenting, i.e. if you delete a property you will delete the docstring.

opty/tests/test_direct_collocation.py
Peter230655: > no need to create a problem for this test, just create a `ConstrainCollocator`.

Would this not be a bit dangerous? When I replaced attritube by _attribute I do this in my VSC replacing all occurencies. So, in the class Problem I also replaced it by accident. I noticed when I ran a simulation.

opty/direct_collocation.py
Peter230655: > I think if you make properties, then numpydoc's guidance is to put the docstring for each property here at the property. Then they become more self documenting, i.e. if you delete a property you will delete the docstring.

Do I understand correctly:
Instead of lumping all attributes as I have done, I would put the explanation as a docstring under each function?

opty/tests/test_direct_collocation.py
moorepants: This change only affects the ConstraintCollocator class.

opty/direct_collocation.py
moorepants: The recommendation says "When referring to an attribute anywhere within the docstring, enclose its name in single backticks. Attributes that are properties and have their own docstrings can be simply listed by name:"

https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring

Peter230655: image

moorepants: The doctring of an attribute/property should have it's type, shape (if matrix or array), and a description of the attribute.

Peter230655: > The doctring of an attribute/property should have it's type, shape (if matrix or array), and a description of the attribute.

I understand.
But as I understood numpydocs, this information should also be in the list of attributes, or would there be no list and just the functions with their docstrings?

moorepants: You'll have to experiment.

moorepants: I searched "how to docstring property attribute numpydoc" which led to:

https://stackoverflow.com/questions/48864008/how-to-properly-document-properties-with-numpydoc

which led to:

https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

Peter230655: > I searched "how to docstring property attribute numpydoc" which led to:
> 
> https://stackoverflow.com/questions/48864008/how-to-properly-document-properties-with-numpydoc
> 
> which led to:
> 
> https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

Thanks!
I think, clearer now.

Peter230655:
Due to the conflict, this was replace by PR #315
Hence, I close it.

PR #315, opened 2025-02-03
Read only attributes 1

Peter230655:
This replaces PR #313, which had conflicts.
The attributes of ConstraintCollocator should be read only.


Peter230655:
There was a warning in a test, treated as error, that I had a double description of the attributes. So, I removed the description in the attributes list and let it at the properties only.

moorepants:
You'll need to fix merge conflicts in this branch. I think I can merge otherwise.

Peter230655:
> You'll need to fix merge conflicts in this branch. I think I can merge otherwise.

I was afraid this would happen when I saw that you merged the ``duplicate state variables`` :-)
``duplicate state variables`` is only a few lines plus a small test function, while ''read only atrributes`` is an error prone mess if I try to start again in a new branch.
Question: can I manually add the changes of ``duplicate state variables`` into ``read only attributes`` and then push again?

moorepants:
No, you should always pull upstream master and then merge it into this branch. On the merge you make have to fix merge conflicts.

moorepants:
Some documentation on it: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/addressing-merge-conflicts/resolving-a-merge-conflict-using-the-command-line?platform=windows

Peter230655:
> Some documentation on it: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/addressing-merge-conflicts/resolving-a-merge-conflict-using-the-command-line?platform=windows

So basically like this?
- I get my local master up-to-date. (That I know how to do)
- I checkout to the local branch read-only-attribute-1
-  I put git merge master.

Then I should see the conflicts and be able to manually correct them.
Then I push again.

moorepants:
Yep!

Peter230655:
> Yep!

Thanks!
I will try my luck this afternoon, now I go to my math teaching - fever or otherwise :-)

opty/tests/test_direct_collocation.py
```diff
@@ -61,7 +61,7 @@ def test_pendulum():
         bounds={T(t): (-2.0, 2.0)},
         show_compile_output=True)
 
-    assert prob.collocator.num_instance_constraints == 4
+    assert prob.collocator._num_instance_constraints == 4
```
moorepants, line 64: No existing tests should be changed.

opty/tests/test_direct_collocation.py
```diff
@@ -1747,6 +1747,98 @@ def test_one_eom_only():
...
+        'show_compile_output',
+        'state_derivative_symbols',
+        'state_symbols',
+        'time_interval_symbol',
+        'time_symbol',
+        'tmp_dir',
+        'unknown_input_trajectories',
+        'unknown_parameters',
+        ]:
+        print(XX)
```
moorepants, line 1838: don't print in tests

opty/tests/test_direct_collocation.py
Peter230655: > No existing tests should be changed.

Not sure I understand: .num_instance constraints is not an attribute anymore- It is **._**num_instance_constraints now?

opty/tests/test_direct_collocation.py
Peter230655: Shit! I forgot to remove!
Will correct it!

opty/tests/test_direct_collocation.py
moorepants: We have not changed that `collocator.num_instance_constraints == 4` should be true. None of these tests should be changed.

Peter230655: > We have not changed that `collocator.num_instance_constraints == 4` should be true. None of these tests should be changed.

In ``ConstraintCollocator`` i changed all occurences of ``collocator.attribute`` to ``._collocator._attribute``. Then I added all these @priority functions. Does this mean that ``.attribute`` is still considered an attribute?
As I recall, before I did all these changes in ``test_direct_collocation.py`` I got lots of errors when I pushed.

moorepants: Yes, it is an attribute. The `@property` decorator does not change that. The decorator lets you do more interesting things with attributes when people "get" or "set" them.

moorepants: https://stackoverflow.com/questions/17330160/how-does-the-property-decorator-work-in-python

moorepants: >  As I recall, before I did all these changes in test_direct_collocation.py I got lots of errors when I pushed.

When you get errors from tests, that means you are breaking tests that shouldn't be broken. It isn't the tests that are wrong, but the code you changed.

Peter230655: In direct_collocation.py I changed all .attributes to ._attributes.
Was this wrong?

moorepants: No, that is fine (and correct), but you shouldn't change tests.

opty/direct_collocation.py
```diff
@@ -183,15 +183,15 @@ def __init__(self, obj, obj_grad, equations_of_motion, state_symbols,
             self.collocator.jacobian_indices()
 
         self.num_free = self.collocator.num_free
-        self.num_constraints = self.collocator.num_constraints
+        self._num_constraints = self.collocator._num_constraints
```
moorepants, line 186: This line should have not changed.

opty/direct_collocation.py
```diff
@@ -183,15 +183,15 @@ def __init__(self, obj, obj_grad, equations_of_motion, state_symbols,
...
 
         self._generate_bound_arrays()
 
         # All constraints are expected to be equal to zero.
-        con_bounds = np.zeros(self.num_constraints)
+        con_bounds = np.zeros(self._num_constraints)
```
moorepants, line 191: Should not change.

opty/direct_collocation.py
```diff
@@ -183,15 +183,15 @@ def __init__(self, obj, obj_grad, equations_of_motion, state_symbols,
...
 
         super(Problem, self).__init__(n=self.num_free,
-                                      m=self.num_constraints,
+                                      m=self._num_constraints,
```
moorepants, line 194: should not change

opty/direct_collocation.py
moorepants: Nothing in `Problem` should be changed.

opty/direct_collocation.py
```diff
@@ -858,72 +783,398 @@ def __init__(self, equations_of_motion, state_symbols,
             shown.
 
         """
-        self.eom = equations_of_motion
+        self._eom = equations_of_motion
```
moorepants, line 786: should not change

opty/tests/test_direct_collocation.py
Peter230655: I will change the test back and try tomorrow morning.

opty/direct_collocation.py
```diff
@@ -991,28 +1242,28 @@ def _sort_parameters(self):
         The unknown parameters are sorted by name."""
 
         # TODO : Should the full parameter list be sorted here for consistency?
-        parameters = self.eom.free_symbols.copy()
-        if self.time_symbol in parameters:
-            parameters.remove(self.time_symbol)
+        parameters = self._eom.free_symbols.copy()
+        if self._time_symbol in parameters:
+            parameters.remove(self._time_symbol)
```
moorepants, line 1247: When you use the attribute it should not have a leading underscore, only when you set the attribute should it have a leading underscore.

opty/direct_collocation.py
Peter230655: Shit! I missed this! I will fix tomorrow morning. I may have accidentally changed more in Problem.

moorepants:
You have a minor misunderstanding of how this works. Here is an example class:

```python
class MyClass():
    def __init__(self, a):
        self._a = a  # only use the underscore when setting the attribute (internal to the class)

    @property  # this makes a read only attribute because we haven't defined a setter method
    def a(self):
        return self._a

    def square_a(self):  # use the attribute normally when "getting" it
        return self.a**2
```

opty/direct_collocation.py
Peter230655: Then I screwed up completely!
Let me try to fix it tomorrow.

Peter230655:
> You have a minor misunderstanding of how this works. Here is an example class:
> 
> ```python
> class MyClass():
>     def __init__(self, a):
>         self._a = a  # only use the underscore when setting the attribute (internal to the class)
> 
>     @property  # this makes a read only attribute because we haven't defined a setter method
>     def a(self):
>         return self._a
> 
>     def square_a(self):  # use the attribute normally when "getting" it
>         return self.a**2
> ```
> 
> A user never sees `_a`:
> 
> ```
> In [3]: m = MyClass(12)
> 
> In [4]: m.a
> Out[4]: 12
> 
> In [5]: m.square_a()
> Out[5]: 144
> 
> In [6]: m.a = 15
> ---------------------------------------------------------------------------
> AttributeError                            Traceback (most recent call last)
> Cell In[6], line 1
> ----> 1 m.a = 15
> 
> AttributeError: property 'a' of 'MyClass' object has no setter
> ```

Not a minor misunderstanding, a **complete** one! 
I think, easiest for me to fix it in a new branch, instead of trying to remove the scores of wrong _ everywhere.
Should get in done over the weekend.

moorepants:
New PR #337 replaces this.

PR #318, opened 2025-02-05
time_vector method added. Should Fix #316

### LINKED ISSUE #316: Would be helpful to have a convenience function that returns the time vector
```python
@property
def time_vector(self):
    return np.arange(0.0, num_nodes*h_val, h_val)
```

```
p = Problem(...)
p.time_vector
```

This can be an attribute as it if very fast to compute. I could be computed on initialization. I guess it should be both on ConstraintCollocator and Problem.

Peter230655:
Added identical time_vector method in Problem and in ConstraintCollocator. Should fix issue #316


moorepants:
This needs to resolve conflicts.

opty/direct_collocation.py
```diff
@@ -686,6 +686,35 @@ def parse_free(self, free):
 
         return parse_free(free, n, q, N, variable_duration)
 
+    def time_vector(self, t0=0.0, solution=None):
```
moorepants, line 689:
```suggestion
    def time_vector(self, solution=None, start_time=0.0):
```

I think people will rarely use the start time kwarg, so make it second. And rename to have more informative variable name.

opty/direct_collocation.py
```diff
@@ -686,6 +686,35 @@ def parse_free(self, free):
...
+            The initial time of the problem. Default is 0.0.
+        solution : (n*N + q*N + r + s)-ndarray, optional
+            The solution to to problem. Needed if the interval is variable.
+
+        Returns
+        -------
+        A numpy num_collocation_nodes-array of time instances.
+
+        """
+        if isinstance(self.collocator.node_time_interval, sm.Symbol):
```
moorepants, line 704: There is a flag `self.collocator._variable_duration` that should be used for this check.

opty/direct_collocation.py
```diff
@@ -686,6 +686,35 @@ def parse_free(self, free):
...
+            if solution is None:
+                msg = 'Solution vector must be provided for variable duration.'
+                raise ValueError(msg)
+            else:
+                return np.arange(t0, t0+self.collocator.num_collocation_nodes*
```
moorepants, line 709:
```suggestion
                return np.arange(t0, t0 + self.collocator.num_collocation_nodes*
```

opty/direct_collocation.py
```diff
@@ -686,6 +686,35 @@ def parse_free(self, free):
...
+                    solution[-1], solution[-1])
+
+        else:
+            return np.arange(t0, t0+self.collocator.num_collocation_nodes*
```
moorepants, line 713:
```suggestion
            return np.arange(t0, t0 + self.collocator.num_collocation_nodes*
```

opty/direct_collocation.py
Peter230655: I now wonder, whether we should give t0 at all: I would not know how to tell opty not to start at t= = 0.0

opty/direct_collocation.py
Peter230655: When do you put a space and when do you not put a space?
I thought in signatures one does not put a space?
Of course I will correct it.

opty/direct_collocation.py
moorepants: We can leave it off and add it later if we ever think we need it.

opty/direct_collocation.py
moorepants: PEP8 has the guidelines.

opty/direct_collocation.py
Peter230655: > We can leave it off and add it later if we ever think we need it.

I left it in. Do you want me to remove it?

moorepants:
This needs a unit test.

Peter230655:
> This needs a unit test.

Started on it. I think, I will get it done tomorrow.

opty/tests/test_direct_collocation.py
```diff
@@ -1923,3 +1923,98 @@ def test_attributes_read_only():
...
+        num_nodes,
+        interval_value,
+        time_symbol=t,
+        backend='numpy'
+    )
+
+    # solution must be given
+    with raises(ValueError):
+        time_vector = prob.time_vector(start_time=t0)
+        time_vector = prob.time_vector()
```
moorepants, line 2001: These need to be in two with clauses, one for each.

Peter230655:
I added ValueErrors for:
- start time >= final time (only with variable time_intervall)
- time intervall <= 0 (only with variable time intervall)

I felt for fixed time interval user will know that t0 < tf

opty/direct_collocation.py
```diff
@@ -686,6 +686,43 @@ def parse_free(self, free):
 
         return parse_free(free, n, q, N, variable_duration)
 
+    def time_vector(self, solution=None, start_time=0.0):
+        """Returns the time instances of the problem as an numpy ndarray.
+
+        Parameters
+        ==========
+        solution : (n*N + q*N + r + s)-ndarray, optional
```
moorepants, line 694: This is not the correct numpy doc format.

The basic format is:

```
variable : array_like, shape(5, 3, 2), optional
```

moorepants:
This is good now! Thanks.

Peter230655:
> This is good now! Thanks.

Thanks!

Issue #257, opened 2024-11-01
Problems with sympy.pi in the equations of motion


Peter230655:
This morning I ran into a problem using opty for a simulation (Example 10.59 from Betts' book)
With the help from @tjstienstra I found the problem was that I used sympy.pi in the equations of motion. Replacing it with numpy.pi solved the problem.
Here I think, one can see that opty objects to sympy.pi:
![image](https://github.com/user-attachments/assets/b4f2353a-de35-49fd-b468-f1699c0a18db)

in the simulation below, I added a factor sympy.cos(symypy.pi) / sympy.cos(numpy.pi) and sympy.pi / numpy.py, in lines 37, 38.
For some reason, it accepts sympy.cos(sympy.pi), but not sympy.pi alone.

NB: 
1. As I recall, I had this problem before, but I do not think, I raised an issue then.
2. Of course no big deal if one knows it, but maybe this could be mentioned in the docs?

```
"""
Parameter Identification: Betts & Huffman 2003
==============================================

This is the problem presented in section 7 of:

Betts, John T. and Huffman, William P. "Large Scale Parameter Estimation
Using Sparse Nonlinear Programming Methods". SIAM J. Optim., Vol 14, No. 1,
pp. 223-244, 2003.

"""

from collections import OrderedDict

import numpy as np
import sympy as sym
import matplotlib.pyplot as plt
from opty.direct_collocation import Problem

duration = 1.0
num_nodes = 100
interval = duration / (num_nodes - 1)

# %%
# Symbolic equations of motion
mu, p, t = sym.symbols('mu, p, t')
y1, y2, T = sym.symbols('y1, y2, T', cls=sym.Function)

state_symbols = (y1(t), y2(t))
constant_symbols = (mu, p)

# %%
# I had to use a little "trick" here because time was explicit in the eoms,
# i.e. setting T(t) to a function of time and then pass in time as a known
# trajectory in the problem.
eom = sym.Matrix([y1(t).diff(t) - y2(t)*sym.cos(sym.pi)/sym.cos(np.pi),
                  y2(t).diff(t) - mu**2 * y1(t) + (mu**2 + p**2) *
                  sym.sin(p * T(t))*sym.pi/np.pi])

# %%
# Specify the known system parameters.
par_map = OrderedDict()
par_map[mu] = 60.0

# %%
# Generate data
time = np.linspace(0.0, 1.0, num_nodes)
[...]

PR #321, opened 2025-02-05
[DO NOT MERGE] Checking to see if failure on Windows with sympy.pi.

moorepants:


moorepants:
Ok, that's clear. I think #257 is fixed. Windows tests failed with compile errors.

PR #322, opened 2025-02-05
Linting, spelling, and wording cleanup on the ball rolling on disc example.

moorepants:
@Peter230655 this is the type of tidying I'd like before a release.

### DIFF
--- a/examples-gallery/plot_ball_rolling_on_spinning_disc.py
+++ b/examples-gallery/plot_ball_rolling_on_spinning_disc.py
@@ -3,22 +3,24 @@ r"""
 Ball Rolling on Spinning Disc
 =============================
 
-A uniform, solid ball with radius :math:`r` and mass :math:`m_b` is rolling
-on a horizontal spinning disc without slipping.
-The disc starts at rest and speeds up like this
+A uniform, solid ball with radius :math:`r` and mass :math:`m_b` is rolling on
+a horizontal spinning disc without slipping. The disc starts at rest and speeds
+up like this
 
-- :math:`u_{3}(t) = \Omega \cdot (1 - e^{(-\alpha \cdot t)})`
+.. math::
+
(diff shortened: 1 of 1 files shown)


examples-gallery/plot_ball_rolling_on_spinning_disc.py
```diff
@@ -3,34 +3,37 @@
...
+ball.
 
 **Constants**
 
 - :math:`m_b`: mass of the ball [kg]
 - :math:`r` : radius of the ball [m]
 - :math:`m_o` : mass of the observer [kg]
 - :math:`\Omega`: final rotational speed of the disc around [rad/sec]
-- :math:`\alpha`: measure of the acceleration of the disc [1/sec]
+- :math:`\alpha`: measure of the acceleration of the disc [rad/sec/sec]
```
moorepants, line 31: @Peter230655 is what I did correct?

Peter230655: > @Peter230655 is what I did correct?

If you look at line 134, it shows that alpha is a measure on how fast the disc reaches its final speed Omega. It dimension should be [1/sec], so alpha*T(t) is dimensionless.

As regards cleaning up, I got your point, I will be unable to catch all the things you can catch - but I will surely try!

moorepants: I changed this because it said "acceleration" which would typically have units like len/time/time or angle/time/time. I'll change it back.

moorepants, line 31:
```suggestion
- :math:`\alpha`: measure of the acceleration of the disc [1/sec]
```

Peter230655: > I changed this because it said "acceleration" which would typically have units like len/time/time or angle/time/time. I'll change it back.

That threw me off for a few minutes.

PR #323, opened 2025-02-05
raise ValueError if duplicate state symbols are present. Fixes #317

### LINKED ISSUE #317: Make better error message if you add a duplicate state
I accidentally added a state variable twice in the list passed to Problem. The compilation fails and there is a C error that says things are declared twice (or something similar). We could catch this before compilation and give a better error message to the user.

Peter230655:
It raises a ValueError if there are either duplicate state symbols, or the number of state symbols does not match the number of eoms. Should fix #317


moorepants:
Looks great, thanks!

Peter230655:
> Looks great, thanks!

Thanks! :-)

PR #330, opened 2025-02-07
removed I wordings

Peter230655:
Removed (hopefully all) remaining I wordings.


Peter230655:
Same bug in different disguise?

PR #333, opened 2025-02-07
Add example for a multiphase problem.

moorepants:
I realized that we can probably do multiphase problems with something like this. But I am currently limited to the duration of each phase having to be equal. @Peter230655 any ideas on how to let this block collide with the wall at any time? My intuition is that I need to move the collision equation from the instance constraints into the DAE somehow.

### DIFF
--- a/examples-gallery/plot_multiphase.py
+++ b/examples-gallery/plot_multiphase.py
@@ -1,0 +1,117 @@
+"""
+Multiphase Collision
+====================
+
+A block is sliding on a surface with Coulomb friction. Push it with a (limited)
+rightward force until it hits a wall 1m on the right. When it hits the wall
+enforce a Newtonian collision with a coefficient of restitution e so that the
+block bounces off the wall and slides backwards to its original location. Apply
+this force such that the block reaches its original location in the minimal
+amount of time and that each phase has the same duration.
+
+"""
+
+import numpy as np
+import sympy as sm
+from opty import Problem
+import matplotlib.pyplot as plt
+
+# %%
+# Start with defining the fixed duration and number of nodes.
+num_nodes = 400
+
+# %%
+# Symbolic equations of motion, note that we make two sets: one before the #
+# collision and one after.
+m, e, mu, g, t, h = sm.symbols('m, e, mu, g, t, h', real=True)
+xr, vr, xl, vl, F = sm.symbols('x_r, v_r, x_l, v_l, F', cls=sm.Function)
+
+state_symbols = (xr(t), vr(t), xl(t), vl(t))
+constant_symbols = (m, e, mu, g)
+specified_symbols = (F(t),)
+
+eom = sm.Matrix([
+    xr(t).diff(t) - vr(t),
+    m*vr(t).diff(t) + mu*m*g - F(t),
+    xl(t).diff(t) - vl(t),
+    m*vl(t).diff(t) - mu*m*g,
+])
+sm.pprint(eom)
+
+# %%
+# Specify the known system parameters.
+par_map = {
+    m: 1.0,
+    e: 0.8,
+    mu: 0.6,
+    g: 9.81,
+}
+
+
+# %%
+# Specify the objective function and it's gradient.
+def obj(free):
+    """Return h (always the last element in the free variables)."""
+    return free[-1]
+
+
+def obj_grad(free):
+    """Return the gradient of the objective."""
+    grad = np.zeros_like(free)
+    grad[-1] = 1.0
+    return grad
+
+
+# %%
+# Specify the symbolic instance constraints, i.e. initial and end conditions
+# using node numbers 0 to N - 1
+dur = (num_nodes - 1)*h
+instance_constraints = (
+    xr(0*h) - 0.0,
+    xr(dur) - 1.0,
+    vr(0*h) - 0.0,
+    # TODO : figure out how to let the collision happen at any time, not just
+    # the halfway point in time
+    vl(0*h) + e*vr(dur),
+    xl(0*h) - 1.0,
+    xl(dur) - 0.0,
+)
+
+bounds = {
+    F(t): (0.0, 160.0),
+    h: (0.0, 0.5),
+    vr(t): (0.0, np.inf),
+    vl(t): (-np.inf, 0.0),
+}
+
+# %%
+# Create an optimization problem.
+prob = Problem(obj, obj_grad, eom, state_symbols, num_nodes, h,
+               known_parameter_map=par_map,
+               instance_constraints=instance_constraints,
+               time_symbol=t,
+               bounds=bounds)
+
+# %%
+# Use a zero as an initial guess.
+initial_guess = np.zeros(prob.num_free)
+
+# %%
+# Find the optimal solution.
+solution, info = prob.solve(initial_guess)
+print(info['status_msg'])
+print(info['obj_val'])
+
+# %%
+# Plot the optimal state and input trajectories.
+prob.plot_trajectories(solution)
+
+# %%
+# Plot the constraint violations.
+prob.plot_constraint_violations(solution)
+
+# %%
+# Plot the objective function as a function of optimizer iteration.
+prob.plot_objective_value()
+
+plt.show()


Peter230655:
> I realized that we can probably do multiphase problems with something like this. But I am currently limited to the duration of each phase having to be equal. @Peter230655 any ideas on how to let this block collide with the wall at any time? My intuition is that I need to move the collision equation from the instance constraints into the DAE somehow.

Let me think about it. :-)
(As I understand, you want it with Newtonian collission, so we have different eoms before and after the impact?)

moorepants:
There is essentially just 1 eom in this case because the block is always sliding:

$$
m\dot{v} = -\textrm{sgn}({v})\mu m g + F
$$

but at $x=1$ the impact occurs and you apply a mometum balance for before and after impact:

$$
v_\textrm{after} = e v_\textrm{before}
$$

If you break the EoM into the moving right and moving left portions, then you can remove the dicontinuity in the EoM. So then you do have 2 EoMs, one for before and one after.

Peter230655:
> There is essentially just 1 eom in this case because the block is always sliding:
> 
> m v ˙ = − s g n ( v ) μ m g + F

As I understood the equations in your code, F = 0 after the impact, the equation above seems to imply F is present always?
> 
> but at x = 1 the impact occurs and you apply a mometum balance for before and after impact:
> 
> v a f t e r = e v b e f o r e
> 
> If you break the EoM into the moving right and moving left portions, then you can remove the dicontinuity in the EoM. So then you do have 2 EoMs, one for before and one after.

I guess with Newtonian impact, by definition the speed will change discontinuously - something opty does not like.
So, two eoms are likely necessary.
Is Newtonian collision important, or would Hunt-Crossley also be o.k.?

I am asking, so I know in which direction to think.

moorepants:
> As I understood the equations in your code, F = 0 after the impact, the equation above seems to imply F is present always?

I wrote one equation in this comment, but I split it into two in the opty formulation.

moorepants:
>  Is Newtonian collision important, or would Hunt-Crossley also be o.k.?

The point of this example is to avoid discontinuities when transitioning between phases of a problem. I can solve this with an impact force, but that's not the purpose. Sometimes you want to model transitions from one eom to another, e.g. standing on the ground, jumping off, and landing but not have to model the contact forces with things such as Hunt-Crossely. Newton's momentum exchange is a perfectly good model fro impact also.

Peter230655:
> > Is Newtonian collision important, or would Hunt-Crossley also be o.k.?
> 
> The point of this example is to avoid discontinuities when transitioning between phases of a problem. I can solve this with an impact force, but that's not the purpose. Sometimes you want to model transitions from one eom to another, e.g. standing on the ground, jumping off, and landing but not have to model the contact forces with things such as Hunt-Crossely. Newton's momentum exchange is a perfectly good model for impact also.

All clear now to me, I think.

Peter230655:
Could this be done as a two stage problem, similar to PR262 ?

moorepants:
> When I did it, I had to optimize the the first and second phases separately - and surely this must be worse (maximally equal) to optimizing them combined. 

Yes, I want to solve these kinds of problems where all phases are solve simultaneously. My example here may be able to be solved by separating into multiple optimal control problems but, in general, we'd like to solve problems this way that cannot be solved separately. That's what I'm trying to do, but there may not be a "hack" to get opty to do this. It may be something we have to build into opty more carefully.

Peter230655:
> > When I did it, I had to optimize the the first and second phases separately - and surely this must be worse (maximally equal) to optimizing them combined.
> 
> Yes, I want to solve these kinds of problems where all phases are solve simultaneously. My example here may be able to be solved by separating into multiple optimal control problems but, in general, we'd like to solve problems this way that cannot be solved separately. That's what I'm trying to do, but there may not be a "hack" to get opty to do this. It may be something we have to build into opty more carefully.

I do not know, but I am getting clearer in the direction of thinking.

Peter230655:
In your program, the impact is halfway through the duration. This can be changed, and the optimum does not seem to be the middle. But no progress in finding a hack, yet

moorepants:
There may be no hack. We probably have to make opty handle multiphase problems properly.

Peter230655:
image

examples-gallery/plot_multiphase.py
```diff
@@ -0,0 +1,117 @@
...
+instance_constraints = (
+    xr(0*h) - 0.0,
+    xr(dur) - 1.0,
+    vr(0*h) - 0.0,
+    # TODO : figure out how to let the collision happen at any time, not just
+    # the halfway point in time
+    vl(0*h) + e*vr(dur),
+    xl(0*h) - 1.0,
+    xl(dur) - 0.0,
+)
```
moorepants, line 78: I could change the requirement that the block simply travels 2 meters total and is stationary at the start and the end. Maybe a solution that had varying periods of no motion at the start and end is possible, where these two stationary periods are not equal in time.

Peter230655: In your original set-up the block was not at rest at the end, you left its speed open. So I also did it in my simulation.
My set up mimicks **exactly** your original set-up, except I let opty decide the point of impact - and it does find a better one! :-)

Peter230655: For clarification:
The plots above were created by moving the time of impact around, from early on in the journey to late in the journey.
Bu every time the time of impact as predetermined! I only wanted to see if the time of impact relative to total journey had an influence on the length of the journey - otherwise there would be nothing to optimize.
It does have an influence of the length of the journey, so there was a chance for opty to find a good time of impact - ehich it does.

Peter230655: I did it like this:

```
teiler = 1.667
zs = int((num_nodes - 1)/teiler) * h
dur = (num_nodes - 1) * h

instance_constraints = (
    xr(0*h) - 0.0,
    xr(zs) - 1.0,
    vr(0*h) - 0.0,
    vl(zs) + e*vr(zs),
    xl(zs) - x_impact,
    xl(dur) - 0.0,
)
```

moorepants: The way I set up the two sets of equations of motion require that each set have the same number of nodes, i.e. t_impact - t0 = tf - t_impact. If the node interval is constant, then how can you vary the time of impact without having different number of nodes before and after impact?

moorepants: Are you somehow using fewer or more nodes in the before period relative the after period? If so, that may work for such a simple problem but otherwise you risk not having enough nodes in one phase. I suppose you can always increase the number of nodes to ensure you have the minimum number of nodes needed for one phase.

Peter230655: I must admit I gave no thought to this.
I always play around with the number of nodes, sometimes too many also hinder convergence as I have seen.

If opty is to decide the time of impact, then the number of nodes for left and right trip may also be different.

YES, it is a simple problem, and I do not know whether my 'hack' -which I am convinced works for this problem- will work for a more complicated one.

moorepants: Here are the constraints for this problem:

1. From t0 to tf we discretize evenly spaced time intervals.
2. We use two continuously differentiable equations of motion that govern motion before and after an event.
3. At some time t_event between t0 and tf we switch between the two sets of equations of motion. At the switch time, the state is shared (an possible modified) between both equations of motion.
4. Our objective operates on something calculated over the entire time interval (tf - t0).

Points 1-4 are the general definition of a 2 phase problem (which is a sub-case of a multi phase problem).

I have demonstrated here that if tf - t_event = t_event - t0, you can solve this with an opty hack (i.e. include both equations of motion and solve simultaneously).

The solution I am looking for is strictly one that involves no smoothened discontinous functions. It must also allow for tf - t_event to not equal t_event - t0.

I think the only solution to this is to allow for the node time interval to have a unique value per phase or to allow the node time interval to vary for every time step. Both would require modifications to opty's internals and there is no hack that can do this (I think).

As an analogy, we can hack opty to support inequality constraints by adding an equation of motion with a new state variable and then bounding that variable to only be positive or negative. I was wondering if there was a similar solution to this problem (because that does not need smoothened discontinous functions). It is a hack that isn't a modeling hack, but simply a hack on opty's design.

Peter230655: > Here are the constraints for this problem:
> 
> 1. From t0 to tf we discretize evenly spaced time intervals.
> 2. We use two continuously differentiable equations of motion that govern motion before and after an event.
> 3. At some time t_event between t0 and tf we switch between the two sets of equations of motion. At the switch time, the state is shared (an possibly modified) between both equations of motion.
> 4. Our objective operates on something calculated over the entire time interval (tf - t0).
> 
> Points 1-4 are the general definition of a 2 phase problem (which is a sub-case of a multi phase problem).

I did not change anything in this respect.
> 
> I have demonstrated here that if tf - t_event = t_event - t0, you can solve this with an opty hack (i.e. include both equations of motion and solve simultaneously).

I think, this is possible also if tf - t_event != t_event - t0. At least it looked to me like it worked when I did it.

> 
> The solution I am looking for is strictly one that involves no smoothened discontinous functions. It must also allow for tf - t_event to not equal t_event - t0.

I think as long as t0 < t_event < tf and t_event _predetermined_ it works in your example, unless I make a very basic mistake in my thinking. Without any smoothend discontinuous functions.
What opty does, as I see it from the trajectories is this: it calculates x_l and v_l starting from t0, so that 
x_l(t_event) = 1.0 (in your case) and v_l(t_event) = -e*v_r(t_event), and x_L(tf) = 0. Whatever t_event is.

What I do when I let opty decide when t_event should be, I simply force x_l = x_r whenever x_r = 1,0 and same for v_l.
For this, I need smoothened functions.
> 
> I think the only solution to this is to allow for the node time interval to have a unique value per phase or to allow the node time interval to vary for every time step. Both would require modifications to opty's internals and there is no hack that can do this (I think).
> 
> As an analogy, we can hack opty to support inequality constraints by adding an equation of motion with a new state variable and then bounding that variable to only be positive or negative. I was wondering if there was a similar solution to this problem (because that does not need smoothened discontinous functions). It is a hack that isn't a modeling hack, but simply a hack on opty's design.

I see what you mean: what I did was a modeling hack, not a hack on opty's design - but it did give results.

Peter230655: I think, when I make t_event != (tf - t0)/2, it is still the cases that both the x_r process and the x_l process run from t0 to tf.
In the picture below, I set t_event approx. 2*(tf-t0)/3
x_l(t0) approx. 2.5
image

moorepants: I think I understand, your problem's duration is much longer. It simulates xr past the impact (> ~ 0.11 seconds). That is nonsense for xr, because it shouldn't continue past  1 meter. If you then look at xl at t=0 it start at 2.5 meter (not 1 meter!). So it seems you are not solving the problem as stated.

Peter230655: YES, mine is much longer.
I think, it solves the _physical_ problem, but one after another, not simultaneously so to speak.

Now I understand what you meant by different time intervals: To solve simultaneously with t_impact != (tf - t0)/2 one would need different length time intervals. 

What is the advantage of doing it simultaneously? I guess saving CPU time?

moorepants: Ok, I know see what you did and that it is likely a valid solution. I would have never thought about simulating the equations of motion outside of the x = 0 to 1 meter range. But if you relax that restriction (a hack), then you can find a solution.

moorepants: > What is the advantage of doing it simultaneously? I guess saving CPU time?

You are solving the problem outside of the bounds of interest. So you have to also ensure that portion can solve. This will depend on CPU as well as your ability to get a solution. It is trivial for this test example, but would not be for a complex model.

Peter230655: Even with this simple example, when I moved t_event too close to t0 or to tf, it would not find a solution easily.
(But I did not pay too much attention to this: for just about any simulation I have played around with, it took some trying different things before it would converge...)

For this step it was only a small step to let opty find t_impact. But again it took trying things before it would converge. 
My feel is, it is hard to judge from ther equations, whether a problem is hard: The brachistochrone were very simple equations, but hard to get a solution.

PR #337, opened 2025-02-08
Changes ConstraintCollocator attributes to read only properties. Replaces PR #315

Peter230655:
Read only attributes:
Corrected the mistakes of previous ignorance: leading underscore only when the attribute is set. 
Removed the printing in the test.


opty/tests/test_direct_collocation.py
```diff
@@ -1814,3 +1814,101 @@ def obj_grad(free):
...
 
+def test_attributes_read_only():
+    """
+    Test to ensure the ConstraintCollocator attributes are read-only.
+    """
+    import sympy as sym
+    import sympy.physics.mechanics as mech
+    from opty.direct_collocation import Problem, ConstraintCollocator
+
+    import pytest
```
moorepants, line 1825: All imports should be at the top of the file with the other imports.

moorepants: I fixed it.

Peter230655: > I fixed it.

**Thanks!**
I always make such mistakes, when I copy the test function which I write to test it (and where I need these imports) to ``test_direct_collocation.py``

The ``time_symbol`` is optional, but as we saw recently, it is a good idea to explicitly add it here and also in ``create_objective_function``. Would it make sense to make it required? (might break old code, so maybe not a good idea?)

moorepants:
There were two arange calls, deleted the second.

opty/tests/test_direct_collocation.py
moorepants: In hindsight, time symbol should be required, but it really only fails on CI. I'm not quite sure why but something with the hashing of python objects changes. Not sure...

moorepants: > (might break old code, so maybe not a good idea?)

Best we don't  break old code.

Peter230655:
:-)

PR #338, opened 2025-02-08
Break examples into beginner, intermediate, and advanced. Fixes #334

### LINKED ISSUE #334: Make two sections in the example gallery: beginner and advanced
We have many examples now, which is good, but new users may not find the right examples to start with. We should divide the example gallery page into a beginner and advanced sections.

moorepants:


moorepants:
This will break some existing urls to some examples, so maybe I should include sphinx-redirects to avoid that for any examples in <= 1.3 that moved into the new subdirectories.

moorepants:
@Peter230655 do you agree on this grouping. I don't want to have to change it later because we will then have to include more URL redirects.

Peter230655:
> @Peter230655 do you agree on this grouping? I don't want to have to change it later because we will then have to include more URL redirects.

I think, 

plot_sliding_block should be beginners.
plot_vyasayani, I am unsure whether it should not be intermediate.

Rest I agree completely

moorepants:
vyasayani is basically the same complexity as betts2003 and they are both the simplest parameter id problem you can do.

Peter230655:
> vyasayani is basically the same complexity as betts2003 and they are both the simplest parameter id problem you can do.

Agreed.

Peter230655:
I just played around with it wanting to see what it looked like.
So I did it like I had always done it, but got this error:
image
What do I have to do to avoid the error?
Thanks!

moorepants:
The `opty-dev-env.yml` has been updated with new packages in a number of these pull requests. You have to update your environment:

```
conda env update -f opty-dev-env.yml
```

Peter230655:
> The `opty-dev-env.yml` has been updated with new packages in a number of these pull requests. You have to update your environment:
> 
> ```
> conda env update -f opty-dev-env.yml
> ```

I managed to delete my existing opty-dev environment completely. Don't ask me how I did it, stupidity would be the likely correct answer. So I tried to create an new one as per the instructions in opt, but I got this:
image

moorepants:
The command to create is:

```
conda env create -f opty-dev-env.yml
```

moorepants:
To remove:

```
conda env remove -n opty-dev
```

You can delete and recreate at will. No issues.

Peter230655:
It created the environment without problems!
But now, when I run a simulations, there is the line, say
``from opty import problem``
and I get an error: there is no library opty,
The I list what is in my opty-dev, and there is numpy and many other things but no opty.

What did I do wrong?
Thanks!

moorepants:
If you are in a directory other than the main opty directory, you will need to "install" opty.

```
cd /path/to/opty/
conda env create -f opty-dev-env.yml
conda activate opty-dev
python -m pip install --no-deps --no-build-isolation -e .
```

The last line installs opty as "editable" in the environment. Then you can run opty from any directory when the environment is activated.

moorepants:
>  Then maybe there is a typo in the instructions:

yes typo

Peter230655:
> > Then maybe there is a typo in the instructions:
> 
> yes typo

Should I raise an issue so it does not get forgotten?

moorepants:
Fixing it would be as quick.

Peter230655:
> Fixing it would be as quick.

If I knew how to do it, I certainly would!

Peter230655:
> If you are in a directory other than the main opty directory, you will need to "install" opty.
> 
> ```
> cd /path/to/opty/
> conda env create -f opty-dev-env.yml
> conda activate opty-dev
> python -m pip install --no-deps --no-build-isolation -e .
> ```
> 
> The last line installs opty as "editable" in the environment. Then you can run opty from any directory when the environment is activated.

At long last everythnig works again! (took me 4 hours to fix something you or Timo would not have destroyed in the first place, and if you have you'd have fixed it in 5 min!)
So I looked at the examples-gallery. Looks **really** nice now!
Also everything runs under windows! Before I had problems with the walkers, but that ``sm.pi`` fix (and adding pygait2d to the dependencies, I assume you did) cleared up the problems:

PR #340, opened 2025-02-09
Move the Park 2004 example to the gallery.

### LINKED ISSUE #252: Move the Park2004 example to the examples-gallery
I developed this example to explore control parameter identification in human standing:

https://github.com/csu-hmc/opty/tree/master/examples/park2004

Park solved the problem in 2004 with shooting methods. In 2020, Wang and van den Bogert solved the problem with direct collocation and a stochastic dynamics model. We could move this example to the gallery and then solve it using Wang's method. My original solution worked but would often find local optima (if I remember correctly). The example would still be done with generated motion data instead of measurements from real people, but could highlight the stochastic approach that we figured out in #250.

moorepants:
Fixes #252

I will not add the Gaussian identification method in #250 for now, just move the example over.

- [x] Add complete explanation.
- [x] Add some plots.
- [x] Add an animation.


moorepants:
```
/home/runner/work/opty/opty/docs/examples/advanced/plot_park2004.rst:24: WARNING: download file not readable: /home/runner/work/opty/opty/docs/examples/advanced/model_park2004.py [download.not_readable]
```

moorepants:
Ok, this should be ready. Took much more fiddling than expected. But I caught a couple bugs and added a new feature.

PR #341, opened 2025-02-09
probable typo corrected

Peter230655:
I think there is a very small typo.
I did this mostly to see whether I am on the right track in such matters.


Peter230655:
So know I know how to do it! :-)

PR #343, opened 2025-02-09
examples-gallery now has three subdirectories

Peter230655:
I think the path given to the example was not fully correct anymore - at least if I understood correctly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -158,11 +158,12 @@ Next download the opty source files and install with::
 Usage
 =====
 
-There are several examples available in the ``examples`` and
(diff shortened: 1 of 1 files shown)


Peter230655:
For my understanding: can something be merged before it has passed all the tests - or is my timing off?

moorepants:
I can press merge at any time and if you just change some text that has no chance to break code functionality or rendering, then there is little reason to wait for tests to pass.

moorepants:
I was too quick to merge this one:
image

Peter230655:
Already I pushed to correct it.

PR #346, opened 2025-02-10
using windows explorer

Peter230655:
I added this, because it took me a while to come up with this, and with Firefox I never managed.
But maybe too obvious to be added?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -197,6 +197,9 @@ and open the result with your web browser, for example::
 
    $ firefox _build/html/index.html
 
+Alternatively, in the windows explorer go to ``opty/docs``, open ``docs`` and
+search for ``index.html``. Several files will show, click them.
+
 Acknowledgements
 ================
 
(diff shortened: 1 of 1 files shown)


Peter230655:
Curious aside: even though all tests passed, I go this email message.
image

README.rst
```diff
@@ -197,6 +197,9 @@ and open the result with your web browser, for example::
 
    $ firefox _build/html/index.html
 
+Alternatively, in the windows explorer go to ``opty/docs``, open ``docs`` and
+search for ``index.html``. Several files will show, click them.
+
```
moorepants, line 202: The file you need to open is `opyt/docs/_build/html/index.html`

moorepants:
Better to copy the instructions Christoph made here for windows:

https://github.com/moorepants/DynamicistToolKit?tab=readme-ov-file#documentation

Peter230655:
Do you mean this portion=
image

I will do so, but two question

moorepants:
I'm just showing you what we put for instructions in a different repository.

Peter230655:
Changed to this line:
The file you need to open is opyt/docs/_build/html/index.html

Issue #348, opened 2025-02-10
Create consistent formatting for all examples in the gallery


moorepants:
Here is a suggestion:

```rst
Title
=====

Objectives
----------

Introduction
------------

Other headings as needed
------------------------

References
----------
```

The objectives section would list the purpose and unique aspects of the example.

I basically have this in the park2004 example: https://opty.readthedocs.io/latest/examples/advanced/plot_park2004.html

We should also try to consistently use the mathematical notation in the theory page. I also tried this in the park 2004 example.

PR #350, opened 2025-02-10
added more textrm, first attempt at structure as per #348

Peter230655:
- Tried to add missing ``textrm`` for nicer looks.
- First attempt at a unified structure as per issue #348 
- had to add lines ``prevent_print = 0`` to avoid nonsense being output, a semicolon did not work.

### DIFF
--- a/examples-gallery/advanced/plot_car_around_pylons.py
+++ b/examples-gallery/advanced/plot_car_around_pylons.py
@@ -2,6 +2,16 @@ r"""
 Car around Pylons
 =================
 
+Objective
+---------
+
+- Show a way to circumvent the current inability of *opty* to handle  instance
+  constraints at times that are not specified in advance but selected optimally
+  by *opty* itself.
+
+Introduction
+------------
+
 A car with rear wheel drive and front steering must move close to two points
 and return to its starting point. It can only move forward. The car is modeled
 as a rigid body with three bodies. The car is driven by a force at the rear
@@ -12,7 +22,8 @@ There is no speed allowed perpendicular to the wheels, realized by two velocity
 constraints. (The trajectories show, that the car is at the limit often, as one
 would expect if, as it is the case here, duration has to be minimized)
 
-The **main point** in this simulation is this:
+Method to Achieve the Objective
+-------------------------------
 
 Make the car to come 'close' to two points, but without specifying the time
 when it should be there. *opty* should find the best time for the car to be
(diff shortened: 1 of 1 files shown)


Peter230655:
Unsure, how these unsuccessful test are related to the simulation I pushed.
Error message says something about ``pause`` not defined. I did not touch the code, only optical changes.

Peter230655:
I am very unclear about this error about ``pause``

examples-gallery/advanced/plot_car_around_pylons.py
```diff
@@ -2,6 +2,16 @@
 Car around Pylons
 =================
 
+Objective
+---------
+
+- Show a way to circumvent the current inability of *opty* to handle  instance
+  constraints at times that are not specified in advance but selected optimally
+  by *opty* itself.
```
moorepants, line 10: Can this be applied to #333 ?

Peter230655: Much better!

moorepants: I explicitly want to handle the non-defferntiability as a multiphase problem. It is how we solved the skateboard ollie paper here: https://link.springer.com/article/10.1007/s12283-023-00448-y

moorepants: The only thing I need for that example to work is to allow the instance constraint to be at a time not specified in advance, which is exactly what you write in the description to this example.

Peter230655: I am afraid, my wording is misleading. What I meant was this:
Say, (x, y) is the location of the body.
Then at present, if you want the body to come close to (x0,y0) during its journey you have to specify a time (which may depend on h) when this should happen. 
All my hack does is to avoid having to do this. It does not allow an instance constraint of the form
x.func(t_arbitrary) - x0
y.func(t_arbitrary) - y0

moorepants: Ok, got it.

Peter230655: > Ok, got it.

What would be a less misleading description?
Examples-gallery must no mislead users.

moorepants: I think it is fine, once you read the rest of the example it is clear.

PR #358, opened 2025-02-13
check that initial guesses are within bounds. Fixes #355

### LINKED ISSUE #355: Throw an error or warning if the initial guess is outside of any bounds for the free variables.
It is easy to set the initial guess in `solve()` to values that are outside of any declared bounds. IPOPT will still run, but we could throw and error (or maybe just a warning).

Peter230655:
A method, called from solve to ensure that the initial guesses are within their respective bounds. As such a conflict is not fatal for opty, a UserWarning is raised. Should take care of issue #355

### DIFF
--- a/examples-gallery/advanced/plot_car_in_garage.py
+++ b/examples-gallery/advanced/plot_car_in_garage.py
@@ -299,17 +299,18 @@ instance_constraints = tuple(xi.subs({t: t0}) - xi_val for xi, xi_val
 
 grenze = 25.0
 delta = np.pi/4.
+epsilon = 1.e-5
 bounds1 = {
         Fb: (-grenze, grenze),
         Tf: (-grenze, grenze),
         # restrict the steering angle to avoid locking
-        qf: (-np.pi/2. + delta, np.pi/2. - delta),
+        qf: (-np.pi/2. + delta - epsilon, np.pi/2. - delta + epsilon),
         # these bounds on x, y help convergence a lot!
         x: (-10, 10),
         y: (0.0, 25),
 }
 
-bounds2 = {py[i]: (0, 100) for i in range(number)}
+bounds2 = {py[i]: (0-epsilon, 100+epsilon) for i in range(number)}
 bounds = {**bounds1, **bounds2}
 
 prob = Problem(
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -3,7 +3,6 @@
 import sys
 from functools import wraps
 import logging
-import warnings
 
 import numpy as np
 import sympy as sm
@@ -205,7 +204,7 @@ class Problem(cyipopt.Problem):
 
         self.obj_value = []
 
-    def solve(self, free, lagrange=[], zl=[], zu=[]):
+    def solve(self, free, respect_bounds=True, lagrange=[], zl=[], zu=[]):
         """Returns the optimal solution and an info dictionary.
 
         Solves the posed optimization problem starting at point x.
@@ -226,6 +225,11 @@ class Problem(cyipopt.Problem):
         zu : array-like, shape(n*N + q*N + r + s, ), optional (default=[])
             Initial values for the multipliers for upper variable bounds (only
             if warm start option is chosen).
+
+        respect_bounds : bool, optional (default=True)
+            If True, the initial guess is checked to ensure that it is within
+            the bounds, and a ValueError is raised if it is not. If False, the
+            initial guess is not checked.
 
         Returns
         -------
@@ -250,7 +254,8 @@ class Problem(cyipopt.Problem):
                 gives the status of the algorithm as a message
 
         """
-        self.bounds_conflict_initial_guess(free)
+        if respect_bounds == True:
+            self.bounds_conflict_initial_guess(free)
         return super().solve(free, lagrange=lagrange, zl=zl, zu=zu)
 
     def bounds_conflict_initial_guess(self, free):
@@ -297,7 +302,7 @@ class Problem(cyipopt.Problem):
             if len(violating_variables) > 0:
                 msg = f'The initial guesses for {violating_variables} are in '+\
                 f'conflict with their bounds.'
-                warnings.warn(msg)
+                raise ValueError(msg)
         else:
             pass
 
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -3,6 +3,7 @@
 import sys
 from functools import wraps
 import logging
+import warnings
 
 import numpy as np
 import sympy as sm
@@ -249,8 +250,56 @@ class Problem(cyipopt.Problem):
                 gives the status of the algorithm as a message
 
         """
-
+        self.bounds_conflict_initial_guess(free)
         return super().solve(free, lagrange=lagrange, zl=zl, zu=zu)
+
+    def bounds_conflict_initial_guess(self, free):
+        """
+        Ascertains that the initial guesses for all variables are within the
+        limits prescribed by their respective bounds. As such possible
+        conflicts are not fatal for opty, only a UserWarning is raised.
+
+        Parameters
+        ----------
+        free : array-like, shape(n*N + q*N + r + s, )
+            Initial guess given to solve.
+
+        """
+        if self.bounds is not None:
+            violating_variables = []
+
+            if self.collocator._variable_duration == True:
+                local_ts = self.collocator.time_interval_symbol
+                if local_ts in self.bounds.keys():
+                    if (free[-1] < self.bounds[local_ts][0]
+                        or free[-1] > self.bounds[local_ts][1]):
+                        violating_variables.append(local_ts)
+
+            symbole = self.collocator.state_symbols + \
+            self.collocator.unknown_input_trajectories
+            for symb in symbole:
+                if symb in self.bounds.keys():
+                    idx = symbole.index(symb)
+                    feld = free[idx*self.collocator.num_collocation_nodes:
+                            (idx+1)*self.collocator.num_collocation_nodes]
+                    if (np.any(feld < self.bounds[symb][0])
+                        or np.any(feld > self.bounds[symb][1])):
+                        violating_variables.append(symb)
+
+            startidx = len(symbole)
+            for symb in self.collocator.unknown_parameters:
+                if symb in self.bounds.keys():
+                    idx = self.collocator.unknown_parameters.index(symb)
+                    if (free[startidx+idx] < self.bounds[symb][0]
+                        or free[startidx+idx] > self.bounds[symb][1]):
+                        violating_variables.append(symb)
+
+            if len(violating_variables) > 0:
+                msg = f'The initial guesses for {violating_variables} are in '+\
+                f'conflict with their bounds.'
+                warnings.warn(msg)
+        else:
+            pass
 
     def _generate_bound_arrays(self):
         lb = -self.INF * np.ones(self.num_free)
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -291,7 +291,7 @@ class Problem(cyipopt.Problem):
                         or np.any(feld > self.bounds[symb][1])):
                         violating_variables.append(symb)
 
-            startidx = len(symbole)
+            startidx = len(symbole) * self.collocator.num_collocation_nodes
             for symb in self.collocator.unknown_parameters:
                 if symb in self.bounds.keys():
                     idx = self.collocator.unknown_parameters.index(symb)
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -204,7 +204,7 @@ class Problem(cyipopt.Problem):
 
         self.obj_value = []
 
-    def solve(self, free, respect_bounds=True, lagrange=[], zl=[], zu=[]):
+    def solve(self, free, lagrange=[], zl=[], zu=[], respect_bounds=True):
         """Returns the optimal solution and an info dictionary.
 
         Solves the posed optimization problem starting at point x.
@@ -254,15 +254,15 @@ class Problem(cyipopt.Problem):
                 gives the status of the algorithm as a message
 
         """
-        if respect_bounds == True:
+        if respect_bounds:
             self.bounds_conflict_initial_guess(free)
         return super().solve(free, lagrange=lagrange, zl=zl, zu=zu)
 
     def bounds_conflict_initial_guess(self, free):
         """
         Ascertains that the initial guesses for all variables are within the
-        limits prescribed by their respective bounds. As such possible
-        conflicts are not fatal for opty, only a UserWarning is raised.
+        limits prescribed by their respective bounds. Raises a ValueError if
+        for any variable the initial guess is outside its bounds.
 
         Parameters
         ----------
--- a/opty/tests/test_direct_collocation.py
+++ b/opty/tests/test_direct_collocation.py
@@ -7,7 +7,7 @@ import sympy as sym
 import sympy.physics.mechanics as mech
 from sympy.physics.mechanics.models import n_link_pendulum_on_cart
 from scipy import sparse
-from pytest import raises
+from pytest import raises, warns
 
 from ..direct_collocation import Problem, ConstraintCollocator
 from ..utils import create_objective_function, sort_sympy, parse_free
@@ -1908,3 +1908,151 @@ def test_attributes_read_only():
 
         with raises(AttributeError):
             setattr(test, XX, 5)
+
+def test_bounds_conflict():
+    """Test to ensure that the initial guesses are within the limits
+    set by their respective bounds """
+
+    x, y, z, ux, uy, uz = mech.dynamicsymbols('x y z ux uy uz')
+    u1, u2, u3 = mech.dynamicsymbols('u1 u2 u3')
+    a1, a2, a3 = sym.symbols('a1 a2 a3')
+    b1, b2, b3 = sym.symbols('b1 b2 b3')
+    t = mech.dynamicsymbols._t
+
+    # just random eoms, no physical meaning.
+    eom = sym.Matrix([
+        -x.diff(t) + ux,
+        -ux.diff(t) + a1 * x + u1 + b1,
+        -y.diff(t) + uy,
+        -uy.diff(t) + a2 * y + u2 + b2,
+        -z.diff(t) + uz,
+        -uz.diff(t) + a3 * z + u3 + b3,
+        ])
+
+    state_symbols = (x, y, z, ux, uy, uz)
+    par_map = {b1: 1.0,
+               b2: 2.0,
+               b3: 3.0
+               }
+    num_nodes = 26
+
+    # A: constant time interval
+    t0, tf = 0.0, 1.0
+    interval_value = tf/(num_nodes - 1)
+
+    def obj(free):
+        Fx = free[2*num_nodes:3*num_nodes]
+        return interval_value*np.sum(Fx**2)
+
+    def obj_grad(free):
+        grad = np.zeros_like(free)
+        l1 = 2*num_nodes
+        l2 = 3*num_nodes
+        grad[l1: l2] = 2.0*free[l1:l2]*interval_value
+        return grad
+
+    instance_constraints = (
+        x.func(t0),
+        ux.func(t0),
+        x.func(tf) - 1.0,
+        ux.func(tf),
+    )
+
+    bounds= {
+        a1: (-1.0, 1.0),
+        a2: (-1.0, -0.5),
+        a3: (-1.0, 1.0),
+
+        u1: (-1.0, 1.0),
+        u2: (-1.0, 1.0),
+        u3: (-1.0, 1.0),
+
+        x: (-1.0, 1.0 ),
+        ux: (-1.0, 1.0),
+        y: (-1.0, 1.0),
+        uy: (-1.0, 1.0),
+        z: (-1.0, 1.0),
+        uz: (-1.0, 1.0),
+    }
+
+    prob = Problem(
+        obj,
+        obj_grad,
+        eom,
+        state_symbols,
+        num_nodes,
+        interval_value,
+        known_parameter_map=par_map,
+        instance_constraints=instance_constraints,
+        bounds=bounds,
+        time_symbol=t,
+        )
+
+    initial_guess = np.zeros(prob.num_free)
+    start_idx = np.random.randint(0, num_nodes - 1)
+    for i in range(9):
+        initial_guess[start_idx + i*num_nodes] = 10.0
+    with warns(UserWarning):
+        _ , _ = prob.solve(initial_guess)
+
+    # B: variable time interval
+    h = sym.symbols('h')
+    interval_value = h
+    t0, tf = 0.0, (num_nodes - 1)*h
+
+    def obj(free):
+        Fx = free[6*num_nodes:9*num_nodes]
+        return free[-1]*np.sum(Fx**2)
+
+    def obj_grad(free):
+        grad = np.zeros_like(free)
+        l1 = 6*num_nodes
+        l2 = 9*num_nodes
+        grad[l1: l2] = 2.0*free[l1:l2]*free[-1]
+        grad[-1] = np.sum(free[l1:l2]**2)
+        return grad
+
+    instance_constraints = (
+        x.func(t0),
+        ux.func(t0),
+        x.func(tf) - 1.0,
+        ux.func(tf),
+    )
+    bounds= {
+        a1: (-1.0, -0.5),
+        a2: (-1.0, 1.0),
+        a3: (-1.0, 1.0),
+
+        u1: (-1.0, 1.0),
+        u2: (-1.0, 1.0),
+        u3: (-1.0, 1.0),
+
+        x: (-1.0, 1.0),
+        ux: (-1.0, 1.0),
+        y: (-1.0, 1.0),
+        uy: (-1.0, 1.0),
+        z: (-1.0, 1.0),
+        uz: (-1.0, 1.0),
+        h: (0.0, 0.4 )
+    }
+
+    prob = Problem(
+        obj,
+        obj_grad,
+        eom,
+        state_symbols,
+        num_nodes,
+        interval_value,
+        known_parameter_map=par_map,
+        instance_constraints=instance_constraints,
+        bounds=bounds,
+        time_symbol=t,
+        )
+
+    initial_guess = np.zeros(prob.num_free)
+    start_idx = np.random.randint(0, num_nodes - 1)
+    for i in range(9):
+        initial_guess[start_idx + i*num_nodes] = -10.0
+    initial_guess[-1] = 0.5
+    with warns(UserWarning):
+        _, _ = prob.solve(initial_guess)
--- a/opty/tests/test_direct_collocation.py
+++ b/opty/tests/test_direct_collocation.py
@@ -7,7 +7,7 @@ import sympy as sym
 import sympy.physics.mechanics as mech
 from sympy.physics.mechanics.models import n_link_pendulum_on_cart
 from scipy import sparse
-from pytest import raises, warns
+from pytest import raises
 
 from ..direct_collocation import Problem, ConstraintCollocator
 from ..utils import create_objective_function, sort_sympy, parse_free
@@ -1909,9 +1909,9 @@ def test_attributes_read_only():
         with raises(AttributeError):
             setattr(test, XX, 5)
 
-def test_bounds_conflict():
-    """Test to ensure that the initial guesses are within the limits
-    set by their respective bounds """
+def test_bounds_conflict_method():
+    """Test to ensure that the method of Problem, bounds_conflict_initial_guess
+    raises a ValueError when the initial guesses violates the bounds."""
 
     x, y, z, ux, uy, uz = mech.dynamicsymbols('x y z ux uy uz')
     u1, u2, u3 = mech.dynamicsymbols('u1 u2 u3')
@@ -1960,7 +1960,7 @@ def test_bounds_conflict():
 
     bounds= {
         a1: (-1.0, 1.0),
-        a2: (-1.0, -0.5),
+        a2: (-1.0, 1.0),
         a3: (-1.0, 1.0),
 
         u1: (-1.0, 1.0),
@@ -1992,8 +1992,8 @@ def test_bounds_conflict():
     start_idx = np.random.randint(0, num_nodes - 1)
     for i in range(9):
         initial_guess[start_idx + i*num_nodes] = 10.0
-    with warns(UserWarning):
-        _ , _ = prob.solve(initial_guess)
+    with raises(ValueError):
+        prob.bounds_conflict_initial_guess(initial_guess)
 
     # B: variable time interval
     h = sym.symbols('h')
@@ -2019,7 +2019,7 @@ def test_bounds_conflict():
         ux.func(tf),
     )
     bounds= {
-        a1: (-1.0, -0.5),
+        a1: (-1.0, 1.0),
         a2: (-1.0, 1.0),
         a3: (-1.0, 1.0),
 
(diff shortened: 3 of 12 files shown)


opty/direct_collocation.py
```diff
@@ -249,9 +250,57 @@ def solve(self, free, lagrange=[], zl=[], zu=[]):
                 gives the status of the algorithm as a message
 
         """
-
+        self.bounds_conflict_initial_guess(free)
```
moorepants, line 253: After some thought, I think I prefer that you have:

```python
        if not allow_out_of_bounds:
            self.bounds_conflict_initial_guess(free)
```

and it shoudl be a kwarg of `.solve()` that is set to false by default. Then an error is raised. The user may have a ligitmate reason to not want this, so they can disable it with `allow_out_of_bounds=True`.

Feel free to suggest a better name for the kwarg and/or use the "not" of it, i.e. `error_out_of_bounds` or something. I haven't thought of a shorter name for the `kwarg`.

There is a tendency to avoid use of warnings because they are easy to ignore or may not display at all depending on the settings Python is run with. So forcing an error is usually the right choice.

opty/tests/test_direct_collocation.py
```diff
@@ -1908,3 +1908,151 @@ def test_attributes_read_only():
...
+        instance_constraints=instance_constraints,
+        bounds=bounds,
+        time_symbol=t,
+        )
+
+    initial_guess = np.zeros(prob.num_free)
+    start_idx = np.random.randint(0, num_nodes - 1)
+    for i in range(9):
+        initial_guess[start_idx + i*num_nodes] = -10.0
+    initial_guess[-1] = 0.5
```
moorepants, line 2056: You need a test that calls the new method explicitly as the primary test, i.e. does your new method do what it is supposed to do. This doesn't not require solving the problem.

Then additionally you need the test that `.solve()` does the right thing with and without the suggested flag above.

opty/direct_collocation.py
Peter230655: > After some thought, I think I prefer that you have:
> 
> ```python
>         if not allow_out_of_bounds:
>             self.bounds_conflict_initial_guess(free)
> ```
> 
> and it shoudl be a kwarg of `.solve()` that is set to false by default. Then an error is raised. The user may have a ligitmate reason to not want this, so they can disable it with `allow_out_of_bounds=True`.
> 
> Feel free to suggest a better name for the kwarg and/or use the "not" of it, i.e. `error_out_of_bounds` or something. I haven't thought of a shorter name for the `kwarg`.
> 
> There is a tendency to avoid use of warnings because they are easy to ignore or may not display at all depending on the settings Python is run with. So forcing an error is usually the right choice.

What about ``respect_bounds``? Default would be True, and an error raised. What error should be raised? ``ValueError``?
Just so I get it right: Some old code, where the bounds are ignored by the initial guess would suddenly get an error. Tihs does not matter?

moorepants: Yes, much better name!

ValueError is appropriate.

>  Just so I get it right: Some old code, where the bounds are ignored by the initial guess would suddenly get an error. This does not matter?

We'll have to update any examples or tests to accommodate this. Either by changing initial guess (probably best idea) or setting `respect_bounds=False`.

opty/tests/test_direct_collocation.py
moorepants: Once you have the new `kwarg` you need to test if `True` and `False` do the right things.

Peter230655:
Adjusted as per the suggestions given.

Peter230655:
If I read it correctly, about 10 examples seem to give problems.
What I could do is this:
- set the kwarg to False
- over the next so many days, I could look at the offending examples, and set their kwarg to False?
Or what would you suggest?

moorepants:
My suggestion was made above:

> We'll have to update any examples or tests to accommodate this. Either by changing initial guess (probably best idea) or setting respect_bounds=False.

moorepants:
Good looking error message!

>     ValueError: The initial guesses for [c, k] are in conflict with their bounds.

Peter230655:
> Good looking error message!
> 
> > ```
> > ValueError: The initial guesses for [c, k] are in conflict with their bounds.
> > ```

Thanks!
I wasn't sure whether these brackets [....] would be o.k. - but frankly I could not remove them.

Peter230655:
> My suggestion was made above:
> 
> > We'll have to update any examples or tests to accommodate this. Either by changing initial guess (probably best idea) or setting respect_bounds=False.

I will go over the examples in the coming days, try to fix them and push them if fixed.

opty/direct_collocation.py
```diff
@@ -204,7 +204,7 @@ def __init__(self, obj, obj_grad, equations_of_motion, state_symbols,
 
         self.obj_value = []
 
-    def solve(self, free, lagrange=[], zl=[], zu=[]):
+    def solve(self, free, respect_bounds=True, lagrange=[], zl=[], zu=[]):
```
moorepants, line 207:
```suggestion
    def solve(self, free,lagrange=[], zl=[], zu=[],  respect_bounds=True):
```

Needs to go at end to not break backwards compatibility.

opty/direct_collocation.py
```diff
@@ -249,9 +254,58 @@ def solve(self, free, lagrange=[], zl=[], zu=[]):
...
-
+        if respect_bounds == True:
+            self.bounds_conflict_initial_guess(free)
         return super().solve(free, lagrange=lagrange, zl=zl, zu=zu)
 
+    def bounds_conflict_initial_guess(self, free):
+        """
+        Ascertains that the initial guesses for all variables are within the
+        limits prescribed by their respective bounds. As such possible
+        conflicts are not fatal for opty, only a UserWarning is raised.
```
moorepants, line 265: Say "raises a ValueError"

opty/direct_collocation.py
Peter230655: > Needs to go at end to not break backwards compatibility.

I did not know this! I thought with kwargs the sequence did not matter.

moorepants: Technically it does because this works: `prob.solve(free, my_lagrange, my_zl, my_zu`. If you don't specify the kwarg name when calling the function/method, it uses the order. Some people may do this.

That's why Python introduced a new syntax, something like `def solve(self, free, *, lagrange=[], ...)` that actually enforces you to name the kwarg on use (I think). But if we add that, it is backwards incompatible.

opty/tests/test_direct_collocation.py
```diff
@@ -1908,3 +1908,243 @@ def test_attributes_read_only():
...
+
+    initial_guess = np.zeros(prob.num_free)
+    start_idx = np.random.randint(0, num_nodes - 1)
+    for i in range(9):
+        initial_guess[start_idx + i*num_nodes] = -10.0
+    initial_guess[-1] = 0.3
+    with raises(ValueError):
+        prob.bounds_conflict_initial_guess(initial_guess)
+
+def test_solve_kwarg_respect_bounds():
```
moorepants, line 2060: My guess is there is no reason to have two test functions. Can this not all be in one?

opty/direct_collocation.py
Peter230655: > Technically it does because this works: `prob.solve(free, my_lagrange, my_zl, my_zu`. If you don't specify the kwarg name when calling the function/method, it uses the order. Some people may do this.
> 
> That's why Python introduced a new syntax, something like `def solve(self, free, *, lagrange=[], ...)` that actually enforces you to name the kwarg on use (I think). But if we add that, it is backwards incompatible.

Thanks! My python knowledge still so limited...

opty/tests/test_direct_collocation.py
Peter230655: > My guess is there is no reason to have two test functions. Can this not all be in one?

Can easily be in one. I thought two were wanted.

moorepants: You have lots of set up code often in these, so it is best to minimize that in tests, especially if it is compiling the C code which always takes some time.

Peter230655: > You have lots of set up code often in these, so it is best to minimize that in tests, especially if it is compiling the C code which always takes some time.

I fully agree. When I looked at the running time, most of it was this 'set up'.
So, while normally different things should have different test function, but if the different things are closely related, like here, one test function is acceptable?

moorepants: I often test many things in one test function, especially if setup is computationally costly.

Peter230655:
I proceeded as follows:
- Where possible, I adjusted the initial guesses to meet the bounds.
-  where solutions of previous runs were used as initial guesses, I opened the bounds slightly to allow for numerical errors in the solutions.
- ``plot_park2004.py``: Here the initial guesses for the gain factors ``k00``.... ``k13`` were very far from their bounds. When I set them to meet the bounds, the results seemed to be a bit worse. So here I set ``recpect_bounds=False``
- ``plot_human_trait.py``. Here the initial guess for the earth solution was some previous solution. Its solution is used as the initial guess for the moon solution. THe earth solution seemed far outside the bounds. So, I used the initial guess for the earth solution also for the moon solution

Peter230655:
This test has been running for nearly 2 1/2 hours.
Is this normal, or did something happen?

Peter230655:
seems to run too long, will start again.

moorepants:
it seems that conda cannot resolve the environment on windows with python 3.10.

```
yeadon                    1.5.0              pyhd8ed1ab_2    conda-forge
zstandard                 0.23.0          py310he5e10e1_1    conda-forge
zstd                      1.5.6                h0ea2cb4_0    conda-forge
CI detected...
Channels:
 - conda-forge
 - defaults
Platform: win-64
Collecting package metadata (repodata.json): ...working... done
Solving environment: ...working... 
Error: The operation was canceled.
```

I don't know why that would be, as it has worked before.

moorepants:
Peter, since you are on windows can you make this edit to your conda env file and see if it works locally when creating the environment:

```
name: opty-dev
channels:
  - conda-forge
dependencies:
  # build
  - setuptools
  # run (includes optional run deps)
  - cyipopt >=1.1.0
  - cython >=0.29.19
  - matplotlib >=3.2.0
  - numpy >=1.19.0
  - python =3.10
  - scipy >=1.5.0
  - sympy >=1.13  # version required for the docs
  # dev
  - coverage
  - ipython
  - pytest
  - pytest-cov
  # docs/examples
  - joblib
  - numpydoc
  - pip <24  # pin required for the gait2d pip install below
  - pydy >=0.5.0  # gait2d
  - pyyaml  # gait2d
  - sphinx
  - sphinx-gallery
  - sphinx-reredirects
  - symmeplot
  - yeadon
  - pip:
    - -e git+https://github.com/csu-hmc/gait2d#egg=gait2d
```

Peter230655:
> Peter, since you are on windows can you make this edit to your conda env file and see if it works locally when creating the environment:
> 
> ```
> name: opty-dev
> channels:
>   - conda-forge
> dependencies:
>   # build
>   - setuptools
>   # run (includes optional run deps)
>   - cyipopt >=1.1.0
>   - cython >=0.29.19
>   - matplotlib >=3.2.0
>   - numpy >=1.19.0
>   - python =3.10
>   - scipy >=1.5.0
>   - sympy >=1.13  # version required for the docs
>   # dev
>   - coverage
>   - ipython
>   - pytest
>   - pytest-cov
>   # docs/examples
>   - joblib
>   - numpydoc
>   - pip <24  # pin required for the gait2d pip install below
>   - pydy >=0.5.0  # gait2d
>   - pyyaml  # gait2d
>   - sphinx
>   - sphinx-gallery
>   - sphinx-reredirects
>   - symmeplot
>   - yeadon
>   - pip:
>     - -e git+https://github.com/csu-hmc/gait2d#egg=gait2d
> ```

So I understand correctly:
- the above is a .yml file.
- I store it somewhere on my PC
- then I create a new environment, called opty-dev, using this .yml file like it is described in opty, 
- OR would I use this .yml file to update my existing opty-dev environment? This I do not know how to do.

Thanks!

moorepants:
You can do either, but I only changed this line `  - python =3.10` so it would install python 3.10.

Peter230655:
> You can do either, but I only changed this line ` - python =3.10` so it would install python 3.10.

So to update I proceed as follows?
1. I take the yml file above and store it on my PC as, say, ``update.yml`` 
2. I open the anaconda promt and put ``activate opty-dev``
3. then I put ``conda env update -f path-to-update.yml``

moorepants:
That will work but only if you first remove the existing opty-dev conda environment.

moorepants:
Sorry, nevermind, if you use "update" then it should ve fine. You can also simple add `python =3.10` to the yml file you already have in the repo.

Peter230655:
> Sorry, nevermind, if you use "update" then it should be fine. You can also simple add `python =3.10` to the yml file you already have in the repo.

This looks easiest for me! Then, when I push, this changed .yml file will also be pushed. Is this o.k.?

Peter230655:
Unsure, if this is related to the issue:
When I run ``plot_sit_to_stand`` (This file had not problems with the bounds so I did not touch it!) I get very poorly met eoms:
image
I also get this error:
image

moorepants:
No you should not push this change to the yml file. We do not want to specify the python version in the yml file stored in this repo.

moorepants:
After seeing how much you have changed to get this to work, I'm second guessing my advice to make `respect_bounds=True` the default. This is a backwards incompatible change and will cause a lot of code to break. Maybe we should make the default be False.

Peter230655:
> After seeing how much you have changed to get this to work, I'm second guessing my advice to make `respect_bounds=True` the default. This is a backwards incompatible change and will cause a lot of code to break. Maybe we should make the default be False.

This backwards incompatability was a slight concern I had with setting ``respect_bounds=True``, so I agree with your idea completely.
Maybe best to close this PR and make a new one where I only change ``direct_collocation.py`` and add the test to ``test_direct_collocation.py``?

moorepants:
I'd try to get out of the habit of closing and opening new prs, you can simply update this one until it is where we like it.

Peter230655:
> I'd try to get out of the habit of closing and opening new prs, you can simply update this one until it is where we like it.

I will do so, within today, I hope.

moorepants:
If you have trouble with conda environments, remember that I wrote this guide: https://mechmotum.github.io/blog/setup-development-environment.html

moorepants:
I opened an issue about the build failure: https://github.com/conda-forge/conda-forge.github.io/issues/2460

Peter230655:
> I opened an issue about the build failure: [conda-forge/conda-forge.github.io#2460](https://github.com/conda-forge/conda-forge.github.io/issues/2460)

I am switching to ``respect_bounds=False`` as default, and change the examples back to their original state. Then I will push again. Let's see.

moorepants:
The Windows + Python 3.10 error is unrelated to your changes, so it will likely continue to fail. If you can try to install the environment on your computer with Python 3.10, that may give us more clues (as you can see the whole error message) or let it run longer.

Peter230655:
> The Windows + Python 3.10 error is unrelated to your changes, so it will likely continue to fail. If you can try to install the environment on your computer with Python 3.10, that may give us more clues (as you can see the whole error message) or let it run longer.

If I install Python 3.10 in my environment opty-dev, and then run it on my PC the error should show, correct?

Peter230655:
I may have found an error:
I switched my opty-dev environment to python 3.10
When I ran the examples I got an error which I did not get before. I had;
``print(f'some text {info['obj_val']} more text')`` This gave an error. Then I switched to
``print(f"some text {info['obj_val']} more text")`` and it ran fine. Apparently an issue with python 3.10

Peter230655:
Since I found an error which seems to happen only with python 3.10 I risk a push again.

moorepants:
The error is that the python 3.10 environment on windows will not install (as far as I can tell). If you successfully installed it locally, that is important info. How long did it take to solve?

Peter230655:
> The error is that the python 3.10 environment on windows will not install (as far as I can tell). If you successfully installed it locally, that is important info. How long did it take to solve?

What I did was this:
- I copied your .yml file to some other directory
- I set python =3.10
- then I updated my opty-dev with this .yml file. 
- when I made conda list it showed python 3.10.xx (I do not recall the xx number)

then I ran to create the html files. This took maybe 5 min.

Peter230655:
> If you have trouble with conda environments, remember that I wrote this guide: https://mechmotum.github.io/blog/setup-development-environment.html

I printed this the first time you sent to me! :-) ...... but I have this tendency to forget. So thanks for telling me again!

Peter230655:
Should I switch my opty-dev to Python 3.13 to see what has happened now?

moorepants:
> This took maybe 5 min to run.

If it took 5 minutes to make the environment, that is close to the 360 second time limit on CI. You have a newer computer, so it may be that the solve is just taking too much time for that. I think the only solution is to narrow the version qualifiers in the yaml file to reduce the possible combinations of software.

Peter230655:
> > This took maybe 5 min to run.
> 
> If it took 5 minutes to make the environment, that is close to the 360 second time limit on CI. You have a newer computer, so it may be that the solve is just taking too much time for that. I think the only solution is to narrow the version qualifiers in the yaml file to reduce the possible combinations of software.

Sorry, I meant 5 min to make the html files. Updating my environment to python 3.10 was maybe one min.

moorepants:
Can you create the environment from scratch, not simply update. It is failing to create the environment from scratch in CI.

moorepants:
This one has been running for 5 hours and 30 minutes and hasn't solved the environment: https://github.com/csu-hmc/opty/actions/runs/13353531780/job/37293136717?pr=361

Peter230655:
> Can you create the environment from scratch, not simply update. It is failing to create the environment from scratch in CI.

NO problem! I will use the .yml file in opty, right?

Peter230655:
I built from scratch as follows:
- I deleted the environment opty-dev
- I created a new one as explained in the opty documentation. 
This took about 90 dec

moorepants:
>  Should I do the same thing but change line 12 there from python to python =3.10 ?

Yes, you have to set python = 3.10 to do the same thing in the failing CI build.

Peter230655:
> > Should I do the same thing but change line 12 there from python to python =3.10 ?
> 
> Yes, you have to set python = 3.10 to do the same thing in the failing CI build.

Clear, will do so. Right now I managed once again to mess up **everything** :-((
Hopefully I'll fix it tomorrow morning, and report about the timing.

moorepants:
If you merge master the tests should run correctly again. I fixed it in #363

Peter230655:
Worked !!   :-)

opty/direct_collocation.py
```diff
@@ -244,9 +249,58 @@ def solve(self, free, lagrange=[], zl=[], zu=[]):
...
+        """
+        Ascertains that the initial guesses for all variables are within the
+        limits prescribed by their respective bounds. Raises a ValueError if
+        for any variable the initial guess is outside its bounds.
+
+        Parameters
+        ----------
+        free : array-like, shape(n*N + q*N + r + s, )
+            Initial guess given to solve.
+
```
moorepants, line 266: You can add a "Raises" section to describe the ValueError. See the numpydoc documentation.

opty/direct_collocation.py
```diff
@@ -244,9 +249,58 @@ def solve(self, free, lagrange=[], zl=[], zu=[]):
...
+            if self.collocator._variable_duration == True:
+                local_ts = self.collocator.time_interval_symbol
+                if local_ts in self.bounds.keys():
+                    if (free[-1] < self.bounds[local_ts][0]
+                        or free[-1] > self.bounds[local_ts][1]):
+                        violating_variables.append(local_ts)
+
+            symbole = self.collocator.state_symbols + \
+            self.collocator.unknown_input_trajectories
```
moorepants, line 279: bit better to do:

```
a = (longgggggggggggggggggggggg +
     longgggggggggggggggggggggggggg)
```

opty/direct_collocation.py
```diff
@@ -244,9 +249,58 @@ def solve(self, free, lagrange=[], zl=[], zu=[]):
...
+            for symb in self.collocator.unknown_parameters:
+                if symb in self.bounds.keys():
+                    idx = self.collocator.unknown_parameters.index(symb)
+                    if (free[startidx+idx] < self.bounds[symb][0]
+                        or free[startidx+idx] > self.bounds[symb][1]):
+                        violating_variables.append(symb)
+
+            if len(violating_variables) > 0:
+                msg = f'The initial guesses for {violating_variables} are in '+\
+                f'conflict with their bounds.'
```
moorepants, line 299: strings don't need `\` if you do this:

```python
msg = (f'sstringgggggggggggggggggggggggg'
       f'anotherstringgggggg')
```

opty/tests/test_direct_collocation.py
```diff
@@ -1907,3 +1907,161 @@ def test_attributes_read_only():
...
+        -uy.diff(t) + a2 * y + u2 + b2,
+        -z.diff(t) + uz,
+        -uz.diff(t) + a3 * z + u3 + b3,
+        ])
+
+    state_symbols = (x, y, z, ux, uy, uz)
+    par_map = {b1: 1.0,
+               b2: 2.0,
+               b3: 3.0
+               }
```
moorepants, line 1937: preferred formating for dictionaries is:

```python
a = {
    'b': 1,
    'c': 2,
}
```

opty/tests/test_direct_collocation.py
```diff
@@ -1907,3 +1907,161 @@ def test_attributes_read_only():
...
+        u2: (-1.0, 1.0),
+        u3: (-1.0, 1.0),
+
+        x: (-1.0, 1.0 ),
+        ux: (-1.0, 1.0),
+        y: (-1.0, 1.0),
+        uy: (-1.0, 1.0),
+        z: (-1.0, 1.0),
+        uz: (-1.0, 1.0),
+    }
```
moorepants, line 1977: This case seems to have all unknown variables with bounds. Do you tests cases where bounds are not given for some (or all)?

opty/tests/test_direct_collocation.py
```diff
@@ -1907,3 +1907,161 @@ def test_attributes_read_only():
...
+
+    instance_constraints = (
+        x.func(t0),
+        ux.func(t0),
+        x.func(tf) - 1.0,
+        ux.func(tf),
+    )
+
+    bounds= {
+        a1: (-1.0, 1.0),
```
moorepants, line 1963: What would happen if this was? `a1: (-1.0, 1.0)`?

Peter230655: My method would not care, about IPOPT do not know, never tried.
Would this not be a completely nonstandard way to give an interval?

opty/tests/test_direct_collocation.py
Peter230655: > This case seems to have all unknown variables with bounds. Do you tests cases where bounds are not given for some (or all)?

Yes, there is a test without any bounds in the test function. For the state variables, I ignore the bounds.

opty/tests/test_direct_collocation.py
moorepants: I'm just trying to imagine what inputs may break your function (and thus should be written as tests).

opty/tests/test_direct_collocation.py
Peter230655: You mean 'b1' instead of b1 or you mena the location of } ?

moorepants: I mean the location of the parentheses and the indentation. Some examples here: https://peps.python.org/pep-0008/#code-lay-out

opty/tests/test_direct_collocation.py
Peter230655: > I'm just trying to imagine what inputs may break your function (and thus should be written as tests).

Sorry, I mixed this up my PR for ``create_linear_initial_guess``
I did not try with this PR. I will check.

opty/direct_collocation.py
Peter230655: > You can add a "Raises" section to describe the ValueError. See the numpydoc documentation.

I assume you meant this section:
image

opty/direct_collocation.py
Peter230655: > bit better to do:
> 
> ```
> a = (longgggggggggggggggggggggg +
>      longgggggggggggggggggggggggggg)
> ```

So, parenthesis are always preferred over \ ?

moorepants: according to pep 8 it is

Peter230655: Thanks!
I would have thought the other way around, I will remember in future.

opty/direct_collocation.py
```diff
@@ -244,9 +249,58 @@ def solve(self, free, lagrange=[], zl=[], zu=[]):
...
+        Parameters
+        ----------
+        free : array-like, shape(n*N + q*N + r + s, )
+            Initial guess given to solve.
+
+        """
+        if self.bounds is not None:
+            violating_variables = []
+
+            if self.collocator._variable_duration == True:
```
moorepants, line 271:
```suggestion
            if self.collocator._variable_duration:
```

opty/tests/test_direct_collocation.py
```diff
@@ -1907,3 +1907,161 @@ def test_attributes_read_only():
...
+    initial_guess[-1] = 0.3
+    with raises(ValueError):
+        prob.bounds_conflict_initial_guess(initial_guess)
+
+    # C: respect_bounds=True: initial guess must be within bounds, else a
+    # ValueError is raised.
+    with raises(ValueError):
+        _, _ = prob.solve(initial_guess, respect_bounds=True)
+
+    # D: respect_bounds=False by default. Bounds are ignored.
```
moorepants, line 2066: you could set max iterations to 1 or 2 to save computation time when solve actually runs.

moorepants:
This looks very good. I left a few minor comments, but they aren't necessary to fix for merging.

The only major comment is that you probably need a few more unit tests to check some variation on what type of bounds were set.

Peter230655:
> This looks very good. I left a few minor comments, but they aren't necessary to fix for merging.
> 
> The only major comment is that you probably need a few more unit tests to check some variation on what type of bounds were set.

Thanks!

I just checked an example (your drone) with 'reversed' bounds: {F: (100, -100)}
It did not break, but did not get a reasonable solution at all.
So, if I detect a reversed bound, should I raise an error? If YES, what error? ValueError?

moorepants:
I would have thought that IPOPT would raise an error. That makes me wonder why reverse bounds is allowed for IPOPT. Maybe there is a reason you'd want that.

For this PR, we only need to ensure that your new functionality is well tested. Do we want to support reverse bounds with this functionality? If not, we should probably raise an error if reverse bounds are detected. A ValueError sounds correct.

Peter230655:
> I would have thought that IPOPT would raise an error. That makes me wonder why reverse bounds is allowed for IPOPT. Maybe there is a reason you'd want that.
> 
> For this PR, we only need to ensure that your new functionality is well tested. Do we want to support reverse bounds with this functionality? If not, we should probably raise an error if reverse bounds are detected. A ValueError sounds correct.

My method will not work for reversed bounds. (I must admit I did not think of this possibility)
So, I will have to check for this possibility.
At least with opty, I have never come across a situation where I would have needed reversed bounds.
So, maybe better for the opty user to get a ValueError, as it likely will not work, even if IPOPT allows it?

moorepants:
>  So, maybe better for the opty user to get a ValueError, as it likely will not work, even if IPOPT allows it?

I'm fine with opty telling you no for this.

Peter230655:
> > So, maybe better for the opty user to get a ValueError, as it likely will not work, even if IPOPT allows it?
> 
> I'm fine with opty telling you no for this.

I will do it like this. Maybe here the numpydocs ``raises`` could be used to explain why the ValueError is raised?

moorepants:
Yes, that is the purpose of that documentation section.

Peter230655:
I now check for reversed bounds before I check whether the initial guess meets the bounds. If there is at least one reversed bound a ValueError will be raised.
I test separately for reversed bounds and for violation of the bounds by the initial guess anf for the working of the kwarg in solve.

moorepants:
What is the status of this PR? Is it ready for a review?

Looks like merge conflicts need to be resolved.

Peter230655:
> What is the status of this PR? Is it ready for a review?
> 
> Looks like merge conflicts need to be resolved.

Yes, ready.
Let me try and fix the conflicts and I will push again.

Peter230655:
Tried to fix the conflict.

moorepants:
There are still outstanding comments from my first review that haven't been addressed.

Peter230655:
- added backend='numpy' to the test function to speed up affairs
- added test if there are no bounds
- simplified the method a little bit
- test that the kwarg in solve works fine

moorepants:
Great, but there are still a number of PEP8 related comments I made above that haven't been addressed.

Peter230655:
> Great, but there are still a number of PEP8 related comments I made above that haven't been addressed.

General question regarding test functions:
With this method, and probably more with linear_initial_guess, there is a very large amount of possible combinations. If you test all of them the test function may get very long and slow - if you can think of all of them.
So better to test as many as possible or better to make some 'reasonable' compromise?

Peter230655:
> > So better to test as many as possible or better to make some 'reasonable' compromise?
> 
> You have to be reasonable, but best to have 100% coverage, i.e. test everything.

Thanks! This is what I tried on both: bounds_conflict and linear_initial_guess. (at least what I could think of)

examples-gallery/advanced/plot_car_in_garage.py
```diff
@@ -318,17 +318,18 @@ def in_0_1(x):
...
         # restrict the steering angle to avoid locking
-        qf: (-np.pi/2. + delta, np.pi/2. - delta),
+        qf: (-np.pi/2. + delta - epsilon, np.pi/2. - delta + epsilon),
         # these bounds on x, y help convergence a lot!
         x: (-10, 10),
         y: (0.0, 25),
 }
 
-bounds2 = {py[i]: (0, 100) for i in range(number)}
+bounds2 = {py[i]: (0-epsilon, 100+epsilon) for i in range(number)}
```
moorepants, line 332:
```suggestion
bounds2 = {py[i]: (0 - epsilon, 100 + epsilon) for i in range(number)}
```

opty/tests/test_direct_collocation.py
```diff
@@ -2019,3 +2019,208 @@ def obj_grad(free):
     solution[-1] = 0.0
     with raises(ValueError):
         time_vector = prob.time_vector(solution, start_time=t0)
+
+def test_bounds_conflict():
```
moorepants, line 2023: I added a `plot=False/True` kwarg to this function so I can check your work. But you have deleted that. Please add it all back and I will review the tests.

moorepants: Sorry, wrong PR.

Peter230655: > I added a `plot=False/True` kwarg to this function so I can check your work. But you have deleted that. Please add it all back and I will review the tests.

Not intentionally. I tried to merge your changes  - apparently not too successfully so.
Will add this right now!

moorepants:
I don't see a test that tests the new kwarg to `.solve()`.

Peter230655:
> I don't see a test that tests the new kwarg to `.solve()`.

image
I thought, this will test the kwarg?

opty/tests/test_direct_collocation.py
```diff
@@ -2019,3 +2019,208 @@ def obj_grad(free):
...
+
+    initial_guess = np.zeros(prob.num_free)
+    with raises(ValueError):
+        prob.bounds_conflict_initial_guess(initial_guess)
+
+
+    # C: respect_bounds=True: initial guess must be within bounds, else a
+    # ValueError is raised.
+    with raises(ValueError):
+        _, _ = prob.solve(initial_guess, respect_bounds=True)
```
moorepants, line 2204: Yes, here. Sorry I missed it.

moorepants:
I don't see a test where you provide valid bounds and show that it doesn't raise an error. You have an example of not passing any bounds and it doesn't raise, but you need to check when passing bounds.

Peter230655:
> I don't see a test where you provide valid bounds and show that it doesn't raise an error. You have an example of not passing any bounds and it doesn't raise, but you need to check when passing bounds.

It is not there. I will add and push

opty/direct_collocation.py
```diff
@@ -244,9 +249,81 @@ def solve(self, free, lagrange=[], zl=[], zu=[]):
                 gives the status of the algorithm as a message
 
         """
-
+        if respect_bounds:
+            self.bounds_conflict_initial_guess(free)
         return super().solve(free, lagrange=lagrange, zl=zl, zu=zu)
 
+    def bounds_conflict_initial_guess(self, free):
```
moorepants, line 256:
```suggestion
    def check_bounds_conflict(self, free):
```

I suggest this name. It doesn't have to be an initial guess. The action is to check it and raise if there is a conflict.

Peter230655: > I suggest this name. It doesn't have to be an initial guess. The action is to check it and raise if there is a conflict.

Definitely better! already changed.

moorepants:
Ok, this looks good now. I will merge if tests pass. Thanks!

Peter230655:
> Ok, this looks good now. I will merge if tests pass. Thanks!

Thank YOU for your patience!

PR #361, opened 2025-02-16
Bump human gait speed to 3 m/s, as the solutions are a bit more fun.

moorepants:


moorepants:
Read the Docs build failed with "Command killed due to timeout or excessive memory consumption".

PR #365, opened 2025-02-18
method to generate an initial guess. Fixes #354

### LINKED ISSUE #354: It would be helpful to have a function that generates an initial guess that is linear between all instance constraints (and zero otherwise)

Peter230655:
Is it o.k. to be a method, or should I make it a property?

### DIFF
--- a/opty/tests/test_direct_collocation.py
+++ b/opty/tests/test_direct_collocation.py
@@ -3339,3 +3339,5 @@ def test_linear_initial_guess(plot=False):
         ax = prob.plot_trajectories(initial_guess)
         ax[0].set_title(('State Trajectories B5: state instances in instance'
                          ' constraints'))
+        import matplotlib.pyplot as plt
+        plt.show()


opty/direct_collocation.py
```diff
@@ -686,6 +685,202 @@ def parse_free(self, free):
 
         return parse_free(free, n, q, N, variable_duration)
 
+    def create_initial_guess(self):
```
moorepants, line 688: I think we should call this `create_linear_initial_guess()` or something that is more descriptive. In the future, we may have `create_random_initial_guess` or `create_zero_initial_guess`, etc.

opty/direct_collocation.py
```diff
@@ -686,6 +685,202 @@ def parse_free(self, free):
...
+        """Creates an initial guess for the free variables in the optimization.
```
moorepants, line 689:
```suggestion
        """Returns an initial guess that is the linear interpolation between exact instance constraints otherwise zero."""
```

opty/direct_collocation.py
```diff
@@ -686,6 +685,202 @@ def parse_free(self, free):
...
+
+        Returns:
```
moorepants, line 691:
```suggestion
        Returns
```

opty/direct_collocation.py
```diff
@@ -686,6 +685,202 @@ def parse_free(self, free):
...
+        --------
+        initial_guess : (n*N + q*N + r + s)-ndarray
+            The initial guess for the free variables in the optimization problem.
+
+        Strategy:
```
moorepants, line 696: This is a non-standard heading for numpydoc. It could be "Notes".

opty/direct_collocation.py
```diff
@@ -686,6 +685,202 @@ def parse_free(self, free):
...
+        - If bounds exist for unknown input trajectories or unknown parameters,
+          the values are set to the midpoint of the interval of the bounds.
+        - If bounds exist for the variable time interval, the value is set to
+          the midpoint of its bounds.
+        - If one of the range limit of a bound is ``-np.inf`` or ``np.inf``,
+          the value is set to the other finite limit of bound.
+        - All else is set to zero.
+
+    """
+        if self.collocator.instance_constraints is None:
```
moorepants, line 711: indentation seems off. Or maybe indentation of the docstring is off.

opty/direct_collocation.py
Peter230655: > I think we should call this `create_linear_initial_guess()` or something that is more descriptive. In the future, we may have `create_random_initial_guess` or `create_zero_initial_guess`, etc.

I will change the name.

opty/direct_collocation.py
Peter230655: > This is a non-standard heading for numpydoc. It could be "Notes".

I will change it.

opty/direct_collocation.py
Peter230655: This would not be totally correct: if, say, there are bounds ``{u: (a, b)}`` for an unknown trajectory u, my initial guess is ``(a + b)/2``, unless, say, ``b = np.inf``, then the initial guess would be ``a``

opty/direct_collocation.py
Peter230655: > indentation seems off. Or maybe indentation of the docstring is off.

corrected

opty/direct_collocation.py
moorepants: The first line should say exactly what it returns (as succinctly as possible).

Peter230655: > The first line should say exactly what it returns (as succinctly as possible).

It is described right below in Notes (before I called it Strategy).

moorepants: But the first line of the docstring should be a succinct description of what it does. Yes it creates an initial guess, but what kind of initial guess does it create?

Peter230655: > > But the first line of the docstring should be a succinct description of what it does. Yes it creates an initial guess, but what kind of initial guess does it create?
> 
> I changed it like this: `Creates an initial guess that is the linear interpolation between exact instance constraints. Please see the notes for more information.`
Better?

moorepants:
Is there a good choice of one of the existing examples to add this to? You can show how it is used to the user.

Peter230655:
> Is there a good choice of one of the existing examples to add this to? You can show how it is used to the user.

You mean in examples-gallery?
If yes, surely I will find an example.
Would you have to merge this first, or would I include the example I find in this PR #365?

moorepants:
You can include the change to the example here.

opty/direct_collocation.py
```diff
@@ -686,6 +685,203 @@ def parse_free(self, free):
 
         return parse_free(free, n, q, N, variable_duration)
 
+    def create_linear_initial_guess(self):
+        """Creates an initial guess that is the linear interpolation between
+        exact instance constraints. Please see the notes for more information.
+
+        Returns
+        -------
+        initial_guess : (n*N + q*N + r + s)-ndarray
```
moorepants, line 694:
```suggestion
        initial_guess : ndarray, shape(n*N + q*N + r + s,)
```

opty/tests/test_direct_collocation.py
```diff
@@ -1907,3 +1907,405 @@ def test_attributes_read_only():
...
+    # u2 - guess
+    expected_guess[5*num_nodes:6*num_nodes] = 9.0/2.0
+    # a1 - guess
+    expected_guess[6*num_nodes] = 9/2
+    # a2 - guess
+    expected_guess[6*num_nodes+1] = 11.0/2.0
+
+
+    initial_guess = prob.create_linear_initial_guess()
+    np.testing.assert_allclose(initial_guess, expected_guess)
```
moorepants, line 2038: Plot of first guess:

image

opty/tests/test_direct_collocation.py
```diff
@@ -1907,3 +1907,405 @@ def test_attributes_read_only():
...
+        a2: (-1.0, 12.0),
+
+        u1: (-10.0, 1.0),
+        u2: (-1.0, 10.0),
+
+        x: (-5.0, 5.0 ),
+        ux: (-1.0, 1.0),
+        y: (-4.0, 4.0),
+        uy: (-1.0, 1.0),
+    }
```
moorepants, line 1976: Easier to compare if placed here:

image

opty/tests/test_direct_collocation.py
```diff
@@ -1907,3 +1907,405 @@ def test_attributes_read_only():
...
+
+    instance_constraints = (
+        x.func(t0) - 3.0,
+        y.func(t0) - 2.0,
+        ux.func(t0) - 5.0,
+
+        x.func(t1) - 1.5*b1,
+        y.func(t1) - 4.0,
+        ux.func(t1) - 1.0+b2,
+        uy.func(t1) - 4.5,
```
moorepants, line 1956: In this case I would expect uy to go from 0.0 to 4.5 to 0.0., but you set it to 4.5 to 4.5 to 4.5.

moorepants: i.e. if not specified, the guess should be at zero

moorepants: but maybe this more a design question. if there is only one instance constraint, what assumption to we make about any other values of time?

opty/tests/test_direct_collocation.py
```diff
@@ -1907,3 +1907,405 @@ def test_attributes_read_only():
...
+    # a2 - guess
+    expected_guess[6*num_nodes+1] = 11.0/2.0
+
+
+    initial_guess = prob.create_linear_initial_guess()
+    np.testing.assert_allclose(initial_guess, expected_guess)
+
+    # A2: np.inf, -np.inf in bounds
+    bounds[a1] = (-np.inf, 10.0)
+    bounds[a2] = (-10.0, np.inf)
```
moorepants, line 2042: a1 and a2 are parameters, so it wouldn't affect the initial guess, right? i get same guess as first test:

image

opty/tests/test_direct_collocation.py
```diff
@@ -1907,3 +1907,405 @@ def test_attributes_read_only():
...
+        y.func(t1) - 4.0,
+        ux.func(t1) - 1.0+b2,
+        uy.func(t1) - 4.5,
+
+        x.func(t2) + 2.0,
+        y.func(t2) - 3.0,
+
+        x.func(tf) + 1.0,
+        ux.func(tf) + 5.0,
+    )
```
moorepants, line 2100: image

opty/tests/test_direct_collocation.py
```diff
@@ -1907,3 +1907,405 @@ def test_attributes_read_only():
...
+
+        u1: (-10.0, 1.0),
+        u2: (-1.0, 10.0),
+
+        x: (-5.0, 5.0 ),
+        ux: (-1.0, 1.0),
+        y: (-4.0, 4.0),
+        uy: (-1.0, 1.0),
+        h: (1.0, 2.0),
+    }
```
moorepants, line 2165: image

opty/tests/test_direct_collocation.py
```diff
@@ -1907,3 +1907,405 @@ def test_attributes_read_only():
...
+    expected_guess[6*num_nodes+1] = 11.0/2.0
+    # h - guess
+    expected_guess[-1] = 1.5
+
+    initial_guess = prob.create_linear_initial_guess()
+    np.testing.assert_allclose(initial_guess, expected_guess)
+
+    # B2: np.inf, -np.inf in bounds
+    bounds[a1] = (-np.inf, 10.0)
+    bounds[a2] = (-10.0, np.inf)
```
moorepants, line 2234: image

opty/tests/test_direct_collocation.py
```diff
@@ -1907,3 +1907,405 @@ def test_attributes_read_only():
...
+        time_symbol=t,
+    )
+
+    expected_guess[6*num_nodes] = 10.0
+    expected_guess[6*num_nodes+1] = -10.0
+
+    initial_guess = prob.create_linear_initial_guess()
+    np.testing.assert_allclose(initial_guess, expected_guess)
+
+    # B3: no bounds
```
moorepants, line 2255: image

opty/tests/test_direct_collocation.py
```diff
@@ -1907,3 +1907,405 @@ def test_attributes_read_only():
...
+        y.func(t10) - 4.0,
+        ux.func(t10) - 1.0+b2,
+        uy.func(t10) - 4.5,
+
+        x.func(t20) + 2.0,
+        y.func(t20) - 3.0,
+
+        x.func(tf0) + 1.0,
+        ux.func(tf0) + 5.0,
+    )
```
moorepants, line 2292: image

moorepants:
I added a commit, so pull this branch before you edit more. You can now do:

```
In [1]: from opty.tests.test_direct_collocation import test_linear_initial_guess

In [2]: test_linear_initial_guess(plot=True)

In [3]: import matplotlib.pyplot as plt

In [4]: plt.show()
```

to quickly plot your guesses.

opty/tests/test_direct_collocation.py
Peter230655: > but maybe this more a design question. if there is only one instance constraint, what assumption to we make about any other values of time?

My assumption was that if there is only one instance constraint for a state variable it will have this value at all times. I thought this was the most reasonable one.

opty/tests/test_direct_collocation.py
Peter230655: a1, a2 are unknown parameters. I set their initial value like this: if there is a bound {a1: (s, t)} I set it at (s + t) /2.
If one of s or t is np.inf I set it to the other (finite) value.
I did not consider bounds which have np.inf on both ends, as they do not make sense in my opinion.

Parameters do not show in the plots.

opty/tests/test_direct_collocation.py
Peter230655: These 'peak' plots happen like this If there is no bound for h, h is assumed to be zero.
So, all times ti = integer * h = 0.
The plots for u1, u2 should have a dot at (0 | 0).

moorepants: I guess h should never be zero, right? So maybe we should always assume a value for it...but what value?

Peter230655: > I guess h should never be zero, right? So maybe we should always assume a value for it...but what value?

I have found, that one should **always** bound h to be non-negative, to avoid senseless 'solutions'.
If there is a bound, say, {h: (0, 0.5)} I set h = 0.25.
So, maybe this would not be a bad default value?

moorepants: Without the user giving some idea of what the total duration of the simulation might be, it's hard (or impossible) to pick a starting number. If someone is simulating molecular dynamics, maybe their duration is a tenth of a second. I guess we could default to some very small value of h. Or this function could take an extract kwarg: `expected_duration=10.0` where the user can give us a bit more info.

moorepants: The expected duration could default to 1.0 second (just as a simple number).

moorepants:
All your do is `git pull origin linear-initial-guess`. Then you are fine to continue.

Peter230655:
> All your do is `git pull origin linear-initial-guess`. Then you are fine to continue.

Thanks!
I would do this when I am in the correct local branch, that is in create-linear-guess ?

moorepants:
The branch associated with this PR is named `linear-initial-guess`:

image

opty/tests/test_direct_collocation.py
Peter230655: > The expected duration could default to 1.0 second (just as a simple number).

I will do so.

moorepants: So you like the idea of the kwarg to give a clue on duration?

Peter230655: > So you like the idea of the kwarg to give a clue on duration?

I like it! I did not think of it, as you normally do not like them.
Default would be false?

moorepants: I don't dislike kwargs.

Default would be `1.0`.

Peter230655:
Just found a MAJOR plunder! Have to go back to the drawing board.

moorepants:
What is the status of this PR? Is it ready for a review?

Peter230655:
> What is the status of this PR? Is it ready for a review?

Yes, I think, it is ready. I tried to test all permutations I could think of.

opty/direct_collocation.py
```diff
@@ -686,6 +685,238 @@ def parse_free(self, free):
 
         return parse_free(free, n, q, N, variable_duration)
 
+    def create_linear_initial_guess(self, plot=False):
```
moorepants, line 688: I'm not a fan of this plot kwarg. You can very easily do:

```
guess = problem.create_linear_initial_guess()
problem.plot_trajectories(guess)
```

It is best not to mix calculation methods with plotting methods.

Peter230655: > I'm not a fan of this plot kwarg. You can very easily do:
> 
> ```
> guess = problem.create_linear_initial_guess()
> problem.plot_trajectories(guess)
> ```
> 
> It is best not to mix calculation methods with plotting methods.

I agree, will correct it.

Peter230655:
- removed the plotting capability.
- added optional kwarg ``end_time``. If given, variable_node_time_interval=1.0/(num_nodes-1) unless bounds for it are given.
- expanded the test to include:
  - no bounds, no instance constraints
  - no bounds
  - different end_times given.

opty/tests/test_direct_collocation.py
```diff
@@ -1923,3 +1923,526 @@ def test_attributes_read_only():
 
         with raises(AttributeError):
             setattr(test, XX, 5)
+
+def test_linear_initial_guess(plot=False):
```
moorepants, line 1927: I added the plot kwarg here to be able to show all the plots for this test but you seem to have deleted all the plot commands, so now my kwarg does nothing.

moorepants: https://github.com/csu-hmc/opty/pull/365/commits/1a17afb66b5025b8244d8b93a463fc95d75dfdd6

Peter230655: As I recall, you had the line
```
if plot:
   prob.plot_trajectories(solution)
```
only once, at the very end of the test_linear_initial_guess, and I left it there.

moorepants: I added it so we could easily plot all the test cases.

Peter230655:
My local branch was terribly out of sync, but I think I managed to restore it - hopefully anyway.
I added the kwarg plot in test_linear_initial_guess, if set to True, every initial guess will be plotted.

Peter230655:
- I added the plotting feature in the test functions. The title of each plots says, which test it belongs to. 
- If there is a conflict between bounds for the variable time interval and the end_time, bounds have precedence. I added this in the docstrings.

moorepants:
I don't think the instance constraints is an accident in any way. Instance constraints are equations, just like the equations that make up the constraints generated from the equations of motion. They can be arbitrary equations, so what would you imagine a dictionary key being?

Peter230655:
> I don't think the instance constraints is an accident in any way. Instance constraints are equations, just like the equations that make up the constraints generated from the equations of motion. They can be arbitrary equations, so what would you imagine a dictionary key being?

I thought like: instance_constraint = {x(t0): -5.0} instead of (x8t0) - 5.0)
Not important, I have to learn how to unravel these sympy terms to some extend anyway.

moorepants:
You can do such a thing but what if you have `x(t0) - y(t0) + z(tf) - complex_function_of(parameters`. What is the dictionary key?

Peter230655:
It seems, that with initial_guess = np.zeros(prob.num_nodes) I get warings about zero divides. It keeps on running, but looks 'ugly'
With initial_guess = np.ones(prob.num_nodes) * 0.01 no warnings.
_Should I change to this in my method?_

moorepants:
This needs to be updated with master. @Peter230655 what is the status on this PR?

Peter230655:
I will update it.

Status: 
I wanted to add a version with would 'superimpose' random fluctuations on the linear initial guesses.
You said, this might be a good idea, but I should wait until you had reviewed this one.

moorepants:
If you want this one in, please update and make sure all is in order. Let me know when you want to review it. I suggest adding random fluctuations in a later PR.

opty/tests/test_direct_collocation.py
moorepants: This needs to be resolved, as you deleted my plotting code.

Peter230655:
YES, separate PR for random fluctuations. I di not express myself correctly.

Peter230655:
Just merged with new master.
(When I have to manually resolve conflicts, I tend to destroy about 50%. But it seems to have no conflict)
I should be in order, I believe the tests are pretty exhaustive.

moorepants:
Did you restore the plotting I had in the tests? I've asked about that above several times.

Peter230655:
Yes, of course! I did it immediately after I inadevertently da erased this and you told me so.

moorepants:
Incorrect formatting:
image

Also, plot the initial guess.

moorepants:
The initial guess is not plotted in the sliding block example either.

Peter230655:
No idea what the error was. Removed the plotting as this is the only thing I think I changed.

Peter230655:
@moorepants : Never complain about my German words again! :-)   :-)  !
When I added the plottoing I put
   ```
ax = prob.plot_trajectories(initial_guess)
ax[0].set_title('trajectories of initial guess')       or something similar
```
I got some strange error.

I removed the plotting and it worked
I added the plotting again, but insted of ``ax`` I used ``achsen``, the German word for axes.
Now it works!

**Comment by moorepants** [comment]
In the docs build the solution is not optimal:

```
b'Algorithm converged to a point of local infeasibility. Problem may be infeasible.'
```

Peter230655:
I will surely find an example which will work. All I wanted to say was that it works for fewer examples than I had hoped for.

PR #368, opened 2025-02-19
Try testing against intel-openmp on Windows.

moorepants:


moorepants:
intel-openmp installs and the warning for not being able to compile with the openmp flag did not show up but there was no openmp performance increase:

```
Run python examples/parallel_example.py
  python examples/parallel_example.py
  shell: C:\Program Files\Git\bin\bash.EXE -el {0}
  env:
    INPUT_RUN_POST: true
    CONDA: C:\Miniconda
  
test.c
Time for non-parallel: 11.773036900000001
Time for parallelized: 11.706586399999999
```

PR #370, opened 2025-02-20
better docstring

Peter230655:
I tried to make the docstrings confirm to format given in #348

NB:
It was a bit difficult here to get a solution, see the ``Note`` in the docstrings.
Therefore I did not try

fname = f'.....'
if os.path.exists(fname):
      solution = np.loadtxt(fname)

else:
       inital_guess = .....


moorepants:
The CI builds failed on master with "The job was not started because recent account payments have failed or your spending limit needs to be increased. Please check the 'Billing & plans' section in your settings."

I was not aware there were any limits for open source projects. That is news to me.

PR #373, opened 2025-02-21
make docstrings as per #348, skip solve if solution is available

Peter230655:
Tried to make docstrings as per #348
Skip prob.solve(...) if solution is available. This hardly saved 1 sec, as convergence with a good initial guess is very fast
Put storage of the solution in an .csv file, before it was in a .npy file. 

I made a PR, as you said, the problem may be gone.


moorepants:
It's possible that we should be focusing on the animation generation times. Donalth Knuth is quoted as saying:

""The real problem is that programmers have spent far too much time worrying about efficiency in the wrong places and at the wrong times; premature optimization is the root of all evil (or at least most of it) in programming. "

We should make the documentation consistent (as one goal) and we should reduce run times but do so in a way where we target the worst offenses first.

Peter230655:
> It's possible that we should be focusing on the animation generation times. Donalth Knuth is quoted as saying:
> 
> ""The real problem is that programmers have spent far too much time worrying about efficiency in the wrong places and at the wrong times; premature optimization is the root of all evil (or at least most of it) in programming. "
> 
> We should make the documentation consistent (as one goal) and we should reduce run times but do so in a way where we target the worst offenses first.

Makes sense!
I will keep this in mind when I go over the simulations.

PR #374, opened 2025-02-22
Reduce combination of os and python versions tested in CI.

moorepants:
Changes to testing all supported python versions on ubuntu and only the latest two python versions on the other OS.


moorepants:
This is a memory consumption exceeded on readthedocs error.

PR #377, opened 2025-02-22
adjusted docstrings as per #348, reduced execution time

Peter230655:
While the program was designerd to run both versions, fixed time interval and variable interval, only the animation for variable interval was shown. I commented out the fixed time interval portions. Reduced execution time on my PC from around 53 sec to around 23 sec.


examples-gallery/beginner/plot_sliding_block.py
```diff
@@ -273,41 +294,36 @@ def update(frame):
...
-# %%
 # Now the results of **minimized energy** are shown.
-selection = 1
-print('Message from optimizer:', info_list[selection]['status_msg'])
+#selection = 1
+#print('Message from optimizer:', info_list[selection]['status_msg'])
 
-# %%
-prob_list[selection].plot_objective_value()
+#_ = prob_list[selection].plot_objective_value()
```
moorepants, line 316: Don't comment out lines in examples, best to remove them vs. commenting.

Peter230655: > Don't comment out lines in examples, best to remove them vs. commenting.

I thought like this a potential user could easily return to running both. I put in the Notes why I did it.
Still better to delete

moorepants: I think it is a bit bad form. If you want to tell them some kind of option that will not actually be run in our example do this in a comment block as rst:

```py
# %%
# If you want to check ... they use code like::
# 
#    prob_list[selection].plot_objective_value()
```

This will show up as a monospaced text they can easily copy and paste into their script but will not be run in the example.

Peter230655:
does this look better?
(Actually now I see, why it is better than commented out)

Peter230655:
This simulation seems to be free of the ``f string error``

examples-gallery/beginner/plot_sliding_block.py
```diff
@@ -273,44 +298,40 @@ def update(frame):
...
+#
+# _ = prob_list[selection].plot_objective_value()
+#
+# _ = prob_list[selection].plot_constraint_violations(solution_list[selection])
+#
+# _ = prob_list[selection].plot_trajectories(solution_list[selection])
+#
+# fig, ax = plt.subplots(figsize=(8, 8))
+#
+# anim, _ = drucken(selection, fig, ax)
```
moorepants, line 332: I don't think this will render as you expect. To do code blocks in RST format it looks like:

```rst
Type this code::

   a, b = 1, 2
   print(a + b)
```

Peter230655: I played around with it for a long time. Then, finally, on the ``make html`` rendering it looked o.k.
I ran out of ideas.
I start like this:
``# %%``
and the added the text.


> I don't think this will render as you expect. To do code blocks in RST format it looks like:
> 
> ```
> Type this code::
> 
>    a, b = 1, 2
>    print(a + b)
> ```

Does this mean I have to typle like
```
code
```

examples-gallery/beginner/plot_sliding_block.py
```diff
@@ -105,7 +124,13 @@ def strasse(x, a, b):
 
 # %%
 # Set up the optimization problems and solve them.
-for selection in (0, 1):
+# if you want to run both optimizations, replace the two lines below with this
+# line:
```
moorepants, line 128:
```suggestion
# line::
```

examples-gallery/beginner/plot_sliding_block.py
```diff
@@ -105,7 +124,13 @@ def strasse(x, a, b):
...
+#
+# for selection in [0, 1]:
```
moorepants, line 130:
```suggestion
#    for selection in [0, 1]:
```

moorepants: Those are the two fixes for this to render correct (I think).

examples-gallery/beginner/plot_sliding_block.py
```diff
@@ -273,44 +298,40 @@ def update(frame):
...
-# %%
-# Plot the trajectories of the block.
-prob_list[selection].plot_trajectories(solution_list[selection])
-
-# %%
-# Animate the solution.
-fig, ax = plt.subplots(figsize=(8, 8))
-anim, _ = drucken(selection, fig, ax)
+# If you want to run the solution with a fixed time interval, you should add the
+# following code to the code here:
```
moorepants, line 318:
```suggestion
# following code to the code here::
```

examples-gallery/beginner/plot_sliding_block.py
moorepants: All the code lines should be indented relative to the text line above it that ends in a `::`.

https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks

examples-gallery/beginner/plot_sliding_block.py
Peter230655: Thanks!!
I will be out of station for two days, will try on Tuesday.

moorepants: There is no rush on anything.

examples-gallery/beginner/plot_sliding_block.py
```diff
@@ -3,6 +3,16 @@
 Block Sliding Over a Hill
 =========================
 
+Objective
+---------
+
+- a simplex example to show how to use ``opty's`` capability of variabel node
```
moorepants, line 9:
```suggestion
- a simple example to show how to use ``opty's`` capability of variabel node
```

Peter230655:
using ``backend='numpy' `` reduced running time again from about 27 sec to 16 sec on my PC.

Peter230655:
Error said unexpected indentation in line 309, but there was on indentation that I could see.

Peter230655:
@moorepants I am at a loss with the error message:

my code looks like this:
image

Peter230655:
There is an error ``unexpected indentation`` which I do not understand. So all I can do is try different things and keep on pushing.

Peter230655:
I am at a loss with this error of ``unexpected indetion``
When I look at ``docs/readthedocs.org:opty`` it seems to build fine...

moorepants:
It is a RestructuredText formatting error.

Peter230655:
> It is a RestructuredText formatting error.

So, not my error?

moorepants:
Yes, it is your formatting error. Anything in a `# %%` "cell" is processed by RestructureText.

Peter230655:
> Yes, it is your formatting error. Anything in a `# %%` "cell" is processed by RestructureText.

it objects to line 309:
image

moorepants:
It isn't line 309 in the py file it is line 309 in the generated rst file: `plot_sliding_block.rst`

Peter230655:
> It isn't line 309 in the py file it is line 309 in the generated rst file: `plot_sliding_block.rst`

Thanks! At least now I got in the area of the problem.

Peter230655:
It objects to worng indentation in line 309 of the rst file.
I found the .rst file in opty/docs/example/beginners on my PC
It looks like this:
image

moorepants:
The formatting rules for RestructuredText literal blocks is here: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks

Peter230655:
As this example now has a 'gif', this PR would be a wrong step. I will close it.

PR #380, opened 2025-02-23
Bicycle countersteer example

moorepants:
This example shows how to deal with kinematic input (instead of a force or torque) where u is the input, but u' and u'' may show up in the equations also and you need to ensure they are all consistent.

Uses model from: https://plotly.com/python/v3/ipython-notebooks/bicycle-control-design/

### DIFF
--- a/examples-gallery/intermediate/plot_countersteer.py
+++ b/examples-gallery/intermediate/plot_countersteer.py
@@ -209,14 +209,14 @@ def bicycle_points(x):
 
         rear_contact = np.array([x, y, 0.0])
         com_on_ground = rear_contact + np.array([par_map[a]*np.cos(psi),
-                                                par_map[a]*np.sin(psi),
-                                                0.0])
+                                                 par_map[a]*np.sin(psi),
+                                                 0.0])
         com = com_on_ground + np.array([par_map[h]*np.sin(theta)*np.sin(psi),
                                         par_map[h]*np.sin(theta)*np.cos(psi),
-                                        par_map[h]*np.cos(theta)])
+                                        -par_map[h]*np.cos(theta)])
         front_contact = rear_contact + np.array([par_map[b]*np.cos(psi),
-                                                par_map[b]*np.sin(psi),
-                                                0.0])
+                                                 par_map[b]*np.sin(psi),
+                                                 0.0])
         front_steer = front_contact + np.array([0.2*np.cos(delta + psi),
                                                 0.2*np.sin(delta + psi),
                                                 0.0])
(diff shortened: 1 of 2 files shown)


Peter230655:
As I understood, the distinction between ``state variables`` and ``unknown_input_trajectories`` is a bit a matter of expediency.
- you need exactly as many state variable as you have eoms.
- certain things, e.g. ''instance_constraints`` can only be done with state variables

I had a very simple simulation a long tine ago, I think, from Betts' book where I had an additional eom (I forgot the details). I simply declared one of the inputs to be a state variable - and all worked fine.

moorepants:
This is a similar example and may show some weakness of how we describe the states and inputs.

moorepants:
@Peter230655 this is now ready for review if you want to comment on it. Click "Files Changed" near the top of this page and then you can comment line-by-line.

examples-gallery/intermediate/plot_countersteer.py
```diff
@@ -0,0 +1,279 @@
...
+end = (num_nodes - 1)*dt
+
+instance_constraints = (
+    # upright, no motion at t = 0
+    theta.func(0*dt),
+    thetadot.func(0*dt),
+    x.func(0*dt),
+    y.func(0*dt),
+    psi.func(0*dt),
+    delta.func(0*dt),
```
moorepants, line 121: Interestingly I had `0*h` here before (instead of `dt`) but the problem still solved.

examples-gallery/intermediate/plot_countersteer.py
```diff
@@ -0,0 +1,279 @@
...
+from opty.utils import MathJaxRepr
+import matplotlib.pyplot as plt
+import matplotlib.animation as animation
+import numpy as np
+import sympy as sm
+import sympy.physics.mechanics as me
+
+# %%
+# Specify the equations of motion
+# -------------------------------
```
moorepants, line 35: A figure would be helpful in this section.

examples-gallery/intermediate/plot_countersteer.py
```diff
@@ -0,0 +1,290 @@
...
+    theta.diff(t) - thetadot,
+    (I1 + m*h**2)*thetadot.diff(t) +
+    (I3 - I2 - m*h**2)*(v*sm.tan(delta)/b)**2*sm.sin(theta)*sm.cos(theta) -
+    m*g*h*sm.sin(theta) +
+    m*h*sm.cos(theta)*(a*v/b/sm.cos(delta)**2*deltadot +
+                       v**2/v*sm.tan(delta)),
+    x.diff(t) - v*sm.cos(psi),
+    y.diff(t) - v*sm.sin(psi),
+    psi.diff(t) - v/b*sm.tan(delta),
+    delta.diff(t) - deltadot,
```
moorepants, line 83: this will display as $\dot{\delta} - \dot{\delta}$, not clear mathematically, should probably introduce other variables.

moorepants:
This should be good, we can tweak details later if needed. Merging.

PR #382, opened 2025-02-23
fixed f string mistake

Peter230655:


Peter230655:
Now I see that this was fixed already by JM. So I close it.

PR #385, opened 2025-02-26
Double pendulum

Peter230655:
- stored solution in .csv file. Needed accurtacy of ``fmt='%.12f' `` for good results
- backend = 'numpy' reduced running time from around 24 sec to around 8 sec
- docstrings as per #348 

I wonder whether the thumbnail shows.


moorepants:
That fixed it!

Peter230655:
> That fixed it!

Thanks!

PR #387, opened 2025-02-27
docstrings as per #348

Peter230655:
Docstrings as per #348
In this case ``backend = 'numpy' `` slowed down forming the problem. As the eoms have over 300,000 operations, maybe to be expected?
Could not reduce fps in the animation without making the video 'jumpy'.
Added solution, but this did not really help the speed much, at least not on my PC


Peter230655:
For some reason the thumbnail does not spin anymore?
I changed nothing in that area.

moorepants:
>  In this case backend = 'numpy' slowed down forming the problem. As the eoms have over 300,000 operations, maybe to be expected?

Yes, to be expected.

moorepants:
This PR broke the animation thumbnail which causes it not to show in the gallery and not to show in the README (and thus front page of the documentation).

@Peter230655

PR #389, opened 2025-02-27
added solution, reduced frames in animation

Peter230655:
By adding a solution and by reducing the number of frames in the animation, on my PC running time decreased from around 42 sec to around 9 sec.


moorepants:
Nice reduction!

PR #396, opened 2025-03-05
simple example, but difficult to converge

Peter230655:
Basically, a particle should follow a line in 2D.
I tried two ways:
- restricting the movement of the particle by an inequality, something done many times before
- Multiplying the forces acting on the particle by a hump function, which is approx. one inside the band, and approx. zero outside.

Details are in the docstrings.
The amazing thing for me was - and this is why I am proposing it here: Despite the **absolutely innocent look**, it was quite difficult to get it to converge, at least the ways I could think of. (I never tried to give the curve itself as initial guess - this would have felt a bit like cheating.)
The example shows a way to iterate towards a solution, of course solutions are used, thes iterations took over 5 min on my PC
I feel, it is a bit much for beginner, and definitely not advanced, so I proposed it here.

Running time was very short on my PC, less than 7 sec.

NB:
The version with the hump function (version1) behaves like one would expect:
- if very large forces are allowed, it shoots directily to the final destination
- if only small forces are allowed, it stays within the band close to the curve
- with intermediate forces, it leaves the band a bit only.


moorepants:
This is a crazy example where you are essentially wrapping an optimization problem inside another optimization problem.

Peter230655:
> This is a crazy example where you are essentially wrapping an optimization problem inside another optimization problem.

This was the only way I could get it to converge: I start with a simple problem, beta = 1 and radius = 5 and then step by step to beta = 6 and radius = 1.
If I start with beta = 6 and radius = I I could not get it to converge.

moorepants:

Note that "difficultly in getting to converge" probably means not a good candidate for the examples gallery for learners.

Peter230655:
Nothing new. I do not thin, I ever intended it for examples-gallery.
If I make a simulation, which I think you may be interested in looking at, I can only push it to examples-gallery. This is why it landed there.

Peter230655:
moved to pst-notebooks

PR #398, opened 2025-03-06
added coment about backend=numpy, takes care of #397

Peter230655:
I put the comment under the Problem class, because I would look there first for the meaning of any kwards in Problem.


Peter230655:
Thanks!
Did you notice that I am getting better with these Latex symbols? :-)  :-)

moorepants:
I notice many many improvements :)

PR #399, opened 2025-03-06
Support `parallel=True` by getting OpenMP working on macOS

### LINKED ISSUE #161: openmp does not work on macOsx
From the arm builds:

```
Run python examples/parallel_example.py
clang: error: unsupported option '-fopenmp'
/Users/runner/miniconda3/envs/test/lib/python3.10/site-packages/opty/utils.py:572: UserWarning: openmp is not installed or not working properly, request for parallel execution ignored.
  warnings.warn(msg)
Time for non-parallel: 5.751712458000043
Time for parallelized: 5.257540124999991
```

brocksam:
Closes #161.

After a bunch of digging, it looks like the only thing missing was an additional `-Xclang` flag to the C compiler on Darwin systems.

Having a look at macOS CI jobs, this changes looks to give ~4x speedup on the OpenMP example with `parallel=True`. So I think it's working. Also tested locally and working there too.

### DIFF
--- a/opty/utils.py
+++ b/opty/utils.py
@@ -551,6 +551,8 @@ int main(void) {
         # .compile() should return ['test.o'] on linux
         if sys.platform == "win32":
             ccompiler.compile([filename], extra_postargs=['/openmp'])
+        elif sys.platform == "darwin":
+            ccompiler.compile([filename], extra_postargs=['-Xclang', '-fopenmp'])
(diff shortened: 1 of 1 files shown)


moorepants:
Sweet!

moorepants:
Yes, definitely working on the CI jobs.

Issue #395, opened 2025-03-02
Create an example that shows simultaneous optimization of a parameter and input trajectories


moorepants:
This could be done in the one legged time trial, for example. If you let the seat post height be a free variable, then it would find the optimal seat post height to win the race.

In biomechanics we like to optimize a prosthetics foot damping/stiffness or exoskeleton for carrying a weight, etc. There are many possible examples. Or things like the skateboard paper: board geometry for highest ollie.

Peter230655:
Maybe even simpler, but boring: The curve parameters of the particle in a tube. Presumably a straight line would be optimum.

Peter230655:
The motion of the mississippi steamboat (also) depends on two drag coefficients, which at the time I set to 0.74 each.
Now I let them be free and bounde them 0 < cS, cW < 100.
It picked cS 0 0.29, cW 0 100 but the optimum time did not change much at all.
This only shows that it works, boring otherwise.
Need to come up with something more interesting - like your chair height.

moorepants:
>  Again, it works, maybe a very little bit more interesting, I would have expected Omega = 0

Yes, curious.

Peter230655:
When I let in select alpha =
measure of acceleration of the disc it set it to 10^(-9) and Omega = 60, effectively meaning the disc never started to move. A contradiction. You want me to check it in more detail? I have time!

Peter230655:
I played around a bit more. Strange results:
I let Omega free:  Omega = -2.08 is found objective value = 4,800
I fix Omega = -2.08 now Objective value = 4,200
Another time if found Omega = -0.7, objective value = 5,000

I bound -10 < Omegas < 10. It finds optimnal Omega = 5.4, objective value = 4,400

This is a variable time interval simulation.

Either I make stupid mistakes, or it seems to have many local minimae.

Peter230655:
To see local minima, I ran the simulation for different values of Omega, and plotted them.
The plots seem to indicate that there are local minima - as was to be expected.
[Objective_value_tight.pdf](https://github.com/user-attachments/files/19141905/Objective_value_tight.pdf)
For the last plot, I ran the same range of Omega twice to check for numerical errors.

moorepants:
If you want to iterate through a problem, I suggest making a PR and putting the comment there or opening a new issue about that.

Peter230655:
> If you want to iterate through a problem, I suggest making a PR and putting the comment there or opening a new issue about that.

[...]

PR #401, opened 2025-03-08
optimize shape of steamboat, example for #395

Peter230655:
I took the steamboat example form opty/examples-gallery/intermediate and optimized its shape:
The shape is a rectangle with sides $a_S, b_S$ and I let opty find the best values so that $s_S \cdot b_S = \text{constant}$
- the old steamboat reaches its destination in 6.5 sec, the optimizerd one needs 5 sec only. Also the path changed as one can see by comparing the two steamboats. 
- When I set $0 \leq a_S, b_S \leq 20$ it got slightly better results than when I set $0.1 \leq a_S, b_S \leq 20$ - but in both cases the optimal vales were far away from the lower limit.
- Maybe an example for issue #395.


Peter230655:
Fixed f string

moorepants:
It looks like you are creating a whole new example, would it not be possible to simply modify the existing one with this new feature? Even if it is a second part in the same example.

Peter230655:
> It looks like you are creating a whole new example, would it not be possible to simply modify the existing one with this new feature? Even if it is a second part in the same example.

It IS the existing one. All I changed was that it now optimizes the length and width of the boat, while keeping the area constant.
I'd suggest to remove the existing example, and put this on into examples gallery instead.

moorepants:
> I'd suggest to remove the existing example, and put this on into examples gallery instead.

No, you should make a new branch and edit the existing file. Don't change the filename or make a new copy. It is the same process you do when you edit `direct_collocation.py` for example.

Peter230655:
Say, I want to change a simulation which already is, say,  in opty/examples-gallery/advanced.
Would this approach be better?
- I set up a new local branch
- in it I move to /../advanced.
- I modify the simulation there, I do not change its name
- when it works, I push the new branch
(I am still a little bit scared of git, so unless I get your approval, I would not dare such an approach)

NB: was the example what you had in mind?

Peter230655:
> > I'd suggest to remove the existing example, and put this on into examples gallery instead.
> 
> No, you should make a new branch and edit the existing file. Don't change the filename or make a new copy. It is the same process you do when you edit `direct_collocation.py` for example.

I will do, Should I close the other PR?

moorepants:
I would close this PR and open one that edits the existing file.

Peter230655:
> I would close this PR and open one that edits the existing file.

Already working on it. :-)

Peter230655:
Replaced by PR #402

PR #402, opened 2025-03-09
optimize shape of boat. Example for #395

Peter230655:
This replaces PR #401 which I closed. Maybe example for issue #395


examples-gallery/intermediate/plot_mississippi_steamboat.py
```diff
@@ -8,6 +8,8 @@
 - Show the usage of opty's variable node time interval feature.
 - Show an objective function which is a weighted average of the time and of the
   energy needed.
+- Show how to use an additional state variable to allow opty to optimize
+  parameters.
```
moorepants, line 12: Why would you make this an additional state variable? opty can optimize a constant just fine.

Peter230655: I want aS * bS = constant, the surface of the boat.
How could this be done otherwise?

moorepants: If you introduce `A` as the area and then `R` as the ratio of the length and width, couldn't `R` be an unknown parameter?

moorepants: Or both?

moorepants: Are you imagining that the boat area changes over time? If so, then I don't know if your equations of motion would be valid anymore if area is a function of time.

Peter230655: R := aS/bS
A = aS*bS = const =R*b^2
Right now, not sure how to proceed from here.

Peter230655: > Are you imagining that the boat area changes over time? If so, then I don't know if your equations of motion would be valid anymore if area is a function of time.

No, it does not change over time.
I just do not know how to enforce ``function(parameters) = 0`` other than using a state variable, which I can bound to be constant.

moorepants: If you have an equation of motion:

```
m*v.diff(t) + c*v + k*x = 0
```

then you can put bounds on `c: (0.0, 10.0)` and ask opty to find the optimal value of c. Are you trying to do something different than that?

moorepants: Maybe this is more helpful. You can write an eom like:

```
m*v.diff(t) + A*v**2 = 0
```

or

```
m*v.diff(t) + l*w*v**2 = 0
```

Peter230655: > Maybe this is more helpful. You can write an eom like:
> 
> ```
> m*v.diff(t) + A*v**2 = 0
> ```
> 
> or
> 
> ```
> m*v.diff(t) + l*w*v**2 = 0
> ```

I think, what I want to do is much simpler:
I want ``aS * bS = constant = surface of the boat``, and opty should select the best aS, bS.
The only way I know to enforce such an equality is to introduce a new state variable, ``area``, add
``area - aS * bS`` to the eoms, and force ``area`` 0 constant with bounds on ``area``

moorepants: For #395 I'm interested in showing how you can optimize an unknown constant and unknown trajectory simultaneously (without any hacks). We can add this boat edit but it makes a layer of complications that does not show the simply feature of opty that I'm suggesting in #395.

If you write your equations of motion as a function of the area and a ratio of width to height, then you I believe you can avoid the hack.

Peter230655: > For #395 I'm interested in showing how you can optimize an unknown constant and unknown trajectory simultaneously (without any hacks). We can add this boat edit but it makes a layer of complications that does not show the simply feature of opty that I'm suggesting in #395.
> 
> If you write your equations of motion as a function of the area and a ratio of width to height, then you I believe you can avoid the hack.

Like this:
R := a/b
A = a * b

==> b = sqrt(A/R)
==> a = sqrt(A*R)

Let me try this!

moorepants: Your hack works, but if you want this to solve #395 I don't want a hack involved.

moorepants: In fact we really should have a beginner example that is super simple that shows what #395 is after, otherwise the reader may not see the forest for the trees.

Peter230655: > In fact we really should have a beginner example that is super simple that shows what #395 is after, otherwise the reader may not see the forest for the trees.

Got your point. Maybe like this:
- the steamboat goes to intermediate to show how the hack would work in general.
- I think, that my block sliding on a hill can be changed, so that the shape of the hill is optimized.

Peter230655:
The steamboat works fine like this, without the hack.
I will beautiufy it a bit and push it.

Interesting:
The minimum it finds is a bit worse, than the minimum with the hack. No idea why.

examples-gallery/intermediate/plot_mississippi_steamboat.py
moorepants: No the steam boat stays in intermediate, as it is complex due to other reasons.

Peter230655: > No the steam boat stays in intermediate, as it is complex due to other reasons.

Let me beautify it and push it

moorepants:

We have the quarter car model that shows simultaneous trajectory and parameter ID.

Peter230655:
Now that we have the quarter car, indeed this does not offer anything new.

Peter230655:
Did not move to past-notebooks, not an interesting example.

PR #406, opened 2025-03-13
Add example: One Wheeler on Bumpy Road

Peter230655:
This models a one wheeler on a bumpy road, a simple example for #395 
The convergence was more tricky than I thought, mainly due to the constraint that the force on the wheel should always be negative = no jumping off the road.


moorepants:
Hi Peter, please try to make informative titles of your PRs. I updated this one with what I would expect.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,370 @@
+r"""
+One Wheeler on Bumpy Road
```
moorepants, line 2: The model you are building here is traditionally referred to as a "quarter car model" not a "one wheeler". I think the former would be better for the title and language.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,370 @@
...
+  [m/s^2]
+- :math:`help_4` : holds the accelerations of the wheel, only needed for the
+  animation. [m/s^2]
+
+
+
+**Fixed Parameters**
+
+- :math:`m_{car}` : mass of the car [kg]
+- :math:`m_{wheel}` : mass of the wheel [kg]
```
moorepants, line 47: These are typically called the "sprung" and "unsprung" masses in this model. It isn't just the mass of the wheel, per se.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,370 @@
...
+- :math:`ux_{car}` : x velocity of the car [m/s]
+- :math:`uz_{car}` : z velocity of the car [m/s]
+- :math:`help_1` : to ensure the wheel always has a negative load, meaning it
+  will not jump [N]
+- :math:`help_2` : to ensure that the vertical motions of the body are not too
+  large [m]
+- :math:`help_3` : holds the acceleration of the body, which is to be minimized
+  [m/s^2]
+- :math:`help_4` : holds the accelerations of the wheel, only needed for the
+  animation. [m/s^2]
```
moorepants, line 40: "help" isn't a very informative variable name, why just name the something related to what they are?

moorepants:
What is the reason to have the car start from 0 speed, go to a max speed, and then back to zero? versus, for example, just riding over the road at a constant speed?

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,370 @@
...
+coordinates = coordinates.row_join(P_wheel.pos_from(O).to_matrix(N))
+
+pL, pL_vals = zip(*par_map.items())
+coords_lam = sm.lambdify(list(state_symbols) + [c, fx, k] + list(pL),
+    coordinates, cse=True)
+
+def init_plot():
+    fig, ax = plt.subplots(figsize=(7, 7))
+    ax.set_xlim(xmin, xmax)
+    ax.set_ylim(ymin, ymax)
```
moorepants, line 313: You are plotting below y = 0, I think you can just plot above y = 0 to save lots of whitespace in the figure.

moorepants:
Here is a function for animating a spring: https://github.com/moorepants/resonance/blob/master/resonance/functions.py#L43

moorepants:
Here is a notebook on this problem from my old course: https://moorepants.github.io/resonance/07/07_vertical_vibration_of_a_quarter_car.html, maybe that helps with context, language, etc.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
Peter230655: > These are typically called the "sprung" and "unsprung" masses in this model. It isn't just the mass of the wheel, per se.

I did not know this, will change the name.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
Peter230655: > "help" isn't a very informative variable name, why just name the something related to what they are?

In the docstrings I explain, but can easily change them. Will think of a better name.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,370 @@
...
+aux_1 = (rough_surface(x_car).diff(t)).subs({x_car.diff(t): ux_car})
+aux_1 = aux_1.diff(t)
+eom = eom.col_join(sm.Matrix([
+                                help1 - F_contact,
+                                help2 - (z_car - rough_surface(x_car)),
+                                help3 - (uz_car.diff(t)),
+                                help4 - aux_1
+]))
+
+print(f'eoms contains {sm.count_ops(eom)} equations and have shape {eom.shape}')
```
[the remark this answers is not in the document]
Peter230655: > I'm curious why this has 819 operations? This model can be written by hand on just a few lines (in general).

Mostly on account of the bumpy surface. this function, and its time derivatives and second time derivatives enter the eoms. When I made the road very simply, it had maybe 100 operations.

Peter230655:
> What is the reason to have the car start from 0 speed, go to a max speed, and then back to zero? versus, for example, just riding over the road at a constant speed?

No special reason, just habit I suppose. As a slight benefit, it showes that the acceleration of the wheel is largest when the quarter car is fastest (but of course on great revelation)

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
Peter230655: > You are plotting below y = 0, I think you can just plot above y = 0 to save lots of whitespace in the figure.

If I only< plot above y = 0, the blue arrows cannot be seen. This is why I did it.

moorepants: Currently the interesting part is very small and it is hard to differentiate the arrows as drawn. I think you have too few animation frames also because the arrows jump rapidly.

Peter230655:
> Here is a notebook on this problem from my old course: https://moorepants.github.io/resonance/07/07_vertical_vibration_of_a_quarter_car.html, maybe that helps with context, language, etc.

Clever way to create a saw tooth function!
opty would not accept it. I believe, it wants (preferably infinitely often) differentiable functions only.
Also sympy.piecewise(....) gave trouble to me before, for the same reason, I think.

I just used a sum of four sin functions for my rough road, and just randomly picked the parameters. I did not start with a function and then retained the first few Fourier coefficients.

My main idea was to show on a very simple example, how opty can optimize a unknown trajectory (the dampening coefficient) and a free parameter (the spring constant)
Most of the time I spend trying to find out why it did not converge initially.

Peter230655:
> Hi Peter, please try to make informative titles of your PRs. I updated this one with what I would expect.

I always forget, that commit sets the title. Sorry!

moorepants:
> I always forget, that commit sets the title. Sorry!

Commit messages should also be more informative.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
Peter230655: > Currently the interesting part is very small and it is hard to differentiate the arrows as drawn. I think you have too few animation frames also because the arrows jump rapidly.

I tried to keep down the running time! :-) Was only 4.5 sec on my PC.
It is definitely not the most interesting simulation I ever saw! :-)

moorepants:
> unknown trajectory (the dampening coefficient) 

How would such a thing work in real life? If it could sense the upcoming road profile, then I guess it could adaptively change the damping, but you don't have such a controller built into this model (do you?)

You already have the propulsion force as an unknown trajectory, so why not search for damping and stiffness as the be static constants (like real suspension in a car).

Peter230655:
> > I always forget, that commit sets the title. Sorry!
> 
> Commit messages should also be more informative.

I agree, but I read somewhere ( I believe in the git carpentry?) that they should be short.

moorepants:
Short is fine, but they need to be short and informative.

Peter230655:
> > unknown trajectory (the dampening coefficient)
> 
> How would such a thing work in real life? If it could sense the upcoming road profile, then I guess it could adaptively change the damping, but you don't have such a controller built into this model (do you?)
> 
No I do not
> You already have the propulsion force as an unknown trajectory, so why not search for damping and stiffness as the be static constants (like real suspension in a car).

Convergence was much better when I made c an unknown trajectory.
I just read that luxury cars have it, no idea how it works.

moorepants:
Maybe this is helpful to understand active damping: https://vtechworks.lib.vt.edu/bitstream/handle/10919/35884/Chapter2.pdf

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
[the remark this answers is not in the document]
moorepants: ok

moorepants:
If this is a quarter car model, why don't you plot the point of the unsprung mass?

Peter230655:
> If this is a quarter car model, why don't you plot the point of the unsprung mass?

I thought I did. It is a small red dot on the street.

moorepants:
I wonder if you make the reference point in the plot the vehicle and let the road pass by, you'd get a better animation. Like: https://www.moorepants.info/blog/introducing-resonance.html

The animation is hard to follow and to see that you are somewhere reducing the accelerations of the sprung mass. It is better with more frames (not so jumpy).

Peter230655:
> I wonder if you make the reference point in the plot the vehicle and let the road pass by, you'd get a better animation. Like: https://www.moorepants.info/blog/introducing-resonance.html
> 

Definitely better! Let me try this.

> The animation is hard to follow and to see that you are somewhere reducing the accelerations of the sprung mass. It is better with more frames (not so jumpy).

Already increasd from 25fps to 100 fps. I can do nore. Is the running time still o.k.?
(this is my only concern with fps)

moorepants:
I said the FPS was fine now and it isn't jumpy. But it is hard to follow the arrows as drawn.

Peter230655:
Changed the plot so the street is moving.
Made the animation more slow motion.

moorepants:
That looks really good now! Nice.

I still don't understand why there are not three dots though. Should there be one on the ground, at the wheel and at the vehicle?

Peter230655:
Thanks! Took we a while to get the street to move😊
The way I have it is that the wheel is always at the ground. It is not allowed to jump. Therefore the wheel and where it touches the ground are the same point. My wheel is just a particle.

moorepants:
But there is a spring between the wheel and ground in the quarter car model.

Peter230655:
> But there is a spring between the wheel and ground in the quarter car model.

I assumed my wheel to be completely rigid.
Would this be a spring with fixed k?
Woul there also be some dampening?

moorepants:
The classic quarter car has a sprung and unsprung mass. Between the ground and the unsprung mass is a spring representing vertical stiffness of the tire and between the unsprung and sprung is a spring and damper representing the suspension system.

Peter230655:
> The classic quarter car has a sprung and unsprung mass. Between the ground and the unsprung mass is a spring representing vertical stiffness of the tire and between the unsprung and sprung is a spring and damper representing the suspension system.

Now I finally understand your "three dots"".
I have to think, how to incorporate my no-jump condition. Likely I have to give the tire a diameter > 0.

moorepants:
I think the problem is more interesting if framed as "what is the optimal damping and/or stiffiness of the suspension that results in the best ride comfort for the sprung mass (assuming there is also an unsprung mass)."

Peter230655:
> I think the problem is more interesting if framed as "what is the optimal damping and/or stiffiness of the suspension that results in the best ride comfort for the sprung mass (assuming there is also an unsprung mass)."

I think, I have an idea how to do it nicely. I will try over the weekend.

Peter230655:
- added spring to unsprung mass (is it still called upsprung mass, even if it has a spring?)
- changed doscstrings a bit
- chaned plot a bit.

moorepants:
This looks really good. So sorry about continuing to nit pick. The tire stiffness is not typically a design parameter. The current solution seems to find that you should make a really soft tire, but we typically only have control over the suspension design parameters and the tire stiffness is more or less dictated by the tire manufacturer and their needs for wear, strength, etc. So I would simply set the value of the tire vertical stiffness to a known value.

Peter230655:
> This looks really good. So sorry about continuing to nit pick. The tire stiffness is not typically a design parameter. The current solution seems to find that you should make a really soft tire, but we typically only have control over the suspension design parameters and the tire stiffness is more or less dictated by the tire manufacturer and their needs for wear, strength, etc. So I would simply set the value of the tire vertical stiffness to a known value.

Only like this I learn, so thanks!!
I had no idea about a tire stiffness, so I let opty pick it.
What would be a typical value?

moorepants:
A google search landed me on this first page: https://kktse.github.io/jekyll/update/2021/07/18/re71r-255-40-r17-tire-vertical-stiffness.html

Peter230655:
- Used the wheel stiffnesses you gave me. (Of course the source is from, you, but I think, you told me I should not mention this in simulations)
- adjusted my parameters a bit: before my quarter car had as mass of 10 kg, now it is 350 kg
- I still let opty find the natural length of the tire spring. I need somer length, so opty may as well find the best.
- Iteration si difficult. If no solution is given, it tokk nearly 10 min on my PCV to find it.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,425 @@
...
+Objective
+---------
+
+- Show on a simple example how to simultaneously optimize free
+  parameters and unknown trajectories of a system.
+- Show how one can iterate from a simpler problem (road less bumpy) to a
+  harder problem (road more bumpy).
+- Show how to use an additional state variable to get the acceleration of a
+  body into the objective function.
+- Show how to avoid possible unwanted unknown trajectories ( here
```
moorepants, line 14:
```suggestion
- Show how to avoid possible unwanted unknown trajectories (here
```

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,425 @@
...
+  :math:`\dfrac{d^2}{dt^2}x_{car}`).
+
+
+Introduction
+------------
+
+A quarter car model must move from a starting point to a final point on a bumpy
+road. The body is connected to the wheel on the road by a linear spring and a
```
moorepants, line 22:
```suggestion
road. The body (sprung mass) is connected to the wheel (unsprung mass) on the road by a linear spring and a
```

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,425 @@
...
+I took the tire stiffness from here:
+
+https://kktse.github.io/jekyll/update/2021/07/18/re71r-255-40-r17-tire-vertical-stiffness.html
+
+
+**States**
+
+- :math:`x_{car}` : x position of the car [m]
+- :math:`z_{car}` : z position of the car [m]
+- :math:`z_{wheel}` : z position of the wheel [m]
```
moorepants, line 36:
```suggestion
- :math:`z_\textrm{wheel}` : z position of the wheel [m]
```

Words in variable names should not be italicized text.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,425 @@
...
+eom = eom.col_join(sm.Matrix([
+                    prevent_jump - (z_wheel - rough_surface(x_car)),
+                    steady_body - (z_car - rough_surface(x_car)),
+                    accel_body - (uz_car.diff(t)),
+                    accel_street - aux_1
+]))
+
+print(f'eoms contains {sm.count_ops(eom)} equations and have shape {eom.shape}')
+# %%
+# Set up the Optimization Problem
```
moorepants, line 162:
```suggestion
# Set Up the Optimization Problem
```

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,425 @@
...
+
+# %%
+#To be minimized:
+# :math:`\int (\dfrac{d}{dt}uz_{car})^2 dt + \text{weight} \cdot t_f`
+# ``weight`` is a scalar that can be used to adjust the importance of the
+# speed.
+weight = 1.e9
+
+def obj(free):
+    UZdot = np.sum([free[i]**2 for i in range(8*num_nodes, 9*num_nodes)])
```
moorepants, line 192:
```suggestion
    uz_dot = np.sum([free[i]**2 for i in range(8*num_nodes, 9*num_nodes)])
```

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,425 @@
...
+    return (UZdot)*free[-1] + weight*free[-1]
```
moorepants, line 193:
```suggestion
    return (uz_dot)*free[-1] + weight*free[-1]
```

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,425 @@
...
+
+# %%
+# Add the instance constraints and bounds.
+instance_constraints = (
+    x_car.func(t0) - 0.0,
+    ux_car.func(t0) - 0.0,
+    accel_street.func(t0) - 0.0,
+    accel_body.func(t0) - 0.0,
+    steady_body.func(t0) - 1.0,
+
```
moorepants, line 212: delete line

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,425 @@
...
+- :math:`k` : spring constant [N/m]
+- :math:`k1` : spring constant of the wheel [N/m]
+- :math:`l_{GW}` : equilibrium length of the wheel spring [m]
+
+
+**Unknown Trajectories**
+
+- :math:`f_x` : driving force [N]
+
+"""
```
moorepants, line 71: You need to explain somewhere that you are searching for the optimal suspension parameters to minimize the average acceleration felt by the sprung mass, given the road surface shape and whatever optimal propulsion you are choosing.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,425 @@
...
+# Use an existing solution if available, else iterate to find one.
+fname =f'quarter_car_on_bumpy_road_{num_nodes}_nodes_solution.csv'
+aaa = 10
+if os.path.exists(fname):
+    # use the existing solution
+    par_map[r3] = 0.1 + 0.0725*4
+    par_map[r4] = 0.0725*4
+
+    prob = Problem(obj,
+               obj_grad,
```
moorepants, line 240: over indented, 4 spaces per indentation

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,425 @@
...
+               eom,
+               state_symbols,
+               num_nodes,
+               interval,
+               known_parameter_map=par_map,
+               instance_constraints=instance_constraints,
+               bounds=bounds,
+               time_symbol=t,
+            )
```
moorepants, line 275: A better design would be to not create Problem every time you run an optimization. If the road roughness is parameterized by something in `par_map` you shouldn't have to create new `Problem`'s in the loop.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,425 @@
...
+            if i == 0:
+                initial_guess = np.random.rand(prob.num_free)
+            else:
+                initial_guess = solution
+            for _ in range(3):
+                solution, info = prob.solve(initial_guess)
+                initial_guess = solution
+                print(info['status_msg'])
+                print('Objective value', info['obj_val'])
+#np.savetxt(fname, solution, fmt='%.12f')
```
moorepants, line 287: don't have commented lines

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,425 @@
...
+# %%
+# Print optimal values of the free parameters.
+print('Sequence of unknown parameters',
+               prob.collocator.unknown_parameters)
+print(f'optimal value of dampening constant c =                  ' +
+      f'{solution[-4]:.2f}')
+print(f'optimal value of spring constant k =                     ' +
+      f'{solution[-3]:.3f}')
+#print(f'optimal value of wheel spring constant k1 =              ' +
+#      f'{solution[-3]:.2f}')
```
moorepants, line 297: don't have commented lines

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,425 @@
...
+# %%
+fig, axes = plt.subplots(11, 1, figsize=(7, 20), layout='tight')
+_ = prob.plot_trajectories(solution, axes=axes)
+# %%
+_ = prob.plot_constraint_violations(solution)
+# %%
+# Plot the road.
+r11, r22, r33, r44, r55 = [par_map[key] for key in [r1, r2, r3, r4, r5]]
+rough_surface_lam = sm.lambdify((x_car, r1, r2, r3, r4, r5),
+                                rough_surface(x_car), cse=True)
```
moorepants, line 310: You can move the plot to just below where you create the road roughness function way at the top.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,425 @@
...
+r11, r22, r33, r44, r55 = [par_map[key] for key in [r1, r2, r3, r4, r5]]
+fig, ax = plt.subplots(figsize=(6, 2), layout='tight')
+ax.plot(XX, rough_surface_lam(XX, r11, r22, r33, r44, r55))
+ax.set_xlabel('[m]')
+ax.set_ylabel('[m]')
+_ = ax.set_title('Road Profile')
+# %%
+# Animate the Solution
+# --------------------
+fps = 100
```
moorepants, line 321: You could mention that the fps is higher to clearly show the fast changing acceleration vectors.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,425 @@
...
+
+state_vals, input_vals, _, _ = parse_free(solution, 10, 1, num_nodes,
+                        variable_duration=True)
```
moorepants, line 324: why no use `Problem.parse_free()` now.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
Peter230655: > delete empty line

Deleted. I had left it originally to better see what refers to t0 and what refers to tf.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
Peter230655: > A better design would be to not create Problem every time you run an optimization. If the road roughness is parameterized by something in `par_map` you shouldn't have to create new `Problem`'s in the loop.

I do not know the inner workings of opty :-(
I thought that every time one changes something which is in the signature of ``Problem``, like par_map or bounds or instance_constraints one has to set up a new `Problem``. 
There is an issue, that one presently cannot change num_nodes without setting up a new ``Problem``. Is this the case only for num_nodes, everything else in the signature may be changed without setting up a new `Problem``?

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
Peter230655: > don't have commented lines

Forgot to remove.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
Peter230655: > You can move the plot to just below where you create the road roughness function way at the top.

It has to be below the declaration of par_map, as it needs it.
Where would be the best location?

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
moorepants: I realize that but it doesn't follow PEP8.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
moorepants: Problem should be agnostic to the number of nodes. The C code it generates certainly is, but maybe there are some details that are missed. I would try it.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
Peter230655: > I realize that but it doesn't follow PEP8.

Already changed.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
moorepants: I just thought it would be nice to have it just after the introduction of road roughness equation so you can see the result.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
Peter230655: > Problem should be agnostic to the number of nodes. The C code it generates certainly is, but maybe there are some details that are missed. I would try it.

But all the rest may be changed without re-creating Problem?

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
Peter230655: > I just thought it would be nice to have it just after the introduction of road roughness equation so you can see the result.

I agree, but then I would have to move par_map up there. Would this not be out of the usual structure of an optimization problem?

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
Peter230655: > why no use `Problem.parse_free()` now.

Just noticed an error in Problem.parse_free: it does not accept ``variable_duration``.
I need to correct it asap.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
Peter230655: > Problem should be agnostic to the number of nodes. The C code it generates certainly is, but maybe there are some details that are missed. I would try it.

Removing ``Problem`` in the loop worked!. I will test for num_nodes separately and revert.
I noticed, that there is not check, that ``backend`` is either ``numpy`` or ``cython``. Would such a check not make sense?

Peter230655:
Changed as per suggestions.
- I NEVER expected, that ``Problem`` was not needed in the loop! This saves time for such loops. Should this not be mentioned in the docs?
- I noticed, that the ``backend`` is not checked for wrong assignment. Should this not be checked?
- I noticed a problem with ``Problem.parse_free``. i will correct soonest.

Peter230655:
I checked: ``Problem.parse_free`` does not accept any arguments other than solution, as they are available. So, I think it is o.k.

moorepants:
> I NEVER expected, that Problem was not needed in the loop! This saves time for such loops. Should this not be mentioned in the docs? After all, the naive user will expect it must be.

You are welcome to add things to the documentation, but this is also a basic idea of mutable classes that is inherit to all Python objects.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
moorepants: It should not need to accept this as it is already defined in Problem.

moorepants:
> I noticed, that the backend is not checked for wrong assignment. Should this not be checked?

I don't know what you mean.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
moorepants: There is no defined structure other than you have to define things before they are needed.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
moorepants: I don't know what you mean.

Peter230655:
> > I noticed, that the backend is not checked for wrong assignment. Should this not be checked?
> 
> I don't know what you mean.

should it not raise an error if I set ``backend='nonsense'`` ?

Peter230655:
> > I NEVER expected, that Problem was not needed in the loop! This saves time for such loops. Should this not be mentioned in the docs? After all, the naive user will expect it must be.
> 
> You are welcome to add things to the documentation, but this is also a basic idea of mutable classes that is inherit to all Python objects.

Just shows my limited python skills once again! :-(
Dumb question: in the loop I simply set e.g.  ``par_map[key] = (a+i, b-i)``and it converged.
Should I have set ``prob.par_map[key] = (a+i, b-i)``

moorepants:
> should it not raise an error if I set backend='nonsense' ?

Yes, raising a ValueError is appropriate in this case.

moorepants:
> prob.par_map[key] = (a+i, b-i)

is better

but because the par_map is a mutable dictionary the reference in the main script is the same object

Peter230655:
> > prob.par_map[key] = (a+i, b-i)
> 
> is better
> 
> but because the par_map is a mutable dictionary the reference in the main script is the same object

I will change to make it look better.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,419 @@
...
+
+if os.path.exists(fname):
+    # use the existing solution
+         solution = np.loadtxt(fname)
+else:
+    # Iterate to find the solution. As the convergence is not easy, one has to
+    # start with a smooth road and then increase the roughness gradually.
+    # Before the bounds have to be tightended gardually.
+    # Here backend = 'cython' is used, as it is faster than 'numpy'. for solve.
+    prob.backend = 'cython'
```
moorepants, line 273: There is no need to change the backend from numpy and this wouldn't do anything here anyways. The backend only affects the initialization of the object.

moorepants:
> I have to set prob.collocator._known_parameter_map[key] = a+i

This is bad, we should not be setting private methods.

Peter230655:
> > I have to set prob.collocator._known_parameter_map[key] = a+i
> 
> This is bad, we should not be setting private methods.

This is what I thought, so I leave it as it is.

moorepants:
> > > I have to set prob.collocator._known_parameter_map[key] = a+i
> > 
> > 
> > This is bad, we should not be setting private methods.
> 
> This is what I thought, so I leave it as it is.

You made may of the attributes immutable in ConstraintCollocator, but some probably don't need to be. The safe approach is to make them immutable, but we can also now add setter methods that ensure everything in ConstraintCollocator is updated probably when they are set.

Peter230655:
> > > > I have to set prob.collocator._known_parameter_map[key] = a+i
> > > 
> > > 
> > > This is bad, we should not be setting private methods.
> > 
> > 
> > This is what I thought, so I leave it as it is.
> 
> You made many of the attributes immutable in ConstraintCollocator, but some probably don't need to be. The safe approach is to make them immutable, but we can also now add setter methods that ensure everything in ConstraintCollocator is updated probably when they are set.

This went through my mind: I recall I made them immutable - and still so easy to change them, just set e.g. par_map[key] = a somewhere in the program. In this case it is good that it works, but is there not a danger in general if it is so easy to change them?
Would you prefer it works as it is now, easy for iterations?
I have found that these iterations from a simple to a more difficult problem often work well.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
Peter230655: > Problem should be agnostic to the number of nodes. The C code it generates certainly is, but maybe there are some details that are missed. I would try it.

I set ``prob.collocator.num_collection_nodes = num_nodes + 10 * i``
in a loop around solve, and it raised an error

moorepants: I thought you made all attributes on collocator imutable?

moorepants: The whole loop thing makes this an intermediate problem not beginner. You can just give a good initial guess and leave his loop out.

If you want the loop in and you are changing things that are mutable, you don't need to make a new problem every loop iteration. If you want to change things that are immutable, then you would need the problem in the loop, but I would rather not see this kind of thing in the documentation. It is teaching what not to do, in some sense.

Peter230655: > The whole loop thing makes this an intermediate problem not beginner. You can just give a good initial guess and leave his loop out.
> 
aa
> If you want the loop in and you are changing things that are mutable, you don't need to make a new problem every loop iteration. If you want to change things that are immutable, then you would need the problem in the loop, but I would rather not see this kind of thing in the documentation. It is teaching what not to do, in some sense.

moorepants: > What do you think about this:

I don't want two of the same examples just for this. Yes it is true that making a loop with different initial guesses is a technique for finding an initial guess, but I don't want to promote re-instantiating a Problem within a loop like this. Best we fix Problem so that it isn't needed.

Telling the user that finding an initial guess with iteration can be said in a sentence "you may finding luck in getting a good initial guess by iterating". We don't have to show this process. It is clunky and not helpful for examples we are trying to keep computationally simply and short. It is just one of the nasty parts of solving non-trivial optimization problems

Peter230655: > The par_map is a dictionary. It itself is mutable. The attribute on ConstraintCollocator is not settable. So:
> 
> ```
> problem.known_parameter_map = new_dict
> ```
> 
> fails but this will not:
> 
> ```
> problem.known_parameter_map[v] = 12.0
> ```

I have learned a few things today! Thanks!
- How could ``problem.known_parameter_map[v] = 12.0`` be prevented from happening?

moorepants: > How could problem.known_parameter_map[v] = 12.0 be prevented from happening?

You would have to make a custom immutable dictionary class.

Peter230655: > > How could problem.known_parameter_map[v] = 12.0 be prevented from happening?
> 
> You would have to make a custom immutable dictionary class.

And this is the same with all mutable objects? If I gave state_symbols, say, as a list instead of a tuple, I could change the entries miday?

moorepants:
I prefer that we make Problem and ConstraintCollocator mutable where we can update them without compiling a new problem, but we made the attributes unsettable because it is currently unsafe to set them, i.e. the objects may not be updated in a way they should when you set them.

Peter230655:
> I prefer that we make Problem and ConstraintCollocator mutable where we can update them without compiling a new problem, but we made the attributes unsettable because it is currently unsafe to set them, i.e. the objects may not be updated in a way they should when you set them.

I think, I am slowly getting it - very good Sunday for me! :-)

Peter230655:
The road is a bit more 'boring' now, but the best I managed to find without the loops.

Peter230655:
> Take for example:
> 
> ```python
> class A():
>     def __init__(self, d):
>         self.count = len(d)
>         self.d = d
> ```
> 
> If you do:
> 
> ```python
> a = A({'a': 1, 'b': 2})
> a.d['c'] = 5
> a.count # returns 2 which is now wrong
> ```

I think, I am (slowly) getting it! Thanks!

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,401 @@
...
+
+A quarter car model must move from a starting point to a final point on a bumpy
+road. The body (sprung mass) is connected to the wheel (unsprung mass) on the
+road by a linear spring and a damper.
+Body and wheel are modeled by particles. Movement is in the X/Z plane. Gravity
+points in the negative Z direction.
+
+I took the tire stiffness from here:
+
+https://kktse.github.io/jekyll/update/2021/07/18/re71r-255-40-r17-tire-vertical-stiffness.html
```
moorepants, line 27: The vertical tire stiffness for an automobile car is selected as 250,000 N/m based on `an experiment <https://kktse.github.io/jekyll/update/2021/07/18/re71r-255-40-r17-tire-vertical-stiffness.html>`_.

Don't use "I" in examples.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,401 @@
...
+            + k1 * (l_GW - (z_wheel - rough_surface(x_car))) * N.z
+)]
+
+forces  = F_car + F_wheel
+
+kd = sm.Matrix([x_car.diff(t) - ux_car, uz_car - z_car.diff(t),
+                uz_wheel - z_wheel.diff(t)])
+
+KM = me.KanesMethod(N,
+                    q_ind=[x_car, z_car, z_wheel],
```
moorepants, line 143: indentation is 4 spaces

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,401 @@
...
+                    u_ind=[ux_car, uz_car, uz_wheel],
+                    kd_eqs=kd
+)
```
moorepants, line 146: The recommended indentation is either:

```
thing(
    a,
    b,
    c,
)
```

or 

```python
thing(a, b, c)
```

or 

```python
thing(a, b, c,
           d, e, f)
```

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,401 @@
...
+
+
+# %%
+# Add the constraints.
+# The 'detour' with aux_1 is needed, else an unkonwn trajectory
+# :math:`\dfrac{d^2}{dt^2}x_{car}` will be created by opty.
+aux_1 = (rough_surface(x_car).diff(t)).subs({x_car.diff(t): ux_car})
+aux_1 = aux_1.diff(t)
+eom = eom.col_join(sm.Matrix([
+                    prevent_jump - (z_wheel - rough_surface(x_car)),
```
moorepants, line 158:
```suggestion
                   prevent_jump - (z_wheel - rough_surface(x_car)),
```

moorepants: actually this indentation is not correct, you need to indent relative to `Matrix([` not `col_join(`.

If you set up a pep8 checker in your IDE (VS code?) it will fix/recommend such things for you automatically.

Peter230655: > actually this indentation is not correct, you need to indent relative to `Matrix([` not `col_join(`.
> 
> If you set up a pep8 checker in your IDE (VS code?) it will fix/recommend such things for you automatically.

This would be GREAT! I use VSC on windows. Thanks!

moorepants: Ok, set it up then. VS code has many options.

Peter230655: > Ok, set it up then. VS code has many options.

That's my problem :-)
Getting there with chatGPT..

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,401 @@
...
+
+    return (fig, ax, line1, line2, line3, line4, pfeil1, pfeil2, pfeil3,
+            street, circle)
+
+# Function to update the plot for each animation frame
+def update(t):
+    message = (f'running time {t:.2f} sec' +
+        f'\n The blue arrow is the ' +
+        f'accelerationon the wheel due to uneven street \n' +
+        f'The magenta arrow is the acceleration of the body, magnified 25' +
```
moorepants, line 364: Shouldn't the acceleration vectors be the same scale? Right now it looks like the sprung mass has much larger acceleration than the unsprung mass.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,401 @@
...
+    ax.set_xlim(xmin, xmax)
+    ax.set_ylim(ymin, ymax)
+    ax.set_aspect('equal')
+    ax.set_xlabel('x', fontsize=15)
+    ax.set_ylabel('z', fontsize=15)
+
+    # draw the road
+    XX = np.linspace(0, 10, 200)
+    street, = ax.plot(XX, rough_surface_lam(XX, r11, r22, r33, r44, r55),
+            color='black', lw=0.75)
```
moorepants, line 333: incorrect indentation

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
moorepants: this still needs to be removed

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,401 @@
...
+    # Sometimes random intial guess work fine.
+    np.random.seed(123)
+    initial_guess = np.random.rand(prob.num_free)
+    solution, info = prob.solve(initial_guess)
+    print(info['status_msg'])
+    print('Objective value', info['obj_val'])
+# %%
+# Print optimal values of the free parameters.
+print('Sequence of unknown parameters',
+               prob.collocator.unknown_parameters)
```
moorepants, line 278: bad indentation

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,401 @@
...
+        eom,
+        state_symbols,
+        num_nodes,
+        interval,
+        known_parameter_map=par_map,
+        instance_constraints=instance_constraints,
+        bounds=bounds,
+        time_symbol=t,
+        backend='numpy',
+)
```
moorepants, line 262: Not correct style, should be:

```python
Problem(
    obj,
    eom,
    ...
)
```

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,401 @@
...
+def obj(free):
+    uz_dot = np.sum([free[i]**2 for i in range(8*num_nodes, 9*num_nodes)])
+    return (uz_dot)*free[-1] + weight*free[-1]
+
+def obj_grad(free):
+    grad = np.zeros_like(free)
+    grad[9*num_nodes:10*num_nodes] = 2*free[8*num_nodes:9*num_nodes]*free[-1]
+    grad[-1] = (
+                + np.sum([free[i]**2 for i in range(8*num_nodes, 9*num_nodes)])
+                + weight,
```
moorepants, line 216: odd indentation

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
```diff
@@ -0,0 +1,401 @@
...
+# The 'detour' with aux_1 is needed, else an unkonwn trajectory
+# :math:`\dfrac{d^2}{dt^2}x_{car}` will be created by opty.
+aux_1 = (rough_surface(x_car).diff(t)).subs({x_car.diff(t): ux_car})
+aux_1 = aux_1.diff(t)
+eom = eom.col_join(sm.Matrix([
+                   prevent_jump - (z_wheel - rough_surface(x_car)),
+                   steady_body - (z_car - rough_surface(x_car)),
+                   accel_body - (uz_car.diff(t)),
+                   accel_street - aux_1,
+]))
```
moorepants, line 162: Still odd indentation.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
Peter230655: > Shouldn't the acceleration vectors be the same scale? Right now it looks like the sprung mass has much larger acceleration than the unsprung mass.

The acceleration of the body is 10 times less than the acceleration of the street. I can put them on the same scale, but the one will be hardly visible.

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
Peter230655: > Still odd indentation.
I tried to go by this:
image

examples-gallery/beginner/plot_wheel_on_bumpy_road.py
moorepants: For this problem, I think that would be the point to show off, that the acceleration on the sprung mass is hardly anything.

Peter230655: > For this problem, I think that would be the point to show off, that the acceleration on the sprung mass is hardly anything.

Already made them same scale. It IS more impressive!

Peter230655:
> If I run flake8 (a PEP8 checking tool) on your file this is what it returns:
> 
> ```
> $ flake8 examples-gallery/beginner/plot_wheel_on_bumpy_road.py 
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:87:1: F401 'time' imported but unused
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:90:1: E265 block comment should start with '# '
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:108:1: E302 expected 2 blank lines, found 0
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:111:9: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:117:1: E305 expected 2 blank lines after class or function definition, found 1
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:120:28: E221 multiple spaces before operator
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:127:8: E225 missing whitespace around operator
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:130:1: E124 closing bracket does not match visual indentation
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:135:1: E124 closing bracket does not match visual indentation
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:137:7: E221 multiple spaces before operator
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:146:1: E124 closing bracket does not match visual indentation
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:162:1: E124 closing bracket does not match visual indentation
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:164:80: E501 line too long (80 > 79 characters)
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:167:1: E265 block comment should start with '# '
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:192:34: E127 continuation line over-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:201:1: E265 block comment should start with '# '
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:207:1: E302 expected 2 blank lines, found 1
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:211:1: E302 expected 2 blank lines, found 1
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:222:1: E305 expected 2 blank lines after class or function definition, found 1
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:249:8: E225 missing whitespace around operator
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:252:9: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:253:9: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:254:9: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:255:9: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:256:9: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:257:9: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:258:9: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:259:9: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:260:9: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:261:9: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:262:1: E124 closing bracket does not match visual indentation
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:266:10: E111 indentation is not a multiple of 4
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:266:10: E117 over-indented
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:278:16: E127 continuation line over-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:279:7: F541 f-string is missing placeholders
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:281:7: F541 f-string is missing placeholders
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:283:1: E265 block comment should start with '# '
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:285:7: F541 f-string is missing placeholders
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:320:5: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:322:1: E302 expected 2 blank lines, found 1
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:333:13: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:339:62: E261 at least two spaces before inline comment
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:340:63: E261 at least two spaces before inline comment
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:341:50: E261 at least two spaces before inline comment
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:342:63: E261 at least two spaces before inline comment
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:353:13: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:360:1: E302 expected 2 blank lines, found 1
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:362:9: F541 f-string is missing placeholders
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:362:9: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:363:9: F541 f-string is missing placeholders
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:363:9: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:364:9: F541 f-string is missing placeholders
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:364:9: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:365:9: F541 f-string is missing placeholders
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:365:9: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:366:9: F541 f-string is missing placeholders
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:366:9: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:375:21: E128 continuation line under-indented for visual indent
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:377:7: E225 missing whitespace around operator
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:395:1: E305 expected 2 blank lines after class or function definition, found 1
> examples-gallery/beginner/plot_wheel_on_bumpy_road.py:399:5: E128 continuation line under-indented for visual indent
> ```

I am 50% done installing fluke8 on my VSC. This sort of thing is VERY difficult for me, even with the help of chatGPT! I hope to finish tomorrow.

moorepants:
All I did was:

```
conda install flake8
flake8 plot_bumpy_road.py
```

Peter230655:
> All I did was:
> 
> ```
> conda install flake8
> flake8 examples-gallery/beginner/plot_wheel_on_bumpy_road.py 
> ```

So I would do this in Anaconda Prompt, in my environment?

moorepants:
Yes

Peter230655:
> Yes

I will try this first thing tomorrow morning! Thanks!

PR #407, opened 2025-03-16
check for acceptable backend

Peter230655:
Added a check that backend be either 'numpy' or 'cython', else a value error will be raised.

### DIFF
--- a/opty/tests/test_direct_collocation.py
+++ b/opty/tests/test_direct_collocation.py
@@ -2257,3 +2257,68 @@ def test_check_bounds_conflict():
     initial_guess = np.zeros(prob.num_free)
     prob.check_bounds_conflict(initial_guess)
 
+def test_backend():
+    """Test that only backend='numpy', backend='cython and no backend given are
+    accepted. A value error is raised otherwise"""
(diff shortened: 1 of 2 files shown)


moorepants:
You need a unit test.

opty/tests/test_direct_collocation.py
```diff
@@ -2257,3 +2257,68 @@ def obj_grad(free):
...
+        Problem(
+            obj,
+            obj_grad,
+            eom,
+            state_symbols,
+            num_nodes,
+            interval,
+            time_symbol=t,
+            backend='nonsensical',
+        )
```
moorepants, line 2292: You can simply add these few lines to an existing test function instead of making a whole new test function. All other scenarios are already tested in the other functions.

Peter230655:
As suggested, I added the test to another test function, I used test_prob_parse_free. Strictly speaking, I only test whether a ValueError is raised if a wrong backend is given, but as mentioned, the right backend are tested for implicitly in other tests.

moorepants:
Perfect, thanks!

PR #409, opened 2025-03-17
Add brachistochrone example

Peter230655:
As suggested, I have push it.
I want opty to find the Brachistochrone from A(0/0) to B(b1/b2) to give shortest distance.
I procceded like this:
- let f(x) be the curve to be found.
- if the particle is at x(t), I look at the tangent to f(x(t)).
- I split the gravitational force (in the -N.y direction) in a component normal to the tangent (which does no wrk, as no friction) and a component parallel to the tangent.
- the tangent component is split into its N.x and N.y directions.
- the angle of the tangent with the vertical is the control parameter, ``beta`` in the program.

opty always selects the line from A to B. So there must be a big fla in my thinking - but cannot find it!
As the simulation does not work, it is not yet written to examples-gallery standards.

This is how I split the force:
[Brachistrochone.pdf](https://github.com/user-attachments/files/19282443/Brachistrochone.pdf)


examples-gallery/beginner/plot_brachistochrone.py
```diff
@@ -0,0 +1,152 @@
...
+P.set_vel(N, ux*N.x + uy*N.y)
+
+body = [me.Particle('body', P, m)]
+delta = sm.pi/2 - beta
+forces = [(P, -m*g*sm.cos(beta)**2*N.y + m*g*sm.sin(beta)*sm.cos(beta)*N.x)]
+
+kd = sm.Matrix([ux - x.diff(t), uy - y.diff(t)])
+
+kane = me.KanesMethod(N, q_ind=[x, y], u_ind=[ux, uy], kd_eqs=kd)
+fr, frstar = kane.kanes_equations(body, forces)
```
moorepants, line 55: It's a bit overkill to use Kane's method for such a problem (or sympy mechanics at all).

Peter230655: > It's a bit overkill to use Kane's method for such a problem (or sympy mechanics at all).

I agree - but at least a mistake from the eoms is avoided for sure.

examples-gallery/beginner/plot_brachistochrone.py
```diff
@@ -0,0 +1,152 @@
...
+m, g = sm.symbols('m g')
+x, y, ux, uy = me.dynamicsymbols('x y ux uy')
+beta = me.dynamicsymbols('beta')
+
+P.set_pos(O, x*N.x + y*N.y)
+P.set_vel(N, ux*N.x + uy*N.y)
+
+body = [me.Particle('body', P, m)]
+delta = sm.pi/2 - beta
+forces = [(P, -m*g*sm.cos(beta)**2*N.y + m*g*sm.sin(beta)*sm.cos(beta)*N.x)]
```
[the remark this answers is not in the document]
moorepants: I defined beta as positive rotation about N.z

examples-gallery/beginner/plot_brachistochrone.py
```diff
@@ -0,0 +1,152 @@
...
+
+#%%
+# Calculate the Brachistochrone from (0, 0) to (b1, b2). It does not work for
+# all values of b1 and b2. I did not investigate
+def func(X0):
+    """gives the Brachistochrome equation for the starting point at (0/0) and
+    the final point at (b1/b2)"""
+    R = X0[0]
+    theta = X0[1]
+    return [R*theta - R*np.sin(theta) - b1, R*(1.0 - np.cos(theta)) - b2]
```
moorepants, line 119: I changed x to 20.0 and got:

image

So maybe your function to create the mathematical curve is off.

Peter230655: I took it from chatGPT. I noticed, it does not gtive results for all combinations of (x, y) but did not investigate as it looked 'reasonable' for, say, (10, -10).

examples-gallery/beginner/plot_brachistochrone.py
```diff
@@ -0,0 +1,152 @@
...
+                obj_grad,
+                eom,
+                state_symbols,
+                num_nodes,
+                interval,
+                known_parameter_map=par_map,
+                instance_constraints=instance_constraint,
+                bounds=bounds,
+                backend='numpy',
+)
```
moorepants, line 109: I don't think you have anything in your problem definition that forces beta to be smooth.

Peter230655: > I don't think you have anything in your problem definition that forces beta to be smooth.

I do not - but beta is just an unkown_input_trajectory? I have never made any assumptions about them.

examples-gallery/beginner/plot_brachistochrone.py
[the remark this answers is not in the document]
Peter230655: > Are you missing gravity and only have normal force? This is what I got:
> 
> ```
> g*m*cos(beta(t))**2*N.x + (g*m*sin(beta(t))*cos(beta(t)) - g*m)*N.y
> ```

I started with the gravity force only -m*g*N.y
Then I split it into a component normal to the tangent of the curve the particle should follow and a component parallel to that tangent.
Then I split the tangential force into its N.x, N.y components.
So, I think, I did not forget gravity?
(I any case, your modified eoms seem to give a straight line, too.)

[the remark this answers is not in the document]
moorepants: There should be a bare -mg term right?

examples-gallery/beginner/plot_brachistochrone.py
moorepants: input trajectories do not have to be continous, but you do want a continuous solution in this case

Peter230655: > input trajectories do not have to be continous, but you do want a continuous solution in this case

Yes, beta must be continuous. Maybe beta.diff(t) = u_beta, as additional eom? I will try.

examples-gallery/beginner/plot_brachistochrone.py
Peter230655: Other internet sources gave the same equations :-)
But I will double check.

examples-gallery/beginner/plot_brachistochrone.py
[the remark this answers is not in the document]
Peter230655: > There should be a bare -mg term right?

I do not think so. Say, the particle was sliding on a plane. Then the splitting up of the forces as I did would describe the motion of the particle on that plane?

examples-gallery/beginner/plot_brachistochrone.py
moorepants: maybe your root finding finds the wrong solution

examples-gallery/beginner/plot_brachistochrone.py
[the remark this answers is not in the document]
Peter230655: > there is two forces acting on the particle: normal and gravity
> 
> ```
> gravity = -m*g*y_hat
> normal = m*g*cos(beta)*u_hat
> ```
> 
> where u_hat is the direction normal to the surface
> 
> it does not seem to me you have the gravity term

I just failed mechanics 101! :-(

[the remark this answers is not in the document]
Peter230655: Now I went completely formal (sympy.mechanics be **praised!**)
frame A rotates around N.z with angle beta.
The 'plane' is A.x
speed constraint: P.vel(N).dot(A.y) = 0
Now it seems to look better!

examples-gallery/beginner/plot_brachistochrone.py
Peter230655: > maybe your root finding finds the wrong solution

I need to check this. But now, with my formal approach it works!! :-)
sysmpy.mechanics is just GREAT for guys who fail mechanics 101

Peter230655:
- eoms are working now.
- finding the parameters R, $\Theta$ of the brachisatochrone is very sensitive to the initial conditions.
- The problem does not seem so simple for opty: It takes over 5,000 iterations to find a solution.
- The gif does not work. I did not play around with it, I do not want to destroy all gifs as I did once before.
- I hope indentations are better. I did not dare using flake8 yet - afraid to mess up things.

examples-gallery/beginner/plot_brachistochrone.py
```diff
@@ -0,0 +1,275 @@
...
+
+def obj_grad(free):
+    grad = np.zeros_like(free)
+    grad[-1] = 1.0
+    return grad
+
+# %%
+# If a solution exists, backend = 'numpy' is better as it sets up ``Problem``
+# much faster. If no solution is available, use 'cython' as the backend, as it
+# is faster for solving the problem.
```
moorepants, line 143: This probably isn't true given that the equations of motion are so simple. Most time is spent in ipopt not in evaluating the equations of motion.

Peter230655: > This probably isn't true given that the equations of motion are so simple. Most time is spent in ipopt not in evaluating the equations of motion.

In my tests, it iterated 5000 to 10000 times to find a solution. I also thought the eoms are so simple, but iteration did take  time.

moorepants: Yes, but there are two things that cost time in the iterations 1) evaluation of the equations of motion, 2) interior point method in ipopt. I think this problem is slow because of 2) not 1). So cython is unlikely to have much effect.

Peter230655: > Yes, but there are two things that cost time in the iterations 1) evaluation of the equations of motion, 2) interior point method in ipopt. I think this problem is slow because of 2) not 1). So cython is unlikely to have much effect.

I do not know the inner workings.
Let me run some tests and report.
Did you notice that I used flake8? :-)  :-)

examples-gallery/beginner/plot_brachistochrone.py
```diff
@@ -0,0 +1,275 @@
...
+    # Take the given solution
+    solution = np.loadtxt(fname)
+else:
+    # Solve the problem.
+    initial_guess = np.random.randn(prob.num_free) * 0.1
+    prob.add_option('max_iter', 15000)
+    for _ in range(10):
+        solution, info = prob.solve(initial_guess)
+        initial_guess = solution
+        print(info['status_msg'])
```
moorepants, line 177: I think a loop like this is an antipattern. If it doesn't converge after 15000 iterations why don't you simply set the max iteration higher? Also, you don't warm start in the next solve in the loop, so you don't have that benefit of the prior optimization descent parameters.

examples-gallery/beginner/plot_brachistochrone.py
```diff
@@ -0,0 +1,275 @@
...
+beta = me.dynamicsymbols('beta')
+
+A.orient_axis(N, beta, N.z)
+
+P.set_pos(O, x*N.x + y*N.y)
+P.set_vel(N, ux*N.x + uy*N.y)
+
+body = [me.Particle('body', P, m)]
+forces = [(P, -m*g*N.y)]
+speed_constr = sm.Matrix([P.vel((N)).dot(A.y)])
```
moorepants, line 84: Isn't this integrable and thus a holonomic constraint? If so, I think it would be better to make a DAE with the holonomic constraint instead of the speed constraint only, as it will suffer from drift in integration.

examples-gallery/beginner/plot_brachistochrone.py
Peter230655: > I think a loop like this is an antipattern. If it doesn't converge after 15000 iterations why don't you simply set the max iteration higher? Also, you don't warm start in the next solve in the loop, so you don't have that benefit of the prior optimization descent parameters.

I ran a trial, I closed the program after each trial to avoid 'caching':
backend='numpy':  82 sec running, 10,000 iterations
backend ='cython'  32 sec running    6,000 iterations 

I have observed, that sometimes, increasing max_iter is helpful, sometimes looping is better, no idea why.
Here max_iter is better, I run the loop only once - but forgot to correct it.

examples-gallery/beginner/plot_brachistochrone.py
Peter230655: > Isn't this integrable and thus a holonomic constraint? If so, I think it would be better to make a DAE with the holonomic constraint instead of the speed constraint only, as it will suffer from drift in integration.

I think, it is. I tried, but not very concentrtatedly so and it did not work,
I will try again.

examples-gallery/beginner/plot_brachistochrone.py
moorepants: Evaluating the objective and constraints are sped up by cython always. For small equations of motion, the evaluation time difference between cython and numpy is very small, but if you have thousands and thousands of iterations that time difference is summed. But each iteration time is the sum of the objective and constraint evaluation + how much time Ipopt uses to execute the interior point method. Every problem I have solved so far seems to show that the interior point method evaluation is slower than the objective and constraint evaluation. This problem seems to be an exception, where the time spent on the two things in the iteration are someone equal. So, the overall time is decreased with cython significantly because of the high iteration count and the equal time spent on the two aspects of an iteration. Interesting. Also interesting is that it takes so many iterations. I don't know why.

moorepants: > I have observed, that sometimes, increasing max_iter is helpful, sometimes looping is better, no idea why.

When you do the looping you don't warm start the solution, so you are starting the optimization from scratch, just with a different initial guess. This will give different results than not looping as well as looping with a warm start.

Peter230655: > > I have observed, that sometimes, increasing max_iter is helpful, sometimes looping is better, no idea why.
> 
> When you do the looping you don't warm start the solution, so you are starting the optimization from scratch, just with a different initial guess. This will give different results than not looping as well as looping with a warm start.

how would I loop with a warm start?
What I report a just observations.

moorepants: Warm start is in the cyipopt and ipopt documentation.

examples-gallery/beginner/plot_brachistochrone.py
moorepants: >  Did you notice that I used flake8? :-) :-)

Yes, nice. Thanks.

examples-gallery/beginner/plot_brachistochrone.py
moorepants: Actually, no this is a nonholonomic constraint. It is the same as the Chaplygin sleigh in my book!

examples-gallery/beginner/plot_brachistochrone.py
Peter230655: > > Did you notice that I used flake8? :-) :-)
> 
> Yes, nice. Thanks.

Thank YOU! On **very** rare occasions, if you say something like tihs is simple, it is simple for me! :-)

moorepants:
This is looking good. Nice example. I'm very surprised it takes so many iterations.

Peter230655:
> This is looking good. Nice example. I'm very surprised it takes so many iterations.

Thanks!
It is because of these surprises that it is so interesting!
Getting the right brachistochrone seems to depend critically on the right initial guess you give to ``root```
Presumably, this also makes it difficult for opty.

You did notice that the gif does not work. I do not want to touch this, I messed up all gifs before.

examples-gallery/beginner/plot_brachistochrone.py
Peter230655: > Actually, no this is a nonholonomic constraint. It is the same as the Chaplygin sleigh in my book!

Presumably because beta.diff(t) != const. ?
Should I add the proof of why these are not holonomic constraints from your book inot this simulation, or would this be too much?

moorepants: You don't need a proof but maybe it is worth saying that this is a nonholonomic constraint.

examples-gallery/beginner/plot_brachistochrone.py
```diff
@@ -0,0 +1,275 @@
...
+kane = me.KanesMethod(
+            N, q_ind=[x, y],
+            u_ind=[ux],
+            u_dependent=[uy],
+            kd_eqs=kd,
+            velocity_constraints=speed_constr)
+fr, frstar = kane.kanes_equations(body, forces)
+
+eom = kd.col_join(fr + frstar)
+eom = eom.col_join(speed_constr)
```
moorepants, line 96: I think that it is redundant to add this constraint, it  has already been solved and incorporated into the equations of motion.

Peter230655: If I do not add speed_constr to the eoms, i do not have enough eoms for opty (?)

moorepants: you can delete x.diff() - ux I think, because it is eliminated.

Peter230655: I will try - but would this be a good example for beginners? 
When you start with opty and you know sympy mechanics, you know that for every speed constraint you must add it to the eoms for opty. For every config_constr you must add config_constr.diff(t)  to sympy mechanics and add config_constr to the eoms for opty.
I somehow see opty and sympy mechanics as closely related - but maybe wrong?

moorepants: > you know that for every speed constraint you must add it to the eoms for opty.

that's not true with Kane's method. All speed constraints are solved and included in Fr + Fr*.

moorepants: I just tried this and it seems to solve the problem in like 12 iterations if you remove the redundant constraint.

examples-gallery/beginner/plot_brachistochrone.py
moorepants: It may be a red hearing, I couldn't reproduce.

moorepants: I think these are the equations of motion with the dependent speed uy eliminated:

```
Matrix([
[                                                                              u_x(t) - Derivative(x(t), t)],
[                                                                 u_x(t)*tan(beta(t)) - Derivative(y(t), t)],
[-m*(g*tan(beta(t)) + omega(t)*u_x(t)*sin(beta(t))/cos(beta(t))**3 + Derivative(u_x(t), t)/cos(beta(t))**2)],
[                                                                         omega(t) - Derivative(beta(t), t)]])
```

Peter230655: > I think these are the equations of motion with the dependent speed uy eliminated:
> 
> ```
> Matrix([
> [                                                                              u_x(t) - Derivative(x(t), t)],
> [                                                                 u_x(t)*tan(beta(t)) - Derivative(y(t), t)],
> [-m*(g*tan(beta(t)) + omega(t)*u_x(t)*sin(beta(t))/cos(beta(t))**3 + Derivative(u_x(t), t)/cos(beta(t))**2)],
> [                                                                         omega(t) - Derivative(beta(t), t)]])
> ```

How would you do this?
- eliminate uy = y.diff(t)
- in the balance eoms replace uy by y.diff(t) and uy.diff(t) by y.diff(t, 2)

This is a very neat trick! Is it suitable for a beginner's example?

Peter230655: > It may be a red hearing, I couldn't reproduce.

Not sure what you mean.

examples-gallery/beginner/plot_brachistochrone.py
Peter230655: > > you know that for every speed constraint you must add it to the eoms for opty.
> 
> that's not true with Kane's method. All speed constraints are solved and included in Fr + Fr*.

I remember: the famous Ars matrix or even Ars_tilde matrix (I messed around with the a while ago)

examples-gallery/beginner/plot_brachistochrone.py
moorepants: I played around and it was solving very fast and then it wasn't. I can't reproduce it.

examples-gallery/beginner/plot_brachistochrone.py
moorepants: KanesMethod applies Ars but does not try to eliminate the dependent speeds.

examples-gallery/beginner/plot_brachistochrone.py
Peter230655: > You don't need a proof but maybe it is worth saying that this is a nonholonomic constraint.

I wonder whether they are truly nonholonomic.
I played around according to your lecxture - but likely made a mistake somewhere.

moorepants: The constraint is exactly the constraint I have in my lecture. As you've posed the problem, with beta(t) as a time varying variable, I think it ends up being a nonholonomic constraint. But this is due to construction. Of course moving on a fixed path is holonomic, but you haven't constructed the curve as a fixed path.

Peter230655: > The constraint is exactly the constraint I have in my lecture. As you've posed the problem, with beta(t) as a time varying variable, I think it ends up being a nonholonomic constraint. But this is due to construction. Of course moving on a fixed path is holonomic, but you haven't constructed the curve as a fixed path.

EXACTLY the same! I should think before I talk (or write).

Peter230655:
Changed some things we discussed.
I did not switch to the ``uy elimination``. If you fell this would be o.k. for a beginner's example, I can easily do it.
The **gif** does not work. I did not touch anything in that area for fear to mess up all the gifs again.

examples-gallery/beginner/plot_brachistochrone.py
Peter230655: It seems, you are trading ``uy = y.diff(t)`` for ``omega = beta.diff(t)``
Of course one never knows with opty - but why would opty prefer this trade?

moorepants: Yes, but that only resulted from the fact that beta and beta' end up in the final equations of motion.

Peter230655: I ran tests comparing the 'standard' eoms to your 'hack'

Standard eom / 'cython': 
49. - 52 sec, around 10,000 to 11,000 iterations
Standard eom / 'numpy:
91 - 100 sec, around 12,000 iterations

Hack eom / 'cython'
56 - 150 sec, 15,000 + iterations, no useful result
Hack eom / 'numpy'
100 - 120 sec, 6,000 - 15,000+ iterations no useful result.

As I could see no advantage with the hack, I did not play around, trying to bound the state variables much.
I bounded omega, but this also did not really do any good.

My summary feel is, that this is not an easy problem numerically - if it were, it would likely not be in John Betts' book.

moorepants: Are you calling my formulation that eliminates the dependent speed entirely a hack?

I feel like we may be solving a harder problem by formulating this as a nonholonomic constraint when it seems that it should not have such a constraint.

Peter230655:
> This is looking good. Nice example. I'm very surprised it takes so many iterations.

examples-gallery/beginner/plot_brachistochrone.py
moorepants: Interesting. That does seem pertinent to what you are doing here. Crazy that this problem is so hard to solve.

Peter230655:
Thanks! :-)
- around line 165 you changed the indentation of prob = Problem. flake8 did not object to my indentations? Of course, optically yours looks nicer
- Same around line 287 animation = FuncAnimation.. flake8 did not object.
- you removed the feference to your lecture. Is your lecture not public?

moorepants:
> around line 165 you changed the indentation of prob = Problem. flake8 did not object to my indentations? Of course, optically yours looks nicer

You had indentations of 8 or 12 spaces, I changed them to 4, no need to indent more

I don't think my lecture is the best reference. Wikipedia tells about the sleigh.

Peter230655:
> > around line 165 you changed the indentation of prob = Problem. flake8 did not object to my indentations? Of course, optically yours looks nicer
> 
> You had indentations of 8 or 12 spaces, I changed them to 4, no need to indent more
> 
> I don't think my lecture is the best reference. Wikipedia tells about the sleigh.

Clear!
It still does not show the gif.
I am afraid to touch this area, you recall I messed up all the gifs at one time.
(not important, not the most interesting gif anyway)

moorepants:
The gif seems to display as the thumbnail when I open the documentation.

Peter230655:
> The gif seems to display as the thumbnail when I open the documentation.

I did not see it move, probably my old eyes.

Peter230655:
> Note that I don't see the gif thumbnails in the local build, for some reason they only seem to show up consistently in the readthedocs builds.

Clear, I only looked at the local build.

PR #411, opened 2025-03-20
brachistochrone with interference

Peter230655:
This is essentially the brachistochrone example in examples-gallery/beginner.
I added an interference, as suggested by John Betts in his book.

[Betts_page_206.pdf](https://github.com/user-attachments/files/19372701/Betts_page_206.pdf)

The simulation takes 20,000 to 45,000 iterations to find a valid solution, depending on where the interference is located.
Interestingly, if I move the 'interfering triangle' out of the way, by, say, increasing h_shift, so it does not intersect the brachistochrone anymore, it takes the most interations to find a valid solution. Of course no idea why this is so.

I do not think, I have ever had a simulation which used so many iterations to reach a **valid** solution.
I have had simulations, which would run out of iterations, but no matter how many I set, they did not find a solution.

I am not suggesting this should be added to examples-gallery, it is too similar to the example already there. I just wanted you to see it.

NB: This also shows the animated gif in the local build, and I did not touch anything there.


moorepants:

Peter230655:
No, there is not. As Betts considered it to be very difficult for optimization when having this interfereing line, I wanted to try it.
Iteresting, but not so for examples-gallery, was the fact that it needed a huge amount of iterations when the line did not interfere, but of course was in the eoms.

Peter230655:
Moved to pst-notebooks

PR #417, opened 2025-03-25
Light diffraction

Peter230655:
I try to simulate light diffraction, which, as I recall is caused the the speed of light being different in different media.
So, I let a particle travel from A to B where in different areas, its speed is limited to different values.
The force applied is large enough so that acceleration to the respective max. speed takes practically no time.
The various speeds are realized using a differentiable hump function.
This can cause undesired effects at the edges, which I solved by bounding $u_x$, the speed component in x direction.
It is for this reason, that I sauggest it for intermediate, not for beginner.

flake8 showed no problems. :-)


moorepants:
I read this but do not understand how this example relates to light diffraction. It seems like a particle with bounds on its speed at certain locations. I don't see what physics that is modeling.

Also, there are other examples that use smoothed approximations of discontinous functions. So what does this example add that isn't already in the gallery? I can see that the smooth function is more prominent and the only special thing in this example versus the others may have the smooth function buried in with other things.

Peter230655:
Basic idea was this:
A photon has different speeds in different media.
When it moves from A to B and passes through different media it takes a path which minimizes the time it takes to go from A to B.
I wanted to simulate this, as a second example of a well know variational problem in physics. (the first of course is the brachistochrone, which you suggested).

To model the different media I used a differentiable hump function, used in other examples, too.
However, unless I forced $u_x$ to have a minimum speed, it did not behave as expected near the edeges of two media.
I assume, this behaviour is because the hump function is not an ideal hump, but must be differentiable.

Other than the fact that it worked. I though this playing around with bounds in a case like this might be interesting.
(An optimization, as per my experience with opty never works the firast time, one has to play around with the parameters, this is one more example to this effect,)

moorepants:
More explanation is needed in the example to explain this. And if this adds nothing new to the gallery, then we should move it to pst-notebooks.

Peter230655:
No, nothing new.

Peter230655:
Moved to pst-notebooks

PR #419, opened 2025-03-25
Add option to plot bounds on the trajectories plots.

Peter230655:
Added an optional kwarg ``plot_bounds`` If true, bounds will be plotted in the plots of the respective trajectory, unless they are np.inf, or -np.inf Takes care of Q&A #414 
I tested with around 15 simulations.

### DIFF
--- a/examples-gallery/intermediate/plot_parallel_park.py
+++ b/examples-gallery/intermediate/plot_parallel_park.py
@@ -192,7 +192,7 @@ print(info['obj_val'])
 
 # %%
 # Plot the optimal state and input trajectories.
-_ = prob.plot_trajectories(solution)
+_ = prob.plot_trajectories(solution, show_bounds=True)
 
 # %%
 # Plot the constraint violations.
--- a/examples-gallery/intermediate/plot_parallel_park.py
+++ b/examples-gallery/intermediate/plot_parallel_park.py
@@ -160,7 +160,7 @@ instance_constraints = (
 bounds = {
     delta: (np.deg2rad(-45.0), np.deg2rad(45.0)),
     T: (-100.0, 100.0),
-    F: (-10000.0, 10000.0),
+    F: (-10000.0, np.inf),
 }
 
 # %%
--- a/examples-gallery/intermediate/plot_parallel_park.py
+++ b/examples-gallery/intermediate/plot_parallel_park.py
@@ -159,8 +159,8 @@ instance_constraints = (
 # Add some physical limits to some variables.
 bounds = {
     delta: (np.deg2rad(-45.0), np.deg2rad(45.0)),
-    T: (-100.0, 100.0),
-    F: (-10000.0, np.inf),
+    T: (-39.0, 41.0),
+    F: (-177.0, 252.0),
 }
 
 # %%
--- a/examples-gallery/intermediate/plot_parallel_park.py
+++ b/examples-gallery/intermediate/plot_parallel_park.py
@@ -159,8 +159,8 @@ instance_constraints = (
 # Add some physical limits to some variables.
 bounds = {
     delta: (np.deg2rad(-45.0), np.deg2rad(45.0)),
-    T: (-39.0, 41.0),
-    F: (-177.0, 252.0),
+    T: (-50.0, 50.0),
+    F: (-300.0, 300.0),
 }
 
 # %%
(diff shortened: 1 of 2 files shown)


Peter230655:
Changed the kward.
[plot_parallel_park_show_bonds.pdf](https://github.com/user-attachments/files/19446653/plot_parallel_park_show_bonds.pdf)
This is a picture.

moorepants:
Yes, please add it to an example, as that is a "unit test" for all intents and purposes.

moorepants:
The default color cycle in matplotlib is `C0`, `C1`, ... So if the lines are `C0` (blue) then you can use `C1` for the bounds.

Peter230655:
Used C1.
Added ``plot_parallel_park.py`` to this PR.

examples-gallery/intermediate/plot_parallel_park.py
```diff
@@ -160,7 +160,7 @@
 bounds = {
     delta: (np.deg2rad(-45.0), np.deg2rad(45.0)),
     T: (-100.0, 100.0),
-    F: (-10000.0, 10000.0),
+    F: (-10000.0, np.inf),
```
moorepants, line 163: I think you should set the bounds to new values that are just above the max of the solutions, so that the solutions are reduced to straight lines or it is hard to see the variation in the line.

examples-gallery/intermediate/plot_parallel_park.py
```diff
@@ -192,7 +192,7 @@
 
 # %%
 # Plot the optimal state and input trajectories.
-_ = prob.plot_trajectories(solution)
+_ = prob.plot_trajectories(solution, show_bounds=True)
```
moorepants, line 195: The plot looks good other than the above comment.

examples-gallery/intermediate/plot_parallel_park.py
Peter230655: I will try now.

moorepants: yes, i missed "not"

opty/direct_collocation.py
```diff
@@ -555,6 +558,15 @@ def plot_trajectories(self, vector, axes=None):
         for ax, traj, symbol in zip(axes, trajectories, traj_syms):
             ax.plot(time, traj)
             ax.set_ylabel(sm.latex(symbol, mode='inline'))
+
+            if self.bounds is not None and show_bounds:
+                if symbol in self.bounds.keys():
+                    if self.bounds[symbol][0] not in [-np.inf, np.inf]:
```
moorepants, line 564: I don't think this is a valid check. Comparing something to infinity must use `np.isinf()`. Also I don't think you even need the check. I don't think matplotlib will plot infinity. I till just ignore, so we don't need this check here.

opty/direct_collocation.py
```diff
@@ -555,6 +558,15 @@ def plot_trajectories(self, vector, axes=None):
...
+                        ax.axhline(self.bounds[symbol][0], color='C1',
+                                lw=1.0, linestyle='--')
```
moorepants, line 566: bad indentation

opty/direct_collocation.py
```diff
@@ -555,6 +558,15 @@ def plot_trajectories(self, vector, axes=None):
...
+                    if self.bounds[symbol][1] not in [-np.inf, np.inf]:
+                        ax.axhline(self.bounds[symbol][1], color='C1',
+                            lw=1.0, linestyle='--')
```
moorepants, line 569: bad indentation

opty/direct_collocation.py
Peter230655: This check is not needed, I just tired.

opty/direct_collocation.py
Peter230655: I doubt I will ever fully get this. :-(

moorepants: There are two options:

```python
function(
    a,
    b,
    c,
)
```

or

```python
function(veryveryveryverylongthing, anotherverylongthing,
         nextlinelongthing)
```

moorepants: actually three:

```python
function_name_that_is_super_long(
   unbelievablylonglonglonglonglonglonglonglonglongarg, b, c)
```

Peter230655:
I made the changes as you suggested, but I am not allowed to push them...

PR #422, opened 2025-03-26
Use `time_vector()` where possible in the example gallery.

### LINKED ISSUE #415: Use `Problem.time_vector()` in the examples.
We added this new method but few to none of the examples use it. All of them should use it.

moorepants:
Fixes #415.

### DIFF
--- a/examples-gallery/advanced/plot_sit_to_stand.py
+++ b/examples-gallery/advanced/plot_sit_to_stand.py
@@ -258,7 +258,7 @@ _ = prob.plot_constraint_violations(solution)
 # Use symmeplot to make an animation of the motion.
 # sphinx_gallery_thumbnail_number = 3
 xs, rs, _, h_val = prob.parse_free(solution)
-times = np.linspace(0.0, (num_nodes - 1)*h_val, num=num_nodes)
+times = prob.time_vector(solution=solution)
 
 
 def animate(fname='animation.gif'):
--- a/examples-gallery/advanced/plot_sit_to_stand.py
+++ b/examples-gallery/advanced/plot_sit_to_stand.py
@@ -258,7 +258,7 @@ _ = prob.plot_constraint_violations(solution)
 # Use symmeplot to make an animation of the motion.
 # sphinx_gallery_thumbnail_number = 3
 xs, rs, _, h_val = prob.parse_free(solution)
-times = prob.time_vector(solution=solution)
+times = np.linspace(0.0, (num_nodes - 1)*h_val, num=num_nodes)
 
 
 def animate(fname='animation.gif'):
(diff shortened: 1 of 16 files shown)


moorepants:
Getting this only on readthedocs:

```
        ../examples-gallery/advanced/plot_sit_to_stand.py failed leaving traceback:
    
        Traceback (most recent call last):
          File "/home/docs/checkouts/readthedocs.org/user_builds/opty/checkouts/422/examples-gallery/advanced/plot_sit_to_stand.py", line 344, in <module>
            animation = animate('human-sit-to-stand.gif')
          File "/home/docs/checkouts/readthedocs.org/user_builds/opty/checkouts/422/examples-gallery/advanced/plot_sit_to_stand.py", line 324, in animate
            sim_data = np.vstack((
                xs,  # q, u shape(2n, N)
            ...<3 lines>...
                          len(times), axis=1),  # p, shape(r, N)
            ))
          File "/home/docs/checkouts/readthedocs.org/user_builds/opty/conda/422/lib/python3.13/site-packages/numpy/_core/shape_base.py", line 292, in vstack
            return _nx.concatenate(arrs, 0, dtype=dtype, casting=casting)
                   ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        ValueError: all the input array dimensions except for the concatenation axis must match exactly, but along dimension 1, the array at index 0 has size 60 and the array at index 1 has size 61
```

examples-gallery/advanced/plot_sit_to_stand.py
```diff
@@ -258,7 +258,7 @@ def obj_grad(free):
 # Use symmeplot to make an animation of the motion.
 # sphinx_gallery_thumbnail_number = 3
 xs, rs, _, h_val = prob.parse_free(solution)
-times = np.linspace(0.0, (num_nodes - 1)*h_val, num=num_nodes)
+times = prob.time_vector(solution=solution)
```
moorepants: This change seems to cause a failure. Not sure why.

moorepants: Uh.. then time_vector() shouldn't be using arange. If we knew from the other issue that arange isn't what we want, why use it in time_vector?

moorepants: arange is indeterminate on the length of what it returns depending on the floating point round off of the step, right? I don't think we want that.

moorepants: Also, time_vector() should always return an array the same length as the number of nodes.

Peter230655: There was a discussion about this aound the time I wrote this method. I forgot the details. (I had hardly known about np.arange until this discussion came up)

As I recall, np.arange sometimes misses the last point in time.
Let me look again, maybe I will change it to np.linspace.

Peter230655: I will change it to np.linspace tomorrow.

Peter230655: Now I recall: the discussion I refered to above was PR #283

PR #423, opened 2025-03-27
time_vector. Changed np.arange to np.linspace

Peter230655:
Changed np.arange to np.linspace
I tested with sit_to_stand and did not get the error gotten before (When I run it I always get an overflow warining in the animate part, I got this before every time I run it on my PC)

### DIFF
--- a/opty/tests/test_direct_collocation.py
+++ b/opty/tests/test_direct_collocation.py
@@ -1976,8 +1976,8 @@ def test_time_vector():
         time_symbol=t,
         backend='numpy'
     )
-    expected_time_vector = np.arange(t0, t0 + num_nodes*interval_value,
-                        interval_value)
+    expected_time_vector = np.linspace(t0, t0 + (num_nodes-1)*interval_value,
+                                       num_nodes)
     time_vector = prob.time_vector(start_time=t0)
     assert np.allclose(time_vector, expected_time_vector)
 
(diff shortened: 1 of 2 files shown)


Peter230655:
``test_time_vector`` also had np.arange(..). Changed to np.linspace(..)

opty/tests/test_direct_collocation.py
```diff
@@ -2018,14 +2018,14 @@ def obj_grad(free):
...
+                                       num_nodes)
     assert np.allclose(time_vector, expected_time_vector)
 
     # final time > initial time
     solution[-1] = 1.e-75
-    expected_time_vector = np.arange(t0, t0 + num_nodes*solution[-1],
-                solution[-1])
+    expected_time_vector = np.linspace(t0, t0 + (num_nodes-1)*solution[-1],
+                                       num_nodes)
     with raises(ValueError):
```
moorepants, line 2029: This would be better tested if the example has only 10 nodes or so and you manually write what the array should be. We also clearly need a test that is subject to the rounding error issue we see with arange() so that we prove the current implementation does not fall prey to it.

moorepants: Right now you are typing the exact same code in the test as you are in the method, so it isn't really a strong test (other than maybe a regression test for some cases).

Peter230655: I realized that my test was not really a test at all, but I had no better idea.
I will try to find an example, where arange and linspace differ.

PR #425, opened 2025-03-30
multiphase collision.

Peter230655:
This is not yet a fully 'beautified' version.
I pushed it for you to see, if this is along the lines you thought.
In principle, your idea of adding DEAs to the eoms seems to work in this case.
The convergence is not trivial: I had ot play around so I could understand how to make it converge.


moorepants:
I scanned through it, but I don't easily understand what you have done.

moorepants:
One comment is that the solution I'm (likely) interested in would not involve any approximations of non-continuous functions. The reason to break the simulation into the leftward and rightward motion is to avoid such functions because the motion is continuous in each of those periods.

Peter230655:
Right now, I am working on a better explanation, will push it when finished.
Essentially what I did was this:
- run your program, except that instead of having the collision right at the midpoint of the journey I moved it to around 2/3 of the journey. This is only to create a good initial guess.
- Run the 'main program' with this initial guess. Here I add two additional eoms which enforce that x_l = x_r when x_l reaches the wall and v_l = e*v_r when x_l reaches the wall.

Peter230655:
I added explanations and 'beautified' it.
As I now understand, it does not solve 333 as you wanted it solved. 
If it does not add anything to examples-gallery, I will colse it and push it to pst-notebooks.

moorepants:
We have enough examples where you have shown how to make a smooth step or hump function. I'm not sure what this adds to the examples.

moorepants:
Reopen if you think this bring something new.

Peter230655:
Better delete. The idea was to 'switch' eoms along the path. It worked for this simple example, but totally failed on a more difficult one.

Issue #204, opened 2024-08-04
Reduce time it takes to run the examples


moorepants:
Here is the current timing from a clean build:

```
computation time summary:
    - ../examples-gallery/plot_pendulum_swing_up_variable_duration.py:   110.06 sec   0.0 MB
    - ../examples-gallery/plot_drone.py:                                  75.58 sec   0.0 MB
    - ../examples-gallery/plot_two_link_pendulum_on_a_cart.py:            75.11 sec   0.0 MB
    - ../examples-gallery/plot_sliding_block.py:                          59.82 sec   0.0 MB
    - ../examples-gallery/plot_parallel_park.py:                          58.78 sec   0.0 MB
    - ../examples-gallery/plot_pendulum_swing_up_fixed_duration.py:       56.85 sec   0.0 MB
    - ../examples-gallery/plot_one_legged_time_trial.py:                  44.69 sec   0.0 MB
    - ../examples-gallery/plot_vyasarayani.py:                            11.22 sec   0.0 MB
    - ../examples-gallery/plot_betts2003.py:                              10.89 sec   0.0 MB
```

Reducing the time it takes to create animations is an easy adjusted, for example. We can safely skip generating a frame for every time instance by skipping frames with a minor adjustment to the code. Here is an example:

https://github.com/csu-hmc/opty/blob/master/examples-gallery/plot_drone.py#L299-L300

moorepants:
Another way to reduce time is to give better initial guesses. We can even store a solution and give that as a guess.

Peter230655:
> Another way to reduce time is to give better initial guesses. We can even store a solution and give that as a guess.

How can I store a solution, so it does not get lost when the program is closed?
I thought about this often, but never came up with an idea... :-(

moorepants:
You can save any value to a txt file with many different methods. NumPy has builtin ways to do this: https://numpy.org/doc/stable/reference/routines.io.html

Peter230655:
> You can save any value to a txt file with many different methods. NumPy has builtin ways to do this: https://numpy.org/doc/stable/reference/routines.io.html

Amazing how this works - and it is two lines of code initially, no additional line in the final simulation. (After your hint, chatGPT told me the lines) I once modeled a drone with propellers, the one that wiggled and it cut down the simulation tine from over one minute to 16 sec. - still wiggles, though :-)

moorepants:
Here is the output from a recent build on the readthedocs system, which is slower than running it locally:

```
[...]

PR #428, opened 2025-04-02
one legged time trial. Towards #204 for this simulation

Peter230655:
- added a solution
- backend='numpy'
- reduced frames with the animation.

Reduced running time on my PC from around 32.5 sec to 16.5 sec
This simulation has so many things, I did not dare formulating an ``Objective`` to go towards #348


Peter230655:
Missed to add / push the solution

moorepants:
The formatting is broken:

image

Peter230655:
Should this be closed? All it does is reduce the execution time by around 50%.
I think our opty PR list should not look like sympy's :-)

moorepants:
I recall not favoring some of the changes here, but haven't had time to review it again.

Peter230655:
I vaguely recall the following: If the existing solution is used, I do not solve the problem again, so e.g. the objective value cannot be plotted. 
As my changes only speed up the execution, this may not be important anymore.

PR #430, opened 2025-04-05
estimate time shift

Peter230655:
I wanted to play around with #294 
Unfortunately, I misunderstood the issue, so this does not solve #294

What it does is this:
It estimates $\delta$ (and on the way also F and $\omega$) if the driving force is e.g. $F\sin(\omega(t - \delta))$
(of course sin could be replace by another given function)

### DIFF
--- a/examples-gallery/beginner/plot_estimate_time_delay.py
+++ b/examples-gallery/beginner/plot_estimate_time_delay.py
@@ -1,0 +1,264 @@
+# %%
+r"""
+Time Delay Estimation
+=====================
+
+Objective
+---------
+
+- Show how a time delay in a driving force of a mechanical system may be
+  estimated from noisy measurements by introducing a state variable which
+  mimicks the time.
+
+
+Introduction
+------------
+
+A simple pendulum is driven by a torque of the form:
+:math:`F \sin(\omega (t - \delta))`, and :math:`F, \omega, \tau`
+are to be erstimated based on noisy measurements of the angle.
+
+The driving force is set to zero for :math:`t < \delta`.
+
+**States**
+
+- :math:`q` : angle of the pendulum [rad]
+- :math:`u` : angular velocity of the pendulum [rad/s]
+- :math:`T(t)` : variable which mimicks the time t [s]
+
+
+**Known parameters**
+
+- :math:`m` : mass of the pendulum [kg]
+- :math:`g` : gravity [m/s^2]
+- :math:`l` : length of the pendulum [m]
+- :math:`I_{zz}` : inertia of the pendulum [kg*m^2]
+- :math:`steep` : steepness of the differentiable step function [1/s]
+
+
+**Unknown parameters**
+
+- :math:`F` : strength of the driving torque [Nm]
+- :math:`\omega` : frequency of the driving torque [rad/s]
+- :math:`\delta` : time delay of the driving torque [s]
+
+"""
+import numpy as np
+import sympy as sm
+from scipy.integrate import solve_ivp
+import sympy.physics.mechanics as me
+from opty import Problem
+from opty.utils import MathJaxRepr
+import matplotlib.pyplot as plt
+
+# %%
+# Set Up the System
+# -----------------
+
+N, A = sm.symbols('N A', cls=me.ReferenceFrame)
+O, P = sm.symbols('O P', cls=me.Point)
+q, u = me.dynamicsymbols('q u')
+O.set_vel(N, 0)
+t = me.dynamicsymbols._t
+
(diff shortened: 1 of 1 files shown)


examples-gallery/beginner/plot_estimate_time_delay.py
```diff
@@ -0,0 +1,264 @@
...
+A.orient_axis(N, q, N.z)
+A.set_ang_vel(N, u*N.z)
+
+P. set_pos(O, -l*A.y)
+P.v2pt_theory(O, N, A)
+
+inert = me.inertia(A, 0, 0, iZZ)
+bodies = [me.RigidBody('body', P, A, m, (inert, P))]
+# Driving force set to zero for t < delta
+torque = F * sm.sin(omega*(t - delta)) * sm.Heaviside(t - delta)
```
moorepants, line 76: This is equivalent to a phase shift in the sin(): `sin(w*t + phi)`.

opty does not yet support having time explicit in the equations  of motion and we have shown how to make time explicit by introducing a variable `T(t)` in another example. So I'm not sure this adds any new tricks for a reader of the documentation. Let me know otherwise.

Peter230655: I agree, nothing new here.
I will think of an example with explicit time, maybe Betts has one. My ``rolling ball on a spinning disc`` had explicit time, but I used your trick there with a known trajectory, as it made more sense.

Peter230655:
Now I ran into something odd:

I wrote $torque = F \cdot T(t)^2 \cdot sin(\omega(t - \delta)  \cdot step_diff(t, \delta, steep)$

Peter230655:
I used $torque = F \cdot T(t)°2 \cdot \sin(\omega(t - \delta)) \cdot stepdiff(t, \delta, steep)$
I played around some more and I found this:
- if I fix $\delta$ and leave ``t`` in the functions, 'numpy' will set up ``Problem``, an error comes with ``solve``. In this situation 'cython' already raises an error in ``Problem``
- if I leave $\delta$ an unknown parameter, both backends raise an error in ``Problem`` already.
- if I replace ``t`` with ``T(t)`` both backends get the result.

moorepants:
Are you saying that `omega(T(t) - delta)` works if omega is a state (or input)?

moorepants:
The thing we are trying to solve with Christoph's PR is that an unknown state or input variable is a delayed function of time. It still is not clear if you are addressing the same issue or something else.

Peter230655:
No, this is not the same issue I am addressing. 
I re-opended my PR because opty did **not** accept something like ``sin(omega(t - delta)``, with omega and delta unknown parameters, but it does accept ``sin(omega(T(t) - delta))`` without any problems. I thought, this may be of independet interest.

I have not tried anything like state_variable(omeaga(t - delta)).

moorepants:
If `omega` is an unknown parameter, that is a constant. I don't even know what it means to do something like `omega(...)`. I didn't think that was defined on a symbol.

Peter230655:
Fair enough! An unknown parameter of course is an unknown constant.
I should have written: ``sin(omega*(t - delta)``

moorepants:
This example shows how to use `T(t)` for explicit time: https://opty.readthedocs.io/stable/examples/beginner/plot_betts2003.html

Peter230655:
I mentioned it in my comment. This where I got the idea for my spinning disc.
But is using a state variable and setting T(t).diff(t) = 1.0 in the eoms not still simpler?

But I feel my second observation, if correct (all my tests confirmed it) is more interesting, and does seem to show something new.
NB: Both 'numpy' and 'cython' fail if ``t`` instead of ``T(t)`` is used in the sin function, but the error messages are a bit different,

moorepants:
If you have a workaround to get expclit time working with variable time step, then we should show that. Maybe even in place of the idea shown in https://opty.readthedocs.io/stable/examples/beginner/plot_betts2003.html

> However I feel my second observation, if correct (all my tests confirmed it) is more interesting, and does seem to show something new.

I still don't understand what you are showing.

>  NB: Both 'numpy' and 'cython' fail if t instead of T(t) is used in the sin function or the step function, but the error messages are a bit different,

Testing whether these  backends fail is not that pertinent. I know they will fail because they were not designed to do what you want them to do. The failure is very likely the same reason even if the raised error occurs at a different line in the code.

Peter230655:
I looked at
https://opty.readthedocs.io/stable/examples/beginner/plot_betts2003.html#sphx-glr-examples-beginner-plot-betts2003-py
more closely.
What I want to show here is already there, except that T(t) is realized as a known_input_trajectory rather than as a state_variable.
So, really nothing new here, just expanded a little bit.
I will push this to pst-notebooks and close it.

PR #432, opened 2025-04-06
two link pendulum on a cart. Added explicit time dependence

Peter230655:
This is the double pendulum on a cart in examples_gallery/intermediate.
It uses a variable time interval.

- I only added **explicit** time dependence in the eoms, and explained how to do it.
- I 'beautified' a little bit.

If this is acceptable, the old example should be deleted, I think


moorepants:
> If this is acceptable, the old example should be deleted, I think

The example is already included in a release of opty, so we won't be deleting it.

Peter230655:
I did not think of this!
Since it is really EXACTLY the same like the existing one, except this explicit time dependence, it makes **no** sense to add it, I think.
I will come up with something for sure.
Should I close this PR, or will you do it?

PR #433, opened 2025-04-11
explicit time with variable duration

Peter230655:
This is uprighting a simple pendulum in minimum time where the torque is explicitly time dependent.
It shows how to circumvent this current opty limitation. 

Additionally (this is present in another simulation already) it shows how to get $\dfrac{d^2}{dt^2}(q(t_f) = 0$ using an additional state variable.

I do not get the ``gif`` to work!


Peter230655:
Now that input trajectories may be given instance constraints, I replaced a 'non-state' state variable with an input trajectory.
Improved comments a bit.
As currently time derivatives of unknown input trajectories do no work properly, T had to remain a state variable.

moorepants:
This example already shows how to include time explicitly; https://opty.readthedocs.io/stable/examples/beginner/plot_betts2003.html

Reopen if you want this in otherwise.

PR #434, opened 2025-04-13
car on racecourse, smooth driving force

Peter230655:
known input trajectories can vary discontinuously.
If for whatever reason, this is not wanted, I show how to make them smooth.

There are a few other things, like incorporating speed constraints, using additional state variables to handle inequalities.
This is not new, shown in other simulations, but there is all of it here, This is why I thought it may belong to _intermediate_.

It also demonstrates that in a loop containing ``Problem.solve(..)`` if known parameters are changed, no need to set up ``Problem`` again. (To get opty to converge, I had to go from a less curvy road to a more curvy one).
It takes about 30 min on my PC to find a solution, so apparently not a very easy problem for opty.
Running time with a given solution about 7.9 sec on my PC.


Peter230655:
Since in the meantime inequalities becam available, I replaced all hacks (about 13 of them) with inequalities.
Here the inequalities seem be be considerably faster the the old hacks.

Peter230655:
I set ``subplots=False`` since until we have a method to handle constraint violations with eom_bounds, this looks nicer, I think.
Tried to clarify some comments.
Reduced fps in the animation to reduce running to on my PC from around 16 sec to around 9 sec.

PR #439, opened 2025-04-24
particle moving through gates

Peter230655:
A particle must pass three gates at predetermined times (integer multiples of h, the variable interval value).
opt should select, there it should pass the gate.
As presently opty does not allow instance constraints of the form $x(t_i) - (a, b)$ I use a simple trick to achieve it.

Convergence was a bit tricky. I had to play around a bit to finally get it.


Peter230655:
This is a subest of PR 441, so I close it.

PR #440, opened 2025-04-25
Enables support for additional path constraints in the EoMs.

### LINKED ISSUE #438: Allow additional constraint equations in addition to the essential ODEs or DAEs
opty's current design assumes that there are n states and n equations of motion and enforces that. For DAEs this means that you have to formulate the equations of motion such that there are no redundant equations. But we should be able to support more general equations. A simple example is a particle moving on a plane with a path constraint (for example stay on a circle):

$$
\begin{align}
m \dot{v}_x = F_x \\
m \dot{v}_y = F_y \\
\dot{x} = v_x \\
\dot{y} = v_y \\
x^2 + y^2 = r^2
\end{align}
$$

Here there are 4 states but 5 valid equations.

Internally, opty assumes that n = num equations = num states and then uses n everywhere. To remove this restriction we need to make num equations and num states distinct and then adjust all locations to use the correct number. I think that is all that is needed to allow systems with arbitrary number of path constraints added.

moorepants:
Fixes #438

Problem and ConstraintCollocator can now except M equations of motion that != n states. M = n was a hard requirement but it is unnecessary and prevents adding additional constraint equations that are valid for all time nodes. This is fixed by added a new attribute num_eom and adjusting the various functions that output the constraints to use it rather than num_states. A beginner example showing this works for a particle moving on a circle is included to demonstrate that additional path constraints works.

# TODO

- [x] Update the `theory.rst` documentation to indicate that the equations of motion can have additional path constraints.
- [x] Add tests for adding a motion constraint (make sure it works with extra differential equations that do not introduce new states).
- [x] Update the README to indicate path constraints are supported.
- [x] Add error handling if there are too few state variables provided.


moorepants:
@Peter230655 if you could give this a careful look, I would appreciate it. I'm sure there are other places in the code and documentation that need updating and maybe some more tests are needed.

Peter230655:
I will be happy to do this!
Could you tell me (again I am afraid) how I can get this version in my PC, so I can run this version?

moorepants:
>  Could you tell me (again I am afraid) how I can get this version in my PC, so I can run this version?

No, young birds must fly from the nest at some point :)

Peter230655:
Except the young bird will be 70 soon! :-)
I will try to get it.

Peter230655:
Could you not just merge it? Then I know exactly what to do to test it. :-)

moorepants:
No, we do not merge code until we know that it works.

Peter230655:
I managed to get it! :-)   
I played around with plot_algebraic.py

1.
When I replaced x**2 + y**2 - r**2 with x*vx + y*vy
I got the same result

2.
When I added x*vx + y*vy to the eoms (in addition to x**2 + y**2 - r**2)
I got the message: _too few degrees of freedom_ or similar.

I will continue to play around with other simulations a report any specil findings I might get.

moorepants:
>  I got the message: Algorithm converged to a point of local infeasibility. Problem may be infeasible

That sounds more like you have a problem that isn't feasible by its design, not that the software can't manage a second equation.

Peter230655:
I agree, I have gotten this message many times before.
At the moment I am unsure what plot_algebraic.py should do. I naively assumed the particle should move around on the circle, which it does not seem to do.

moorepants:
The particle move in a helix around the cylinder's surface.

Peter230655:
Works fine with me.

1. I added a meaningless eom: vx^2 + vz^2 - r^2/2. When I corrected the instance constrtaints, this worked without problems, converged.
2. In the old version, if the number of eoms did not equal the number of state variables, an error was given. Now, if fewer eoms than state variables are given, an error is thrown like: ValueError: cannot reshape array of size 202 into shape (3,101). A novice may have problems to conclude where his mistake is. I think, I might be good to give an error like: _At least as many eoms as there are state variables must be given,_ or similar.

moorepants:
> Now, if fewer eoms than state variables are given, an error is thrown like: ValueError: cannot reshape array of size 202 into shape (3,101).

I think we have to search the equations of motion for any derivatives of time and then raise an error if there are not matching state variables for each of those.

Peter230655:
Not sure that will do it: I just removed the last two eoms from your plot_algebraic.py. An error was raised, but I I did not remove any state variable so the a derivative was left dangling.

Are you saying M < n should be allowed?

moorepants:
I'm saying that for every `Derivative(x, t)` found in the equations of motion, `x` should be present in the provided state variables.

Peter230655:
```
eom = sm.Matrix([
    x.diff() - vx,
    y.diff() - vy,
    z.diff() - vz,
    m*vx.diff() - Fx,    
#    m*vy.diff() - Fy,
#    m*vz.diff() - Fz,
    x**2 + y**2 + z**2 - r**2,
])
```
In plot_alegraic.py this modification raises an error: cannot reshape.....

moorepants:
> Question: When a modification, like extra_constraint run through ther test-suite, will all examples of examples-gallery be tested with the new version? If not I can try to run a few to see if it works.

Yes. You should see if some of your hacks can be removed.

Peter230655:
Do you mean to test them whether they run without hacks?

NB: I pushed an example yesterday, maybe a bit in this direction:  There it would have been easy to remove some states, but convergence was much faster with these additional states.

moorepants:
Anywhere we have added equality constraints by  introducing an extra fake state variable and associated bounds shall be removed after this change is merged.

Peter230655:
I will check my simulations for such fake states.
Bu for **in**equalities, there still is only the 'hack solution', correct?

moorepants:
> x.diff(t, 2) = 0 

opty requires the differential equations to be in first order form, this issue is not about supporting higher order differential equations.

Peter230655:
What I meant was this: in one example, I want the object to have zero acceleration at the end of the run.
Let x be its coordinate, so I set in the eoms: ``x.diff(t, 2) = aux`` , aux an additional state variable, and I set an instance constraint
aux(tf) = 0.

moorepants:
Please open another issue if you want support for second (or higher order differential equations). It is helpful if every issue or PR does not diverge into tangent topics.

Peter230655:
> Please open another issue if you want support for second (or higher order differential equations). It is helpful if every issue or PR does not diverge into tangent topics.

No, I do not want this at all! My hack works fine. All I wanted to say is that extra-constraints likely will not eliminate this hack.

Peter230655:
I made r = me.dynamicsymbols('r'), r0 = sm.symbols(r0), and changed the instance constraints accordingly
and added it like this:
```
eom = sm.Matrix([
    x.diff() - vx,
    y.diff() - vy,
    z.diff() - vz,
    m*vx.diff() - Fx,
    m*vy.diff() - Fy,
    m*vz.diff() - Fz,
    x**2 + y**2 - r**2,
    r.diff(t) - r0
])
```
I get this new error:
```
ValueError: Too few state variables provided for state time derivatives found in equations of motion.
```
I did not add r to the state symbols, but to the specified symbols.

moorepants:
There is no support for providing the time derivatives of specified values.

moorepants:
@Peter230655 I am done with this from my side unless you have feedback for me to address. I will merge it sometime tomorrow likely if there is no feedback.

Peter230655:
> There is no support for providing the time derivatives of specified values.

How DUMB of me!

Peter230655:
> @Peter230655 I am done with this from my side unless you have feedback for me to address. I will merge it sometime tomorrow likely if there is no feedback.

I think, fine! The example _plot_path_constraints.py_ is really good now, the plot now shows what is going on.
Once merged, I can play around much better.

moorepants:
> How DUMB of me!

You can open an issue if we should support such a thing. The bicycle countersteering example shows that when models need to have two or more inputs that are time derivatives you have to put one all but one as a state. That doesn't fit the state/input division.

Peter230655:
> > How DUMB of me!
> 
> You can open an issue if we should support such a thing. The bicycle countersteering example shows that when models need to have two or more inputs that are time derivatives you have to put one all but one as a state. That doesn't fit the state/input division.

I will keep this in mind. In my examples I did not pay much attention to this division, just used whatever worked.
Now that you merged, I can play around again more easily! :-)

Peter230655:
Small observation (I never know where to pu such things):
I _added_ the eom x * vx + y * vy to the eoms. I ran fine with this additional eom, but results looked different. Theoretically they should be 'less restricting' than x^2 + y^2 0 = r^^2, but seem to affect the results.

moorepants:
Different results are expected due to constraint drift.

Peter230655:
I agree if I used **only** x * vx + y * vy.
But if i use it in **addition** to x^2 + y^2  = r^2, theoretically there should be no drift?
Anyway, now that it is merged, I can play around more.

moorepants:
If you find this does not work for certain needs, please open new issues with feature requests (or bug reports if that is the case).

PR #441, opened 2025-04-25
particle passes through gates at times determined by opty

Peter230655:
A particle must pass inside three gates, at times determined by opty, and at locations within the gates also determined by opty.
Currently opty does not support instance constraints of the form $x(t_i) - (a < x < b)$ where $t_i$ and the exact location of x is optimized by opty. This simulation shows a way around this.

Maybe interesting: I have two sets of equations , with $gate_i, time_i$ being state variables:
$gate_i$ = function(state variables, parameters)
$\dfrac{d}{dt}time_i$ = $gate_i$

It seems obvious to use
$\dfrac{d}{dt}time_i$ = function(state variables, parameters)

instead and drop $gate_i$.
Both versions eventually converge to the same solution, but the 'reduced' version took 480 sec on my PC vs. 70 sec for the 'full' version. The reason I think is this: With the full version, I could put some helpful restrictions on $gate_i$ which of course not possible with the reduced version.
So, at least here, fewer state variables did not speed up convergence.


Peter230655:
With my better understanding of the purpose of ``examples-gallery`` I do not think, this is a viable example.
As I continue to think it is interesting, I moved it to pst-notebooks and I close this PR.

PR #453, opened 2025-04-30
car in garage, used eom inequalities

Peter230655:
This is the simulation we have in examples-gallery/advanced.
I changed the following:
- replaced the hack to have inequalities in the eoms, with inequalities, which is possible now.
- replaced a state variable, needed to force an eom to have a definite value at tf with an unknown input trajectory, as we can now use instance constraints on them.
- use backend='numpy' to speed up the execution.
- applied flake8 to make it look better

I did not touch anything concerning the animation, but the gif in the preview does now work - at least not on my ``make html``


moorepants:
Nice to see that the new updates are working with a more complex example!

Peter230655:
My gut feeling is that it speeds up finding the solution, too. However, this example converges easily. I will (hopefully) be able to check on an example which does not converge so easily.

moorepants:
LGTM!

Peter230655:
> My gut feeling is that it speeds up finding the solution, too. However, this example converges easily. I will (hopefully) be able to check on an example which does not converge so easily.

Unfortunately, maybe my gut feeling was premature: I have a simulation of a brachistochrone, where an obstacle can be moved in its way, so it has to follow a strraight line for a while. (the basic idea is from betts)
Results of two runs each:
- using the hack: 189 sec on my PC, around 27,000 iterations
- new inequalities: 232sec/294 sec, around 35,000 iterations.

moorepants:
I don't know why that might be the case and the answer lies most likely lies with how IPOPT works internally. Our new way of setting inequality constraints is the way IPOPT is designed to be used. Wither IPOPT does something different with the bounds on x versus the bounds on c, I don't know.

Peter230655:
Of course, I keep on playing around with other simulations.
Are you interested to see this simulation? I can easily set it up so so can switch from one version to another one.

moorepants:
Adding examples to the documentation of SymPy, PyDy, and opty are all welcomed, but they should designed for learners of the software packages. The examples you make that are not designed for learning how to use these packages should go in your pst-notebooks repository and you can show me them there.

Peter230655:
Just moved it to pst-notebooks, under the name ``brachi_shift_hack_inequality_opty.py``
- a variablew ``hack`` may be set to True or False to select which method to use.
- if one runs both versions (take a bit of patience, each one runs about 3 min on my PC), the results will be compared. They differ very little. :-)
- the difference in running time is much less than earlier (no idea, I did not change aynthing that I am aware of), with hack still a bit faster.

Of course, next I will look at more simulations.

PR #459, opened 2025-05-07
Car around pylons inequalities

Peter230655:
This is the example ``car around pylons`` from examples-gallery/advanced.
I changed:
- used eom inequalities
- replaced state variables with unknown input trajectories where possible
- backend='numpy'
- cleanded up with flake8

Interesting aside: while ``no backend`` did not converge, backend='numpy' converged without problems.


Peter230655:
I made a mistake with the branches. I will start over and hence close this one.

PR #461, opened 2025-05-07
car around pylons. replaced hacks with inequalities

Peter230655:
This is the example plot_cars_around_pylons from examples-gallery/advanced.

- replaced hacks with inequalities
- replaced state variables with unknown input trajectories where possible
- backend='numpy'
- cleaned up with flake8

Interesting aside: While ``no backend`` did not converge, ``backend='numpy'`` did so without any problems.


moorepants:
> Interesting aside: While no backend did not converge, backend='numpy' did so without any problems.

That is a bit concerning, as they should give the same results.

PR #464, opened 2025-05-08
Constraint violations inequalities

Peter230655:
I changed along the lines discussed in #488, which it should fix.
I tested every combination/permutation I could think of.

When ``subplots=True`` I changed the numbering of the eoms, now starting with Eq 0 (before Eq 1). Reason is that like this the user can easily see his eom_bounds key matching the label on the respective plöot.

### DIFF
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -716,11 +716,22 @@ class Problem(cyipopt.Problem):
         instance_violations = con_violations[len(eom_violations):]
         eom_violations = eom_violations.reshape((self.collocator.num_eom,
                                                  N - 1))
-        # TODO : figure out a way to plot the inequality constraint violations
-        # don't plot inequality
-        if self.eom_bounds is not None:
-            for k, v in self.eom_bounds.items():
-                eom_violations[k] = np.nan
+        # If an eom has bounds a < eom < b and if subplots is False, this will
+        # be displayed for each  node:
+        # eom - a if eom < a
+        # eom - b if eom > b
+        # zero otherwise.
+        if self.eom_bounds is not None and not subplots:
+            for k in self.eom_bounds.keys():
+                left = self.eom_bounds[k][0]
+                right = self.eom_bounds[k][1]
+                for i in range(N - 1):
+                    if eom_violations[k, i] < left:
+                        eom_violations[k, i] = eom_violations[k, i] - left
+                    elif eom_violations[k, i] > right:
+                        eom_violations[k, i] = eom_violations[k, i] - right
+                    else:
+                        eom_violations[k, i] = 0.0
 
         con_nodes = range(1, self.collocator.num_collocation_nodes)
 
@@ -749,10 +760,16 @@ class Problem(cyipopt.Problem):
         else:
             for i in range(self.collocator.num_eom):
                 if ((self.eom_bounds is not None) and
-                    (i in self.eom_bounds.keys())):  # don't plot if inequality
-                    axes[i].plot(con_nodes, np.nan*np.ones_like(con_nodes))
-                    axes[i].set_ylabel(f'Eq. {str(i+1)} \n not shown',
+                    (i in self.eom_bounds.keys())):
+                    axes[i].plot(con_nodes, eom_violations[i])
+                    axes[i].set_ylabel(f'Eq. {str(i+1)} \n value',
                                        fontsize=9)
+                    axes[i].axhline(self.eom_bounds[i][0], color='C1', lw=1.0,
+                                     linestyle='--')
+                    axes[i].axhline(self.eom_bounds[i][1], color='C1', lw=1.0,
+                                     linestyle='--')
+
+
                 else:
                     axes[i].plot(con_nodes, eom_violations[i])
                     axes[i].set_ylabel(f'Eq. {str(i+1)} \n violation',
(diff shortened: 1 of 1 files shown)


Peter230655:
Again, I teste all combinations I could think of.

opty/direct_collocation.py
```diff
@@ -651,7 +651,24 @@ def plot_constraint_violations(self, vector, axes=None, subplots=False):
...
+
+        if ``eom_bounds`` are given and ``subplots = False``, the eom
+        violations are plotted. The violations are calculated as follows:
+
+        - eom - a if eom < a
+        - eom - b if eom > b
+        - 0 otherwise.
+
+        If only one eom is given and and eom bounds are given, then
+        ``subplots`` is set to ``True``.
```
moorepants, line 666: `subplots` should only be set to true or false by the user.

moorepants: Also "eom" is not a word.

Peter230655: > `subplots` should only be set to true or false by the user.

If only one equation of motion is given, subplots really does not make sense.
I set it to True in the program to simplify the 'logical flow' if only one eom is given.

Peter230655: > Also "eom" is not a word.

Maybe best, I remove this sentence completely?

moorepants: At the minimum, whether the output is an array of axes or a single axis matters also.

Peter230655: > At the minimum, whether the output is an array of axes or a single axis matters also.

I believe this was taken care of earlier, It worked with my one-eom example, but I have to double check.

Peter230655:
The violations of the inequality constraints are shown, as discussed during the recent hackathon.
I added an optional keyword ``show_range``, default is False. If True and if subplots is True, its shows the range of the bounded eoms.
My reason: 
Maybe in a < eom_i < b, a is set way too small, and b way too large. with``show_range == True`` this will be visible. Making a larger and b smaller reduces the size of the ``search space`` for Ipopt and may make it faster.
If this reasoning is wrong, the key word should be removed.

Peter230655:
I am sure I have suitable examples (it's been a while I have to look).
Practical question: If I make an example which shows it, and if I push it, it will fail as this is not merged?
If I add the example to the same branch, will it work?

I am not sure I understand the 'kwarg point': it is optional, so backward compatibility should be no issue?

moorepants:
Show that your code functions in the examples and/or unit tests. You've done this numerous times before, so I'm confused as to why you are confused.

moorepants:
Take for example this PR https://github.com/csu-hmc/opty/pull/365 there is a flag in the unit tests `plot=True` that lets you visually check all of the scenarios you wrote. We have to have some way to verify in the future the code still does what the documentation says it does. If it is sufficient to show in an existing example, that is also good (and preferred because we see any changes every time the docs are built).

Peter230655:
Oh, you mean put a test function in ``test_direct_collocation.py`` which plots various combinations of the constraint violations?

moorepants:
No, not exactly. You should add it to an example if that is sufficient and does not detract from the simplicity of the example, for example setting the kwarg to True in one location is fine. If you have lots of plots to check from different combinations of kwargs, then it is preferable to extend a unit test (or make a new one: less preferable) with a plot kwarg so we can at least manually check these things in the future.

You have been writing unit tests and examples for years now, so I am a bit impatient having to explain every last detail again.

Peter230655:
- I added a test function, which shows (if plot=True) various combinations of the kwargs
- mofdified an existing example (plot_car_on_racecourse_smooth.py) to show the eom bounds
- deleted a line in direct_collocation.py (I used it to let me know some code was correct and forgot to delete it when I first pushed it a while ago.)

opty/tests/test_direct_collocation.py
```diff
@@ -344,6 +344,55 @@ def test_extra_algebraic(plot=False):
...
+        # 2nd test with 5 bounds, balance with 2 bounds on eoms.
+        ax = prob.plot_constraint_violations(solution, show_range=True)
+        ax[0].set_title('One plot of all eom violations, '
+                        'no eom_bounds available')
+
+        prob.eom_bounds = {0:(-0.5, 0.5),
+                           1:(-1.0, 1.0),
+                           2:(-2.0, 2.0),
+                           3:(-2.0, 2.0),
+                           4:(-0.1, 0.1)}
```
moorepants: Setting this after the problem is created does not trigger any of the checks or set the bounds in the IPOPT problem. In the plotting code you only use this attribute, so this works. But eom_bounds should actually not be settable if we follow the design of the other attributes. So the fact that this works is more an error on our part.

moorepants: It does that, which is nice, but it also exposes a hole in the design. We added all the property attributes and prevent setting them because it breaks the problem. Setting eom_bounds breaks the problem.

Peter230655: I can easily change my test when I get home, but if I understood correctly, this would not change the problem it accidentally found?

opty/direct_collocation.py
```diff
@@ -804,28 +816,97 @@ def plot_constraint_violations(self, vector, axes=None, subplots=False):
...
                     axes[i].plot(con_nodes, eom_violations[i])
-                    axes[i].set_ylabel(f'Eq. {str(i+1)} \n violation',
+                    axes[i].set_ylabel(f'Eq. {str(i)} \n violation',
                                        fontsize=9)
                 if i < self.collocator.num_eom - 1:
                     axes[i].set_xticklabels([])
             axes[num_eom_plots-1].set_xlabel('Node Number')
             axes[0].set_title('Constraint violations')
+        else:
+            raise ValueError('Something wrong with EOM constraints.')
```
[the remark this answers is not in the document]
moorepants: If invalid arguments get you to this place, then it should give an informative error message. Just saying something is wrong hides what might be happening.

PR #472, opened 2025-05-24
Support for known trajectories that are functions of the state

moorepants:


opty/tests/test_direct_collocation.py
```diff
@@ -14,6 +14,41 @@
...
+    states = (x, v)
+
+    eom = sm.Matrix([
+        m*v.diff() - f + m*g*sym.cos(theta(x)),
+        x.diff() - v,
+    ])
+
+    N = 20
+
+    def calc_theta(x):
```
moorepants, line 32: opty will need to know internally what to pass to `calc_theta`. This could be like the objective function and it only accepts the `free` array and it is  up to the user to do the calculations with the contents of `free` to ensure that an array is returned. This matches the objective function definitions.

An alternative is to make the symbolic term a function of the specific state, input, and parameter variables and then opty can know what specific values to pass to the function. This is more complicated to implement but maybe cleaner. I guess you'd need to be able to do things like `theta(x, v, m, g, f, t)` in general. But maybe we don't support explicit `t` (same as we don't yet in the EoMs.

moorepants:
@Neville-N I think this has the solution you need. Please have a look. I need to do some more work to clean things up, add it to the documentation, and then test it in an example problem, but I think this is all we need to support use of numerical functions for known trajectories.

moorepants:
After a bike ride in the rain, I convinced myself that this is a hack that happens to work. If there is a variable `theta`, in this case, that is a function of the state `theta(x(t))`, then when the Jacobian of the equations are formed, there should be `d theta(x(t)/dx(t)` terms present in the Jacobian. I'm ignoring that with this design.

Peter230655:
> After a bike ride in the rain, I convinced myself that this is a hack that happens to work. If there is a variable `theta`, in this case, that is a function of the state `theta(x(t))`, then when the Jacobian of the equations are formed, there should be `d theta(x(t))/dx(t)` terms present in the Jacobian. I'm ignoring that with this design.

So, can I play around with it?

examples-gallery/beginner/plot_hilly_ride.py
```diff
@@ -0,0 +1,172 @@
...
+s, v, x, y, p, theta = me.dynamicsymbols('s, v, x, y, p, theta', real=True)
+
+states = (x, y, s, v)
+
+eom = sm.Matrix([
+    x.diff() - v*sm.cos(theta),
+    y.diff() - v*sm.sin(theta),
+    s.diff() - v,
+    m*v.diff() - p/v + m*g*sm.sin(theta) + v**2/3,
+])
```
moorepants, line 44: If `theta = sm.Function('theta')(x)`, then there are these derivatives that should end up in the Jacobian:

```
In [17]: sm.simplify(eom.diff(x))
Out[17]: 
Matrix([
[ v(t)*sin(theta(x(t)))*Derivative(theta(x(t)), x(t))],
[-v(t)*cos(theta(x(t)))*Derivative(theta(x(t)), x(t))],
[                                                   0],
[  g*m*cos(theta(x(t)))*Derivative(theta(x(t)), x(t))]])
```

So, the current implementation is ignoring these Jacobian entries, i.e. setting them to zero, but the solution is still found. If I make theta be a symbolic explicit function of x it takes 29 iterations instead of 52. So I guess IPOPT can get to a solution even if the Jacobian is slightly degenerate.

The proper solution for allowing this kind of numerical function should be to pass in a function for the constraints to use and a function for the derivative terms for the Jacobian to use. This would look something like:

```python
theta = Function('theta')(x, r, p)
Problem(
    ...
    numerical_functions={theta: (eval_theta, eval_dtheta_dx, eval_dtheta_dr, eval_dtheta_dp)}
)
```

or something like that.

Peter230655: Which explicit function did you use in the Hilly Ride?
I tried sm.interpolate(....), but it did not work with me.

moorepants:
```python
In [12]: amp, omega = sm.symbols('A, omega')
    ...: theta = sm.atan(amp*omega*sm.cos(omega*x))
    ...: 
    ...: eom = sm.Matrix([
    ...:     x.diff() - v*sm.cos(theta),
    ...:     y.diff() - v*sm.sin(theta),
    ...:     s.diff() - v,
    ...:     m*v.diff() - p/v + m*g*sm.sin(theta) + v**2/3,
    ...: ])
    ...: 

In [13]: eom.diff(x)
Out[13]: 
Matrix([
[                                                                                -A**2*omega**3*v(t)*sin(omega*x(t))*cos(omega*x(t))/(A**2*omega**2*cos(omega*x(t))**2 + 1)**(3/2)],
[-A**3*omega**4*v(t)*sin(omega*x(t))*cos(omega*x(t))**2/(A**2*omega**2*cos(omega*x(t))**2 + 1)**(3/2) + A*omega**2*v(t)*sin(omega*x(t))/sqrt(A**2*omega**2*cos(omega*x(t))**2 + 1)],
[                                                                                                                                                                                0],
[   A**3*g*m*omega**4*sin(omega*x(t))*cos(omega*x(t))**2/(A**2*omega**2*cos(omega*x(t))**2 + 1)**(3/2) - A*g*m*omega**2*sin(omega*x(t))/sqrt(A**2*omega**2*cos(omega*x(t))**2 + 1)]])
```

Peter230655: In ``plot_car_around_pylons.py`` I replaced 4 symbolic hump functions with numeric hump functions in two eoms.
I did not converge, while with the same initial_guess it converged with symbolic hump functions (needed 2,000....4,000 iterations)

moorepants:
This example probably works because dtheta/dx is small:
image

I had trouble getting it to converge if I make the amplitude of the sin wave larger (steeper hill, thus dtheta/dx getting larger).

Peter230655:
On these hump functions the derivatives at the edges are quite steep. So this would explain what I found, I think.

moorepants:
Ok, I think this now supports forming the Jacobian correctly.

moorepants:
@tvdbogert we may be able to use this new feature to create a variable representing percent gait cycle and then setup two numerical functions to manage its and its derivative calculation.

moorepants:
The derivative checker is off because theta(x) gets replaced by thetai:

```python
In [2]: prob.collocator.discrete_eom
Out[2]: 
Matrix([
[                    -vi*cos(thetai) + (xi - xp)/h],
[                    -vi*sin(thetai) + (yi - yp)/h],
[                                -vi + (si - sp)/h],
[g*m*sin(thetai) - pi/vi + vi**2/3 + m*(vi - vp)/h]])

In [3]: prob.collocator.current_discrete_state_symbols
Out[3]: (xi, yi, si, vi)

In [4]: prob.collocator.discrete_eom.jacobian(prob.collocator.current_discrete_state_symbols)
Out[4]: 
Matrix([
[1/h,   0,   0,            -cos(thetai)],
[  0, 1/h,   0,            -sin(thetai)],
[  0,   0, 1/h,                      -1],
[  0,   0,   0, pi/vi**2 + 2*vi/3 + m/h]])
```

so the extra terms in the Jacobian are not included.

moorepants:
I hope to have it working and finished by the end of the week. We now have a hackathon on Friday with Peter, me, and two students!

moorepants:
```
ipdb> symbolic_partials
([(z0, thetaixi), (z1, 1/h), (z2, sin(z0)), (z10, Derivative(z0, xi)), (z12, -1/h**2), (z13, cos(z0)), (z14, z10*z13)], [Matrix([
[vi*z10*z2 + z1,  0,  0,                     -z13, -z1,   0,   0,     0,     0,   z12*(xi - xp)],
[       -vi*z14, z1,  0,                      -z2,   0, -z1,   0,     0,     0,   z12*(yi - yp)],
[             0,  0, z1,                       -1,   0,   0, -z1,     0,     0,   z12*(si - sp)],
[       g*m*z14,  0,  0, m*z1 + pi/vi**2 + 2*vi/3,   0,   0,   0, -m*z1, -1/vi, m*z12*(vi - vp)]])])
```

There may be derivatives in the cses that need to be replaced properly.

Peter230655:
If I want to play around with this new feature, I would have to merge this:

https://github.com/moorepants/opty/tree/implicit-known-traj

with my local master. Correct?

moorepants:
You shouldn't merge it with your local master. It is best to keep your local master as a match with the master in this repo. You can pull this branch and play with the branch.

moorepants:
This is what you do to try my branch:

```
git remote add moorepants git@github.com:moorepants/opty.git  # if not already there
git checkout --track moorepants/implicit-known-traj
```

Peter230655:
Thanks! I will try and report.

moorepants:
>  If I want to make a simulation, I would branch off the implicit-know-traj into some branch, go to examples-gallery/beginner, say, and put my simulation there. Correct?

Yes.

Peter230655:
_If I understood your last sentence correctly, a surface would not yet be possible._

Is this correct?

moorepants:
This implementation only works with functions of a single variable. So z(x) works but z(x, y) will not.

Peter230655:
Goit my first simulation with numerical known trajectories to work! 😊   😊

Peter230655:
@moorepants If you havw time, could you tell me how to push an example? I asked chatGPT, but that sounded awfully difficult.
Thanks!

examples-gallery/beginner/plot_hilly_race.py
```diff
@@ -0,0 +1,234 @@
...
+axes[1].plot(x_meas, np.rad2deg(theta_meas))
+axes[1].set_ylabel(r'$\theta$ [deg]')
+axes[2].plot(x_meas, np.rad2deg(dthetadx_meas))
+axes[2].set_ylabel(r'$\frac{d\theta}{dx}$ [deg/m]')
+_ = axes[2].set_xlabel(r'$x$ [m]')
+
+# %%
+# The following functions output the slope and its derivative with respect to
+# :math:`x(t)` at all values of :math:`x(t)` using interpolation. The only
+# input to these functionshould be the optimization free vector and the output
```
moorepants, line 88:
```suggestion
# input to these functions should be the optimization free vector and the output
```

PR #476, opened 2025-05-31
Adds example of using linear complementarity constraints and slack variables to manage piecewise friction

### LINKED ISSUE #475: Create an example that uses slack variables for a discontinous input
Here is how we handled friction in our skateboard paper:

Image

https://link.springer.com/article/10.1007/s12283-023-00448-y

A simple beginner example with Coulomb friction could demonstrate this.

moorepants:
Attempt to fix #475 

Observations:

- The solutions for just going right or just going left work as expected but require good initial guesses.
- It is very difficult to get the exact expected solution at the friction switching point.
- I often get "error in step computation" from IPOPT, but don't recall seeing this in other problems in the past. Different initial guesses will sometimes cause an error. This will occur in the Cython or NumPy backends differently for the same problem conditions.
- It seems to require a very good initial guess.

TODO:

- [x] This should probably be intermediate (or even advanced) not beginner.
- [x] Move the code to skip plotting the first node to another PR.
- [x] Move to advanced.

### DIFF
--- a/docs/references.rst
+++ b/docs/references.rst
@@ -24,7 +24,7 @@ References
    and John McPhee.  "Parameter Identification in Dynamic Systems Using the
    Homotopy Optimization Approach". Multibody System Dynamics 26, no. 4 (2011):
    411-24.
-.. [Posa2013] Posa, M., Cantu, C., & Tedrake, R. (2014). A direct method for
+.. [Posa2013] Posa, M., Cantu, C., & Tedrake, R. (2013). A direct method for
    trajectory optimization of rigid bodies through contact. The International
    Journal of Robotics Research, 33(1), 69–81.
    https://doi.org/10.1177/0278364913506757


moorepants:
@Neville-N this may be helpful for you as an option to have elements in your equations that switch discontinuously (in alternative to making smoothened step functions).

@Peter230655 I'm curious what you can do with this and whether you can get better solutions than I am. Have a look!

examples-gallery/beginner/plot_friction_slack.py
```diff
@@ -0,0 +1,170 @@
...
+constant_symbols = (m, mu, g)
+specified_symbols = (F(t), Ffn(t), Ffp(t), psi(t))
+
+eom = sm.Matrix([
+    # equations of motion with positive and negative friction force
+    x(t).diff(t) - v(t),
+    m*v(t).diff(t) - Ffp(t) + Ffn(t) - F(t),
+    # following two lines ensure: psi >= abs(v)
+    psi(t) + v(t),  # >= 0
+    psi(t) - v(t),  # >= 0
```
moorepants, line 39: Peter and I discussed making this `psi(t) - v(t)**2` instead.

moorepants: Or maybe it needs to be `psi - sqrt(v**2)`.

moorepants: I tried this and it seems to work.

Peter230655:
Update 5-6-25:
- With slack variables, I still cannot get the nice results shown above. It says it converges, but Ffp, Ffn have unreasonable values initially.
- If I model the friction as Ffp = Fr * smooth_step(v(t)) and Ffn = Fr * smooth_step(-v(t)) it converges 'immediately',

moorepants:
>  It says it converges, but Ffp, Ffn have unreasonable values initially.

Yes, that happens with me often in this problem.

moorepants:
One more thing: all unknown inputs to this system can be discontinuous with respect to time. We have many problems with such solutions in the examples.

Peter230655:
> Also, note that this is the same formulation we used for solving the skateboard ollie, which is also solved by IPOPT.

This is what kept me wondering, why it does not easily work here.

moorepants:
No, that creates a discontinuous constraint function.

Peter230655:
I finally got these results;
I changed the integration method to midpoint.
I bound as follows: 0 <= Ffp, Ffn <= mu * m *g
It took about 4500 iterations.
The result, approx 1.2 sec is twice as large as what I get with the smooth step, which needs around 120 iterations with (just about) any initial guess.
image

moorepants:
> Question: what is the advantage of the slack variable method over the smooth step method?

In this case, its advantage is that it can find the correct solution for a system with Coulomb friction vs. the solution with an approximation of Coulomb friction. In general, the advantage is to find solutions for differential equations that have discontinuities.

moorepants:
The only stipulation is that the NLP objective and constraint functions are continuous. This says nothing about the differential algebraic equations. Opty converts the description of the optimal control of differential algebraic equations into a NLP problem via direct collocation. If the DAE's are continuous, then the NLP problem is continuous, but if the DAE's are not continuous, then you may be able to still construct a continuous NLP. That's what this example is attempting to show.

Peter230655:
Update 7-6-2025:
1. Integration method ``midpoint" gives highly oscillation results, both with slack variable and with smooth step
2. I used the result of the smooth step as initial guess for slack variable, (and psi = abs(v)). It rapidly converged to a slightly worse result (h = 0.0065 vs. h = 0.0064). At midpoint, when v changes its sign from + to - Ffn jumps to around -6 then becomes 0 as it should. Slightly perturbing this initial guess seems to move the convergence behaviour in all directions.

moorepants:
I was wondering what it would do if given the initial guess as the smooth solution. Interesting.

Note that psi = abs(v) makes the NLP problem discontinuous. SymPy will produce a derivative for abs() but the discontinuity is present in the constraints and its Jacobian.

Peter230655:
I did not express myself correctly:
I did not change the eoms. As initial guess for psi (which does not exist in the smooth step) I used psi = v if v > 0 and psi = .v if v <0.

moorepants:
In the NLP formulation there are the unknown variables x that are subject to `c_l < c(x) < c_u` and `x_l < x < x_u`. So, in that view all unknown variables are the same "kind".

Peter230655:
Makes sense.
I still do not understand why with ``slack variables`` it has such difficulties to converge and with ``smooth step`` it converges so easily.
The problems are different, but are they so different physically?

Peter230655:
It is the same formulation used for the skateboard ollie, which converged easily?

moorepants:
It is the formulation shown in the paper, but I don't know if that is what is in the code.

Peter230655:
Right now I have no ideas, will continue to play around.
While I have no ideas if you could let me know about #464 I could finish it.

Peter230655:
I think, in this sense I have used slack variables in some of my simulations, without being aware of it:
I used a variable, say ``close(t)'' which only increased if my particle was close to a gate, with ``close(t0) = 0.0``

I needed this variable to be positive and tf to ensure the particle was close to the gate at some time during the simulation, but no idea of the value of ``close(tf)``
So I introduced one more variable ``close1(t)`` and set ``close1(t) = aux * clsoe(t)``, and bounded, say, 0.1 < aux < 100
In the instance I set close1(tf) = 1.0

Peter230655:
> I think, in this sense I may have used slack variables in some of my simulations, without being aware of it: I used a variable, say `close(t)` which only increased if my particle was close to a gate, like `close(t).diff(t) = hump(x, a, b)` with hump(x, a, b) = 1.0 for a < x < b, zero otherwise, x the coordinate of the particle. Of course: `close(t0) = 0.0`
> 
> I needed this variable to be positive at the final time tf to ensure the particle was close to the gate at some time during the simulation, but no idea of the value of `close(tf)` - and also not really that important, just positive. So I introduced one more variable `close1(t)` and set `close1(t) = aux * clsoe(t)`, and bounded, say, 0.1 < aux < 100 In the instance constraints I set close1(tf) = 1.0
> 
> This ensured, that the particle was close to the gate at some point, it also worked in my simulations.

moorepants:
Nice!

I'm curious if it solves without having to give the hump solution as the initial guess. If so, I'd prefer that in the gallery to keep it minimal.

If we add this to the gallery, I'd prefer if my commits are retained and you open a PR that is based on a branch off of this PR (with your commits added).

Peter230655:
Sorry, I did not change the eoms. I changed bounds:
I made ``(mu*m*g - Ffp(t) - Ffn(t))*psi(t)``,  an equality, before it was bound to be positive.
I added bounds 0 <= FFn, FRp <= m * g * mu

moorepants:
In a chat with Ton today he showed me how he uses slack variables (linear complementarity functions) to handle the foot contact force. He said something like "I have to use > epsilon instead of > 0 because Ipopt doesn't linke the hard constraint". We used slack variables in the skateboard paper with Ipopt, but I don't recall this limitation (we were using pycollo not opty, but I don't think that should matter for this issue).

I tried somethings like:

```python
eom_bounds = {
    2: (1e-10, np.inf),
    3: (1e-10, np.inf),
    4: (1e-10, np.inf),
}
```

and 

```python
prob.add_option('tol', 1e-06)
```

and combinations thereof with different values and I can get solutions to find an optimal faster, but it still gives similar solutions to what I already get (same as https://github.com/csu-hmc/opty/pull/476#issuecomment-2926750911).

I was hitting an odd bug that Ipopt did not like the value I gave for tol, saying wasn't a double. This did not occur all the time and I couldn't figure it out.

moorepants:
From https://www.researchgate.net/post/How_can_I_have_a_solution_with_the_guaranteed_strict_complementarity_condition_using_available_solvers_like_IPOPT:

> The problem with the use of interior-point methods for complementarity problems is that there isn't an interior solution at the optimal. You can relax the complementarity conditions using i.e. the smoothed Fisher-Burmeister equation. It also enables you to compute good valid bounds by means of convex relaxation, if necessary conditions hold. That way, you can validate the solution of the local minimizer, i.e. IPOPT.

Peter230655:
I tried ``Fb = -tanh(kv(t))``. k some positive number, e.g. k = 50
This means
- Fb always opposite in direction to v(t)
- Fb = 0 iff v(t) = 0

This seems to converge much more easily.
Would such an approximation not be good enough for any practical application?
( for k = 50, |tanh(k*x)| < 0.999 for x in (-0.08, 0.08) )

examples-gallery/beginner/plot_friction_slack.py
```diff
@@ -0,0 +1,279 @@
+"""
+Coulomb Friction with Linear Complimentarity Constraints
+========================================================
+
+A block of mass :math:`m` is being pushed with force :math:`F(t)` along on a
+horizontal surface. Coulomb friction acts between the block and the surface.
```
Peter230655: Would it make sense to say that the surface is straight? Actually it is x axis with gravity acting in -y direction.

moorepants:
> I am curious about how many IPOPT iterations it took, can that be reported? It is a measure of how hard the problem is to solve.

You can see it in the objective plot, about 200.

> I just tried this (in Matlab, sorry) and it would not solve. I remember trying this on the gait model, many years ago, and didn't work either. The slack variables really make it work.

I had prior variations that had fewer slack variables (never none), but also had difficulty getting solutions. It is curious why the slack variables help. I guess the answer lies in understanding how Ipopt solves the NLP problem.

>  I am not sure if there would be additional challenges for problems where the solution involves spending more time at v=0.

Yes, I'm curious about that too.

I think I will merge this, so there is a basic example of how to do these things. It will help some of my students' projects.

PR #480, opened 2025-07-03
plot_find_points. Trajectory through points, hitting times decided by…

Peter230655:
A particle must move from A to B as fast as possible, but must hit some points on the way. Hitting times decided by ``opty``.
In oder to achieve this, I use a certain form of the ``objective function``, described in detail in the docstrings of the simulation.
Basically I try to minimize the distance of the particle from each point on the way. (I have done similar things before, but I used differentiable hump functions then. This seems to be faster.)
I pushed it to examples-gallery/intermediate because I think, the objective function is not totally trivial.

@moorepants : 
A bit disturbing is this: While the program runs fine with backend='cython', it raises an error with backend='numpy'.
As only the objective function is non-trivial here, I assume, it is related to it - of course I do not know.


Peter230655:
Update: The problem does not seem to be the objective function or its gradient. backend='numpy`also raises an error with a simple objective function.

Peter230655:
Changing to backend='numpy' reduced running time from about 23 sec to 16 sec (with solution given)

Peter230655:
I added a somewhat more complicated simulation using the same idea to pst-notebooks/gallery: ``car hits points``. While here the convergence was reliable, with the more complicated simulation it became a bit tricky, it would easily converge to local minima far away from a sensible minimum.
Maybe this is not a very good method to get ``opty`` to determine the time of an instance.

Peter230655:
> Hi Peter, I do not plan to include this in the opty examples. It does not demonstrate any specific feature of opty that we have not already. I think it is best for pst-notebooks. Note that there is a new docs page that links to your notebooks as a resource: https://opty.readthedocs.io/stable/resources.html

Of course I noticed the link! :-)  
With pst-notebooks there is no threat of a time limit, I guess. I will move it there.

PR #487, opened 2025-08-17
Racecar implicit known traj

Peter230655:
This is virtually identical to the race car example already in examples/gallery/intermediate.
The differences are:
1. The two lines limiting the street are now _numerically known trajectories_.
2. I control only one point of the car (its center) to remain on the road at all times. Reason is that only one variable may be used, but the locations of other points on the car depend on two variables. (In the original example any number of points along the body of the car may be required to remain on the road)


Peter230655:
I do not know, why it says I change 5 files. I only (knowingly) changed two.

moorepants:
Good to see this works for you.

Note that I use interpolation in the hilly race only because we will need that when using real data from a road elevation profile. You can simply use the lambdified functions directly here. I think the interpolation causes convergence issues in the hilly race due to the interpolation not being valid outside the range of the interpolation function.

PR #489, opened 2025-08-18
Two coords implicit known traj

Peter230655:
A particle has to move on a 2D numerically given surface in 3D space.

1. I use a hack to handle a numerical function. This was discussed in #488, just a temporary fix.
2. Higher order derivatives of the numerical functions appear in the equations of motion. This shows a way to handle the situation. I think this is independent of the hack above. Described in detail in the docstrings.

The animation is jerky. I had to reduce the fps to reduce running time.

I did not push this with the aim to get it merged, but just so you could see point 2. above.


moorepants:
Yes, I do not plan to merge it.

PR #490, opened 2025-08-19
 check for invalid keys in eom_bounds. Should fix #484

Peter230655:
If keys are not in range(len(eoms)) a ValueError should be raised.


Peter230655:
;-)

PR #491, opened 2025-08-27
Avoid hole implicit traj

Peter230655:
A particle is to move from A to B with a hole obstructing the direct path. The hole is given as two implicit trajectories.
Avoiding that the particle falls into the hole is achieved with two bounded eoms, containing the undefinded functions.
A differentiable step function ensures that only one of these two eoms is active at any one time.

@moorepants Should this simulation does not contain enough 'new stuff', just close it. I can then üut it in pst-notebooks once the next opty is released


moorepants:
This is an interesting problem but I think best for pst noteooks as we already have an example that shows how to use the functions for a known trajectory.

Peter230655:
I fully agree! Where the two bounds meet, they are not differentiable - but still it works!😊

PR #496, opened 2025-09-28
remove known_trajectory_symbols. Should fix #495

### LINKED ISSUE #495: AttributeError: 'ConstraintCollocator' object has no attribute '_known_trajectory_symbols'
Was exploring an opty model and got:

```
problem.collocator.known_trajectory_symbols
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[6], line 1
----> 1 ids.problem.collocator.known_trajectory_symbols

File ~/miniconda/envs/openloopbalancecontrol/lib/python3.13/site-packages/opty/direct_collocation.py:1236, in ConstraintCollocator.known_trajectory_symbols(self)
   1230 @property
   1231 def known_trajectory_symbols(self):
   1232     """
   1233     The known trajectory symbols.
   1234     Type: (m-q)-tuple
   1235     """
-> 1236     return self._known_trajectory_symbols

AttributeError: 'ConstraintCollocator' object has no attribute '_known_trajectory_symbols'
```

Peter230655:
I did not find ``_known_trajectory_symbols`` anywhere in ``direct_collocation.py``.
So I removed that property / method.
Thr correct name seems to be ``-known_input_trajectories``

### DIFF
--- a/opty/direct_collocation.py
+++ b/opty/direct_collocation.py
@@ -1227,13 +1227,6 @@ class ConstraintCollocator(object):
         """
         return self._known_trajectory_map
 
-    @property
-    def known_trajectory_symbols(self):
-        """
-        The known trajectory symbols.
-        Type: (m-q)-tuple
-        """
-        return self._known_trajectory_symbols
 
     @property
     def next_known_discrete_specified_symbols(self):
@@ -1329,7 +1322,7 @@ class ConstraintCollocator(object):
     @property
     def num_known_input_trajectories(self):
         """
-        The number of known trajectories = len(known_trajectory_symbols).
+        The number of known trajectories = len(known_input_trajectories).
         Type: int
         """
         return self._num_known_input_trajectories


Peter230655:
Unclear how the failed tests realted to my small change.

moorepants:
> Unclear how the failed tests relate to my small change.

You have to read the logs in the failed CI runs.

Peter230655:
> > Unclear how the failed tests relate to my small change.
> 
> You have to read the logs in the failed CI runs.

It failed on ``plot_sit_to_stand.py`` there is a line mass_matrix = symbolics[0]
and the objection was that Symbolics is not suscribable.

Peter230655:
> Removing API from a released version requires a deprecation cycle.
> 
> Secondly, you need to check whether this existed before you converted things to properties. If it did, then we definitely have to deprecate it. If it didn't I may be willing to remove it without deprecation if you confirm it never worked.

I looked in direct_collocation.py of v 1.3.0 and did not find ``known_trajectory_symbols`` or ``_known_trajectory_symbols`` there.
Is this enough, or do I have to look at all previous versions of direct_collocation.py?

Peter230655:
> Sounds like a reasonable theory !

Only one I have at the moment! :-)

PR #498, opened 2025-10-17
Macos 15 intel

moorepants:


moorepants:
I do not know of any reason to add a kwarg to the objective or gradient. The problem class will ignore it, as per the design of cyipopt.

Peter230655:
No wonder I did not see it.

PR #499, opened 2025-10-22
Add Problem methods to extract values by symbol name and fill the gradient by symbol name.

moorepants:
TODO:

- [x] add tests
- [x] add to example
- [x] decide if `extract_var` should extract known values (and have `free` as a kwarg`) it is also fairly easy to just pull from par_map and traj_map.
- [x] If traj_map is a function, then we should support returning the correct values.
- [x] decide if names of methods are good/appropriate, e.g. should fill_gradient just be fill_vector or fill_free?
- [x] precalculate all indices and just extract from a dict, will make function faster

### DIFF
--- a/examples-gallery/beginner/plot_hilly_race.py
+++ b/examples-gallery/beginner/plot_hilly_race.py
@@ -218,7 +218,7 @@ theta_vals = prob.extract_value(theta, solution)
 fig, ax = plt.subplots()
 ax.plot(x_vals, np.rad2deg(theta_vals))
 ax.set_xlabel(x)
-ax.set_ylabel(theta)
+_ = ax.set_ylabel(theta)
 
 # %%
 # Animation of the particle traversing the profile:


moorepants:
>  Will this also work in the objective function?

yes.

Peter230655:
Really nice idea! Saves this counting as the where, say, the torque is, if i understood correctly.

examples-gallery/beginner/plot_hilly_race.py
```diff
@@ -211,6 +211,15 @@ def obj_grad(free):
...
 _ = prob.plot_trajectories(solution)
 
+# %%
+# Plot the slope angle in degrees versus the horizontal distance.
+x_vals = prob.extract_value(x, solution)
+theta_vals = prob.extract_value(theta, solution)
+fig, ax = plt.subplots()
+ax.plot(x_vals, np.rad2deg(theta_vals))
+ax.set_xlabel(x)
+ax.set_ylabel(theta)
```
moorepants:
```suggestion
_ = ax.set_ylabel(theta)
```

PR #109, opened 2023-09-03
Use pycompilation if installed.

moorepants:

PR #503, opened 2025-10-27
Enables caching of the compiled code if tmp_dir is set.

### LINKED ISSUE #107: Add ability to cache the "JIT" Cython compiled libraries
Once you have a model defined and correct, there is no reason to recompile it. Thus, it would be nice if the Cython compiled libraries are somehow cached. Opty can already write the files to disk, so we need some kind of check to know if the compiled library is the same as the one generated prior. We could do something like hash the sympy object that corresponds to the inputs to ufuncify_matrix and then store that string in the generated cython file.  Once that is stored you could check for a match and skip the call to ufuncify_matrix.

I've seen this feature in Bjorn's https://github.com/bjodah/pycompilation.

moorepants:
Fixes #107

A hash of the unique evaluation code is stored in its C file and the files present in the tmp_dir are checked to see if they contain the hash. If one does, then that compiled module is loaded if possible. This saves having to recompile the C code if the differential algebraic equations passed to Problem or ConstraintCollocator have not changed.

TODO:

- [x] Add some unit tests.
- [x] Add documentation of the feature.
- [ ] <strike>Include it in an example that has more than one execution of Problem.</strike> can add later with human gait update
- [x] Handle parallel vs not, as that is not present in the eval_code string.
- [x] Deal with const being populated or not.

### DIFF
--- a/opty/utils.py
+++ b/opty/utils.py
@@ -633,7 +633,7 @@ def ufuncify_matrix(args, expr, const=None, tmp_dir=None, parallel=False,
         A matrix of expressions or the output of ``cse()`` of a matrix of
         expressions.
     const : tuple, optional
-        This should include any of the symbols in args that can be constant
+        This should include any of the symbols in args that should be constant
         with respect to the evaluation loop.
     tmp_dir : string, optional
         The path to a directory in which to store the generated files. If None


Peter230655:
Just for my understanding: if the DAEs do not change but, say, the instance constraints or the bounds / eom bounds change this would work?

moorepants:
Correct.

Peter230655:
Just for understanding: does Problem 'normally' create sets two cython files every time it rums?

moorepants:
Over in gait-closed-loop-id I'm getting a failed to import even though it finds a match:

```
08:31:53 INFO utils - ufuncify_matrix: Changed directory to /home/moorepants/src/gait-closed-loop-id/gait_codegen
08:31:53 INFO utils - ufuncify_matrix: Checking ufuncify_matrix_4 for cached code.
08:31:53 INFO utils - ufuncify_matrix: Checking ufuncify_matrix_3 for cached code.
08:31:53 INFO utils - ufuncify_matrix: Checking ufuncify_matrix_2 for cached code.
08:31:53 INFO utils - ufuncify_matrix: Checking ufuncify_matrix_1 for cached code.
08:31:53 INFO utils - ufuncify_matrix: Checking ufuncify_matrix_0 for cached code.
08:31:53 INFO utils - ufuncify_matrix: ufuncify_matrix_0 matches!
08:31:53 INFO utils - ufuncify_matrix: Failed to import ufuncify_matrix_0.
08:31:53 INFO utils - ufuncify_matrix: Compiling matrix evaluation.
```

moorepants:
Interesting I have the opposite result here: https://github.com/csu-hmc/opty/pull/109#issuecomment-1704010089

PR #513, opened 2025-10-30
Move parallel example to gallery.

moorepants:


moorepants:
Running this locally with minimal other processes eating cpu, you get:

```
moorepants@nandi:opty(move-parallel)$ ipython examples-gallery/beginner/plot_parallel.py 
Time for non-parallel: 7.072258315001818
Time for parallelized: 1.9007170830009272
```

So there is a nice speed up.

I think openmp may not even be correctly installed on the readthedocs, so we don't see this result in the built docs. Also if we run the docs build in parallel (which is the current configuration), then it is unlikely to see a gain either because other problems are consuming the cores.

moorepants:
Disabling the sphinx-gallery parallel build so that we can show the parallelization speed up on 2 cores in the examples is probably not a good idea given that it pushes the RTDs building time to ~ 15 minutes.

PR #535, opened 2026-04-04
Use order='none' for cse() calls in C code generation.

### LINKED ISSUE #506: Consider using `cse(..., order='none')` for faster common sub expression extraction
Opty currently does not (likely) rely on the order of the subexpressions, so we could use `order='none'` in `cse()`. Although it may fully foil the current attempt in #503.

I also do not think `lambdify()` lets you set this, unless you can pass a callable to the `cse` kwarg.

moorepants:
Fixes #506

This gives a performance gain when generating common subexpressions for very large equations. order='none' will not guarantee uniqueness of the generated C code across machines but it does not seem to effect the ability to hash and cache the C code in opty, i.e. the generated C code is unique when running cse() on the same machine. If we ever notice that the C code caching does not work for repeated calls on the same machine, we need to come back to this change and revert it or devise an alternative.

### DIFF
--- a/opty/utils.py
+++ b/opty/utils.py
@@ -138,7 +138,8 @@ def _forward_jacobian(expr, wrt):
 
     logger.info('Adding expression nodes to cache...')
     start = timer()
-    replacements, reduced_exprs = sm.cse(expr.args[2], replacement_symbols)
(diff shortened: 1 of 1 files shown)


moorepants:
I'll merge this an monitor the C code caching in https://github.com/mechmotum/gait-closed-loop-id while we use the development version.

PR #537, opened 2026-04-06
disc pushing another disc

Peter230655:
Disc2, at rest at point A is pushed to a point C by another disc (disc1) located at a starting position B, using an input force applied to disc1.
There are no new methods used.
I feel, this is a novel application: disc1 and disc2 are "unrelated" for opty. It must do something to disc2, and has as a tool disc1, But how to get the 'idea' to use disc1 to push around disc2?

I got the basic idea from Posa et al 2013.
I pushed a similar, but much more complicated, simulation to pst-notebooks.


moorepants:
I am not going to add this example to opty, as it demonstrates no new features of opty. Sorry.

Peter230655:
No problem at all!
For me the most fun is doing it.!

PR #538, opened 2026-04-10
park tractor-trailer

Peter230655:
A tractor with two trailers is parked into a slot. Possible example for motivating problems.
I pushed it here, rather than to pst-notebooks a suggested, because
- I use adjustable bounds, not (yet) available with pst-notebooks
- I do not believe we have an example for adjustable bounds, yet.

Interesting: It found a sensible path relatively easily, also meeting all the constraints, but to get an optimal solution turned out to be difficult.


moorepants:
This is a fun example, but I'm not going to add it to opty's documentation. We can add variable bounds to an existing example to demonstrate that feature.

Peter230655:
I did not expect you to merge this monster. I just pushed it so you can see it.
I think, I will push it to pst-notebokks once you release the next opty with this feature.

PR #540, opened 2026-04-11
particle through box. simple example for adjustable bounds

Peter230655:
A particle must move from A to B as fast as possible. At a certain time interval, it must be in a given box, that is
min_x < x(t) < max_x, for $t \in$ time imterval, same for y(t).
Likely the simplest possible simulation to show how to use ``adjustable bounds``

PR #538 also uses adjustable bounds, but for a beginner, this use may be drowned in the many ReferenceFrames and Points used there.


moorepants:
I'll accept an example that has non-constant bounds on variables. But I think we could simply add variable bounds to one of the existing problems, versus adding more problems that further increase the documentation build times.

Peter230655:
I did not think of this, of course this is better!
I will pick one, add variable bounds and push.

PR #543, opened 2026-04-14
add variable bounds on the torque

Peter230655:
This is likely the simplest example we have, so I added variable bounds here.

### DIFF
--- a/examples-gallery/beginner/plot_pendulum_swing_up_fixed_duration.py
+++ b/examples-gallery/beginner/plot_pendulum_swing_up_fixed_duration.py
@@ -7,6 +7,7 @@ Objective
 ---------
 
 - Show the use of opty with likely the simplest example possible.
+- Show how to use adjustable bounds.
 
 
 Introduction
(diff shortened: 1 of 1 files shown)


examples-gallery/beginner/plot_pendulum_swing_up_fixed_duration.py
```diff
@@ -7,6 +7,7 @@
 ---------
 
 - Show the use of opty with likely the simplest example possible.
+- Show how to use adjustable bounds.
```
moorepants:
```suggestion
- Show how to use variable bounds.
```

examples-gallery/beginner/plot_pendulum_swing_up_fixed_duration.py
```diff
@@ -84,8 +85,16 @@
...
-# Limit the torque to a maximum magnitude.
-bounds = {T(t): (-2.0, 2.0)}
+# Limit the torque using variable bounds.
+third = int(num_nodes/3)
+low_T =  np.array([-3 for _ in range(third)] +
+                  [-1 for _ in range(third)] +
+                  [-3 for _ in range(num_nodes - 2 * third)])
+hi_T = np.array([3 for _ in range(third)] +
+                [1 for _ in range(third)] +
+                [3 for _ in range(num_nodes - 2 * third)])
```
moorepants: This are how to use NumPy to do such things:

```python
hi = np.full(n, 3.0)
hi[100:200] = 1.0
```

Peter230655: > This are how to use NumPy to do such things:
> 
> ```python
> hi = np.full(n, 3.0)
> hi[100:200] = 1.0
> ```

Thanks!
My python is still very limited! :-((

Peter230655:
Should we close this?
I think, our opty PRs should not become as cluddered as sympy's  :-)

moorepants:
Seems like I was reviewing it with an intention to get it in.

